Issue #5, opened 2021-05-03
Set up catalogue app to view and test UI components


PascalPrecht:
As discussed in https://github.com/status-im/status-desktop/issues/2438, we need a playground app to put our newly developed components in. We no longer put this inside Status Desktop. It probably makes sense to set it up as part of this repository.

PR #24, opened 2021-05-04
feat: Set up catalog app (sandbox)

borismelnik:
#5 

From Qt Creator:
1. Open in Qt Creator file `sandbox.pro`
2.  Build :)
3.  Run

From console:
1. run `path/to/qt/qmake sandbox.pro -spec macx-clang CONFIG+=debug CONFIG+=x86_64 && /usr/bin/make qmake_all`
2. run `make`
3. Binaries must be placed in `sandbox/bin`

### DIFF
--- a/sandbox/main.qml
+++ b/sandbox/main.qml
@@ -1,0 +1,135 @@
+import QtQuick 2.14
+import QtQuick.Window 2.14
+import QtQuick.Controls 2.14
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+Window {
+    id: rootWindow
+    width: 640
+    height: 480
+    visible: true
+    title: qsTr("Status App Sandbox")
+
+    ButtonGroup {
+        id: topicsGroup
+        buttons: tabs.children
+    }
+
+    Flow {
+        id: tabs
+        width: parent.width
+        Button {
+            text: "Reload QML"
+            onClicked: app.restartQml()
+        }
+        Button {
+            id: iconsTab
+            checkable: true
+            text: "Icons"
+        }
+    }
+
+    ScrollView {
+        width: parent.width
+        anchors.top: tabs.bottom
+        anchors.bottom: parent.bottom
+        contentHeight: rootWindow.height * rootWindow.factor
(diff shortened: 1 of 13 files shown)


jrainville (approved):
Great work!

sandbox/main.qml
```diff
@@ -0,0 +1,112 @@
...
+            checkable: true
+            text: "Bars"
+        }
+        Button {
+            checkable: true
+            text: "Modals"
+        }
+        Button {
+            checkable: true
+            text: "inputs"
```
PascalPrecht, line 50: This should probably be "Inputs" then

sandbox/main.qml
```diff
@@ -0,0 +1,112 @@
...
+        buttons: tabs.children
+    }
+
+    Flow {
+        id: tabs
+        width: parent.width
+        Button {
+            checkable: true
+            text: "Text Styles"
+        }
```
PascalPrecht, line 23: I don't believe it make sense to have this one as there most likely won't be multiple text style components (probably just `StatusText` which can then be adjusted in size and weight).

sandbox/main.qml
```diff
@@ -0,0 +1,112 @@
...
+                scale: rootWindow.factor
+            }
+
+            Rectangle {
+                id: darkTheme
+
+                SplitView.fillWidth: true
+                SplitView.minimumWidth: rootWindow.width / 2
+                height: parent.height
+                color: "#2c2c2c"
```
PascalPrecht, line 79: Now that #23 has landed, we can make use of `Theme.palette.baseColor5` here.

PascalPrecht (changes requested):

PascalPrecht:
Really great work indeed!

I've noticed that, after running this, there's untracked `.qmake.stash` and `MakeFile` files.
Can you add them to `.gitignore` please?

PascalPrecht (approved):

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/3528b2ff440a8fb66609f7087f062643f5082132
(I've updated your commit message so it automatically closes #5)

Awesome work Boris!

PR #25, opened 2021-05-05
feat(Core): introduce StatusBaseText component

### LINKED ISSUE #20: Implement text type styles
Original issue: https://github.com/status-im/status-desktop/issues/2328

PascalPrecht:
Abstract text component that defaults to Status' `Inter` font.
Also known as `StyledText` in Status Desktop.

This will close #20 as it doesn't make sense to implement multiple
components for a single text type. Size and weight of text can be configured
on a per usage basis.

Closes #20

### DIFF
--- a/src/StatusQ/Core/StatusBaseText.qml
+++ b/src/StatusQ/Core/StatusBaseText.qml
@@ -1,0 +1,6 @@
+import QtQuick 2.13
+import StatusQ.Core.Theme 0.1
+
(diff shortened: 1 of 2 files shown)


PascalPrecht:
Since it's unlikely there's going to be another text component, we might consider renaming this one to just `StatusText`.

Thoughts?

borismelnik (approved):

PR #26, opened 2021-05-05
feat(Components): introduce StatusLoadingIndicator

### LINKED ISSUE #7: Move StatusLoadingIndicator over to component library repo
That component was introduced in https://github.com/status-im/status-desktop/issues/2360

PascalPrecht:
A `StatusIcon` that rotates infinitely and can be used for indicating
pending states.

Usage:

```
StatusLoadingIndicator {
    width: 24 // default: 17
    height: 24 // default: 17
    color: "red" // default: loading asset color
}
```

Closes #7

### DIFF
--- a/src/StatusQ/Components/StatusLoadingIndicator.qml
+++ b/src/StatusQ/Components/StatusLoadingIndicator.qml
@@ -1,0 +1,18 @@
+import QtQuick 2.13
+import StatusQ.Core 0.1
+
+StatusIcon {
+    id: statusIcon
+    icon: "loading"
+    height: 17
+    width: 17
+    RotationAnimator {
+        target: statusIcon;
+        from: 0;
+        to: 360;
+        duration: 1200
+        running: true
+        loops: Animation.Infinite
+    }
+}
+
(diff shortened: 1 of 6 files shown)


borismelnik (approved):

src/StatusQ/Components/StatusLoadingIndicator.qml
```diff
@@ -0,0 +1,18 @@
+import QtQuick 2.13
+import StatusQ.Core 0.1
+
+StatusIcon {
+    id: statusIcon
+    icon: "loading"
+    height: 17
+    width: 17
+    RotationAnimator {
```
jrainville, line 9: In my opinion, we should take the opportunity that we have here since we are remaking the UI library to change this loading indicator to an animated gif instead of this. We've seen multiple times in the past that having this loading indicator increases CPU usage absurdly, so using a gif instead should fix the problem.

PascalPrecht: That's a very good point. Do we keep `ColorOverlay` capabilities though? We need to stay flexible and be able to apply colors as we see fit

jrainville: Good question, I'm not sure

PR #40, opened 2021-05-07
refactor(sandbox): change sandbox app layout

PascalPrecht:


PascalPrecht:
@borismelnik this was the direction I was looking for (obviously, this is a work in progress)
Screenshot 2021-05-07 at 14 07 51

One thing I've noticed: the hot reload breaks our theming system. Idea how to fix this? Seems like `ThemePalette` instances are destroyed after reload...

PascalPrecht:
Updated, looks kinda like this now:
Screenshot 2021-05-27 at 19 08 39

Issue #41, opened 2021-05-10
Implement new app layout in sandbox app


PascalPrecht:
We want the sandbox app not only become a place to view and test components of StatusQ, it should also serve as documentation of various components and their different configurations, states etc.

This requires us to change the app layout to a more documentation like layout with the available modules plus their components listed on the left hand side and each component getting their own view/page to explore different options.

I've took a stab on it in #40 just to get the idea rolling.

PascalPrecht:
Now that we've landed #85 and #86 , we could potentially make use of the latter to build a new sandbox layout

PR #43, opened 2021-05-12
feat: Add buttons components

borismelnik:
Example's in `Buttons.qml` #6 
ezgif com-gif-maker

### DIFF
--- a/sandbox/main.qml
+++ b/sandbox/main.qml
@@ -7,19 +7,29 @@ import StatusQ.Core.Theme 0.1
 
 Window {
     id: rootWindow
-    width: 640
-    height: 480
+    width: 1024
+    height: 840
     visible: true
     title: qsTr("Status App Sandbox")
+
+    property ThemePalette lightTheme: StatusLightTheme {}
+    property ThemePalette darkTheme: StatusDarkTheme { }
 
     ButtonGroup {
         id: topicsGroup
         buttons: tabs.children
     }
 
+    ButtonGroup {
+        buttons: switchRow.children
+    }
+
     Flow {
         id: tabs
-        width: parent.width
+        anchors.left: parent.left
+        anchors.leftMargin: 100
+        anchors.right: parent.right
+
         Button {
             text: "Reload QML"
             onClicked: app.restartQml()
@@ -39,7 +49,15 @@ Window {
             checkable: true
             text: "Other"
         }
+
+        Button {
+            id: buttonsTab
+            checkable: true
+            text: "Buttons"
+        }
     }
+
+
 
     ScrollView {
         width: parent.width
@@ -48,70 +66,62 @@ Window {
         contentHeight: rootWindow.height * rootWindow.factor
         contentWidth: rootWindow.width * rootWindow.factor
         clip: true
-        SplitView {
-            width: parent.width
-            height: rootWindow.height
-            handle: Item {}
 
+
+        Rectangle {
+            id: lightThemeBg
+
+            width: rootWindow.width
+            height: parent.height
+            color: Theme.palette.baseColor5
+            clip: true
             scale: rootWindow.factor
 
-            Rectangle {
-                id: lightThemeBg
+            Loader {
+                active: true
+                anchors.centerIn: parent
 
-                SplitView.minimumWidth: rootWindow.width / 2
-                height: parent.height
-                color: lightTheme.baseColor5
-                clip: true
-
-                Loader {
-                    active: true
-                    anchors.centerIn: parent
-                    property var currentTheme: StatusLightTheme { id: lightTheme }
-
-                    sourceComponent: {
-                        switch(topicsGroup.checkedButton) {
-                        case iconsTab:
-                            return iconsComponent;
-                        case controlsTab:
-                            return controlsComponent;
-                        case otherTab:
-                            return othersComponent;
-                        default:
-                            return null;
-                        }
+                sourceComponent: {
+                    switch(topicsGroup.checkedButton) {
+                    case iconsTab:
+                        return iconsComponent;
+                    case controlsTab:
+                        return controlsComponent;
+                    case otherTab:
+                        return othersComponent;
+                    case buttonsTab:
+                        return buttonsComponent;
+                    default:
+                        return null;
                     }
                 }
-
             }
+            Row {
+                id: switchRow
+                scale: 0.8
+                anchors.right: parent.right
+                anchors.top: parent.top
 
-            Rectangle {
-                id: darkThemeBg
-
-                SplitView.fillWidth: true
-                SplitView.minimumWidth: rootWindow.width / 2
-                height: parent.height
-                color: darkTheme.baseColor5
-                clip: true
-
-                Loader {
-                    active: true
-                    anchors.centerIn: parent
-                    property var currentTheme:  StatusDarkTheme { id: darkTheme }
-
-                    sourceComponent: {
-                        switch(topicsGroup.checkedButton) {
-                        case iconsTab:
-                            return iconsComponent;
-                        case controlsTab:
-                            return controlsComponent;
-                        case otherTab:
-                            return othersComponent;
-                        default:
-                            return null;
+                Button {
+                    checkable: true
+                    checked: true
+                    text: "Light Theme"
+                    onCheckedChanged: {
+                        if (checked) {
+                            Theme.setTheme(lightTheme)
                         }
                     }
                 }
 
+                Button {
+                    checkable: true
+                    text: "Dark Theme"
+                    onCheckedChanged: {
+                        if (checked) {
+                            Theme.setTheme(darkTheme)
+                        }
+                    }
+                }
             }
         }
     }
@@ -147,7 +157,7 @@ Window {
         id: iconsComponent
         Icons {
             anchors.centerIn: parent
-            iconColor: parent? parent.currentTheme.primaryColor1 : "#ffffff"
+            iconColor: Theme.palette.primaryColor1
         }
     }
 
@@ -155,7 +165,7 @@ Window {
         id: controlsComponent
         Controls {
             anchors.centerIn: parent
-            theme: parent.currentTheme
+            theme: Theme.palette
         }
     }
 
@@ -163,7 +173,14 @@ Window {
         id: othersComponent
         Others {
             anchors.centerIn: parent
-            theme: parent.currentTheme
+            theme: Theme.palette
+        }
+    }
+
+    Component {
+        id: buttonsComponent
+        Buttons {
+            anchors.centerIn: parent
         }
     }
 }
--- a/sandbox/Buttons.qml
+++ b/sandbox/Buttons.qml
@@ -1,0 +1,239 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.14
+
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+
+Grid {
+
+    columns: 3
+    columnSpacing: 38
+    rowSpacing: 10
+
+    horizontalItemAlignment: Grid.AlignHCenter
+
+    // Large
+    StatusButton {
+        text: "Button"
+    }
+
+    StatusButton {
+        text: "Button"
+        enabled: false
+    }
+
+    StatusButton {
+        text: "Button"
+        loading: true
+    }
+
+    StatusButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+    }
+
+    StatusButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+        enabled: false
+    }
+
+    StatusButton {
+        text: "Button"
+        loading: true
+        type: StatusBaseButton.Type.Danger
+    }
+
+    StatusFlatButton {
+        text: "Button"
+    }
+
+    StatusFlatButton {
+        text: "Button"
+        enabled: false
+
+    }
+
+    StatusFlatButton {
+        text: "Button"
+        loading: true
+    }
+
+    StatusFlatButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+    }
+
+    StatusFlatButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+        enabled: false
+    }
+
+    StatusFlatButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+        loading: true
+    }
+
+    // Small
+    StatusButton {
+        text: "Button"
+        size: StatusBaseButton.Size.Small
+    }
+
+    StatusButton {
+        text: "Button"
+        enabled: false
+        size: StatusBaseButton.Size.Small
+    }
+
+    StatusButton {
+        text: "Button"
+        size: StatusBaseButton.Size.Small
+        loading: true
+    }
+
+    StatusButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+        size: StatusBaseButton.Size.Small
+    }
+
+    StatusButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+        size: StatusBaseButton.Size.Small
+        enabled: false
+    }
+
+    StatusButton {
+        text: "Button"
+        type: StatusBaseButton.Type.Danger
+        size: StatusBaseButton.Size.Small
+        loading: true
+    }
+
+    StatusFlatButton {
+        text: "Button"
+        size: StatusBaseButton.Size.Small
+    }
+
+    StatusFlatButton {
+        text: "Button"
+        enabled: false
+        size: StatusBaseButton.Size.Small
+    }
+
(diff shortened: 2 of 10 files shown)


src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+    enum Size {
+        Small,
+        Large
+    }
+
+    enum Type {
+        Normal,
+        Danger,
+        WithoutBackground,
+        DangerWithoutBackground
```
PascalPrecht, line 19: I wonder if `WithoutBackground` and `DangerWithoutBackground` should really be a type here.

The built-in `Button` type already comes with a flag `flat` which kind of does the same thing. 

What do you think about sticking to that "standard" and have a `flat` property instead, or, if that makes things too complex, how about we introduce something like:

```
StatusFlatButton {
  type: Normal // or Danger
}
```

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+    property int type: StatusButton.Type.Normal
+
+    property alias text: label.text
+    property alias font: label.font
+
+    property real leftPadding: size === StatusButton.Size.Large ? 24 : 12 // by design
+    property real rightPadding: size === StatusButton.Size.Large ? 24 : 12 // by design
+    property real topPadding: size === StatusButton.Size.Large ? 11 : 10 // by design
+    property real bottomPadding: size === StatusButton.Size.Large ? 11 : 10 // by design
+
```
PascalPrecht, line 34: I think we can remove the `// by design` comments here

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+    signal pressed(var mouse)
```
PascalPrecht, line 35: Any reason we aren't calling this "clicked" ?

I personally am a fan of keeping as much of "native" component feeling around. Users that will use `StatusButton` might not expect an `onPressed` signal

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+
+
+
+    /// Implementation
```
PascalPrecht, line 39: This comment we can remove

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+
+    implicitWidth: sensor.width
+    implicitHeight: sensor.height
+
+    radius: 8 // by design
```
PascalPrecht, line 44: Same here `// by design` doesn't add any value here

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+    color: {
+        switch(statusButton.type) {
+        case StatusButton.Type.Normal:
+            if (statusButton.enabled) {
+                return sensor.containsMouse ? Theme.palette.primaryColor2
+                                            : Theme.palette.primaryColor3;
+            } else {
+                return Theme.palette.baseColor2;
+            }
```
PascalPrecht, line 53: Tiny nitpick: since we're return in any case if `statusButton.enabled`, we can have the second `return` without the `else` branch

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+
+        case StatusButton.Type.Danger:
+            if (statusButton.enabled) {
+                return sensor.containsMouse ? Theme.palette.dangerColor2
+                                            : Theme.palette.dangerColor3;
+            } else {
+                return Theme.palette.baseColor2;
+            }
```
PascalPrecht, line 61: Same here

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+
+        cursorShape: Qt.PointingHandCursor
+        hoverEnabled: !loading
+
+
+        onPressed: {
+            if (!loading) {
+                statusButton.pressed(mouse)
+            }
+        }
```
PascalPrecht, line 102: Shouldn't this be `onClicked` ?

src/StatusQ/Buttons/StatusButton.qml
```diff
@@ -0,0 +1,165 @@
...
+                        return Theme.palette.baseColor1;
+                    }
+
+                }
+            }
+        } // Text
+
+        StatusLoadingIndicator {
+            anchors.centerIn: parent
+            visible: loading
```
PascalPrecht, line 140: Should we use a `Loader` and make it `active: loading` instead of just using visibility to render this indicator?

sandbox/main.qml
```diff
@@ -7,19 +7,29 @@ import StatusQ.Core.Theme 0.1
...
     id: rootWindow
-    width: 640
-    height: 480
+    width: 1024
+    height: 840
     visible: true
     title: qsTr("Status App Sandbox")
 
+    property ThemePalette lightTheme: StatusLightTheme {}
+    property ThemePalette darkTheme: StatusDarkTheme { }
```
PascalPrecht, line 16: These changes should live in a separate commit/PR

src/StatusQ/Buttons/StatusIconButton.qml
```diff
@@ -0,0 +1,82 @@
...
+
+Rectangle {
+    id: statusIconButton
+
+    property color iconNormalColor: "transparent"
+    property color iconDisabledColor: "transparent"
+
+    property color backgroundNormalColor: "transparent"
+    property color backgroundHoverColor: "transparent"
+    property color backgroundDisabledColor: "transparent"
```
PascalPrecht, line 15: I'm probably missing something here, but at a first glance, this looks off to me.
All of these colors are `transparent` by default, meaning that, if I put a `StatusIconButton` in my app, I won't see a thing, unless I explicitly assign colors.

`Theme` has been built to take care of that.

Any reason you've made it this way?

src/StatusQ/Buttons/StatusIconButton.qml
```diff
@@ -0,0 +1,82 @@
...
+    property alias iconName: icon.icon
+
+    property bool showBackground: false
```
PascalPrecht, line 18: This flag should probably also rather be encoded in a `type` similar to how it's done in `StatusButton`

src/StatusQ/Buttons/StatusIconButton.qml
```diff
@@ -0,0 +1,82 @@
...
+            anchors.centerIn: parent
+            visible: loading
+            color: {
+                if (statusIconButton.enabled) {
+                    return iconNormalColor
+                } else {
+                    return iconDisabledColor
+                }
+            }
+        } // Indicator
```
PascalPrecht, line 80: Similar to `StatusButton`: should this be rather rendered using a `Loader`?

src/StatusQ/Buttons/StatusIconButton.qml
```diff
@@ -0,0 +1,82 @@
...
+    } // Sensor
+} // Rectangle
```
PascalPrecht, line 82: Another thing to keep in mind here:

This component is currently called `StatusRoundButton` (analogue to QML's `RoundButton`), while `StatusIconButton` is a different component.

If we aren't careful, this is going  to break many places in the app.

How do you feel about sticking with `StatusRoundButton` as a name for now?

src/StatusQ/Buttons/StatusTextWithIconButton.qml
```diff
@@ -0,0 +1,104 @@
...
+    property alias text: label.text
+    property alias font: label.font
+
+
+    property color iconNormalColor: "transparent"
+    property color iconDisabledColor: "transparent"
+
+    property color backgroundNormalColor: "transparent"
+    property color backgroundHoverColor: "transparent"
+    property color backgroundDisabledColor: "transparent"
```
PascalPrecht, line 19: Just like I've mentioned in the other comment: We don't want consumers of this component to be in charge of assigning all the colors.

src/StatusQ/Buttons/StatusTextWithIconButton.qml
```diff
@@ -0,0 +1,104 @@
...
+            color: {
+                if (statusTextWithIconButton.enabled) {
+                    return iconNormalColor
+                } else {
+                    return iconDisabledColor
+                }
+            }
+        } // Indicator
+    } // Sennsor
+} // Rectangle
```
PascalPrecht, line 104: This component should probably be just a composition of existing components.

I wonder if it would be feasible to implement a "slot" in `Status(Flat)Button` instead

PascalPrecht (changes requested):
Great work @borismelnik !

There are a bunch of things I believe need clarification or adjustment. In some places I left a comment to get your insights on. Would love to learn more about why certain decisions have been made.

The Theme toggling should probably live in a separate PR, also because I've addressed it as well in #41 Plus, there's another PR #44 which doesn't make sense on its own without the toggling actually being in place.

Let me know if you need help with this!

src/StatusQ/Buttons/StatusIconButton.qml
PascalPrecht: Referring to `flat` here

src/StatusQ/Buttons/StatusTextWithIconButton.qml
PascalPrecht: To elaborate, we could have something like:

```
StatusFlatButton {
  text: "some text"
  icon.name: "chat"
}
```

PascalPrecht:
One more thing:

Do we want buttons to live in `StatusQ.Controls` instead? All QML buttons live in `Controls` so maybe we want to stay consistent here? Wdyt?

src/StatusQ/Buttons/StatusButton.qml
borismelnik: I think we can use it like base low level component. But without C++ extensions its always looks ugly. Enums in QML is ugly too :)
So we can "hide" it by "alias" like this:
`
/// StatusDangerButton.qml
       StatusButton {
           type: StatusButton.Type.Danger  
       }
`

src/StatusQ/Buttons/StatusButton.qml
borismelnik: Every "clickable" components have signals `pressed` `released `clicked` and `pressAndHold`. So we don't have `pressed` state of button. `clicked` is pressed+released. If you want i can add it. Its just my `legacy`

src/StatusQ/Buttons/StatusButton.qml
borismelnik: Its annotation where `public` api

src/StatusQ/Buttons/StatusButton.qml
borismelnik: The same thing. After 1-2-3 years you could forgot why its `8`?? this annotation for `magic numbers`

src/StatusQ/Buttons/StatusIconButton.qml
borismelnik: Okay, no problem

src/StatusQ/Buttons/StatusIconButton.qml
borismelnik: I think it must be `Loader`, thanks. But maybe we could fix indictor like this:
`
StatusIcon {
    id: statusIcon
    icon: "loading"
    height: 17
    width: 17
    RotationAnimator {
        target: statusIcon;
        from: 0;
        to: 360;
        duration: 1200
        running: statusIcon.visible
        loops: Animation.Infinite
    }
}
`

src/StatusQ/Buttons/StatusIconButton.qml
borismelnik: No problem

src/StatusQ/Buttons/StatusTextWithIconButton.qml
borismelnik: Yes but its can be used as base component. If in future we never want to create some red button with icon - no problem. But i don't think so.

src/StatusQ/Buttons/StatusTextWithIconButton.qml
borismelnik: I can do it, but i think its just one button in Figma.

borismelnik:
> One more thing:
> 
> Do we want buttons to live in `StatusQ.Controls` instead? All QML buttons live in `Controls` so maybe we want to stay consistent here? Wdyt?

yes. i will fix it

src/StatusQ/Buttons/StatusButton.qml
PascalPrecht: Right, then how about we introduce `StatusDangerButton` and `StatusFlatButton` in this PR accordingly?

src/StatusQ/Buttons/StatusButton.qml
PascalPrecht: I guess the question is: What is most likely used/expected by users?

So far we've been using `onClicked` on all buttons in Status Desktop. If you think `onPressed` is rather the way to go, then that's fine by me. We'll just have to touch many places in the app to make sure nothing breaks.

We could also just have both `clicked` and `pressed`. If we want to be consistent, we should probably have all of the events mentioned above, because, it's a button after all (one of the reasons I've inherited from `Button` instead of doing a custom `Rectangle`)

src/StatusQ/Buttons/StatusButton.qml
PascalPrecht: Is that a QtCreator thing? If this is a standard way of marking public APIs in QML, we should make sure we do it in all of our components.

src/StatusQ/Buttons/StatusButton.qml
PascalPrecht: Alright. I see where you're coming from.

My take would then be: What other reason is there, if not design?

Do we want to put `// by design` comments on every `topMargin`, `padding` etc as well? Because all those things are by design

src/StatusQ/Buttons/StatusIconButton.qml
PascalPrecht: So mean, we shouldn't use `StatusLoadingIndicator` and instead conditionially apply the animator?

Because that's eactly what `StatusLoadingIndicator` does.

src/StatusQ/Buttons/StatusTextWithIconButton.qml
PascalPrecht: If  we want to treat it as a base component, then this PR should also introduce the more concrete equivalents.

But since we have a `Theme` system in place for that, I don't think we have to start with a `transparent`-all-the-way component.

src/StatusQ/Buttons/StatusButton.qml
borismelnik: annotation

src/StatusQ/Buttons/StatusButton.qml
borismelnik: So no problem, i will forward signals from `MouseArea`

src/StatusQ/Buttons/StatusButton.qml
borismelnik: No, but if you check Qt QML sources i can find something similar.

src/StatusQ/Buttons/StatusButton.qml
borismelnik: i always do it, but if you want, i can remove it :) its just some info. I think some comments are not superfluous.

PascalPrecht: We can leave them, I think I just need some help from you then to decide in the future, which QML properties receive this comment (since everything is by design).

src/StatusQ/Buttons/StatusIconButton.qml
borismelnik: Just one reason - no logic for colors :) i will fix it by default.

src/StatusQ/Buttons/StatusIconButton.qml
borismelnik: No we just take care about "always running animation". If you check `BusyIndicator` in QML you can find public property for stop/start it.

PascalPrecht: Okay, happy to see that in a PR :)

PascalPrecht (commented):
Amazing work!

PascalPrecht (approved):
Very well done.

I think we can merge this. If there was one thing I'd change, then it's that, the cursor should probably not be a pointer when the buttons are in `loading` state.

I'll leave that to you to decide :)

Great great work!

PascalPrecht (approved):

PascalPrecht:
🎉

PR #44, opened 2021-05-12
chore: Potencial fix hot reloading and remove title bar example

borismelnik:


sandbox/sandboxapp.cpp
```diff
@@ -25,12 +25,23 @@ void SandboxApp::startEngine()
                 QCoreApplication::exit(-1);
         }, Qt::QueuedConnection);
     m_engine.load(url);
+
+    QWindow *rootWindow = qobject_cast<QWindow*>(m_engine.rootObjects().at(0));
+    if (rootWindow) {
+        removeTitleBar(rootWindow->winId());
+    } else {
+        qDebug() << "window is not exists";
```
PascalPrecht, line 33:
```suggestion
        qDebug() << "Window doesn't exist";
```

src/StatusQ/Core/Theme/Theme.qml
```diff
@@ -4,6 +4,9 @@ import QtQuick 2.13
 
 QtObject {
     id: appTheme
+    // Replace it with:
+    //      property QtObject palette: StatusLightTheme {}
+    // for reloading
```
PascalPrecht, line 9: Is this the "potential" fix for hot reloading theme palettes?

Let's put this in a separate commit then, ideally together with the changes that introduce theme switching (otherwise there's nothing to be fixed in the first place)

PascalPrecht (changes requested):

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/8737c06ab8ac5752c245b40534295f016aa43f7c

PR #50, opened 2021-05-20
feat(StatusIconTabButton): introduce image loading state and fallback

### LINKED ISSUE #37: Implement loading/pending state for StatusIconTabButton
In #36 we're introducing `StatusIconTabButton` to StatusQ, which, depending on its configuration can render a `StatusRoundedImage` (introduced in #33).

For a first version of this component, that's fine, however it should be able to properly handle loading states of such images. In Status Desktop, the current implementation uses an `ImageLoader` to fetch images from ipfs if needed.
While we don't want IPFS to be a specific protocol backed into the component, we should ensure that any image, regardless from where it's loaded, triggers a "pending" or "loading" state, that causes `StatusIconTabButton` to render a loading indicator (probably `StatusLoadingIndicator`.

We probably need to decide what should happen if loading such a resources fails. One decent default would be to fall back to `StatusLetterIdenticon`. 

We might need to revisit the component's heuristics that decide which sub component to load, because at the moment, whenever a `name` is given (which is needed for `StatusLetterIdenticon`) a letter identicon component is always rendered.

PascalPrecht:
This commit introduces a loading indicator for cases where `icon.source`
is set (a custom image/icon) to improve UX of the component.

It also falls back to a letter identicon in case loading the image source
wasn't successful.

Usage:

```
StatusIconTabButton {
    icon.source: "SOME URL THAT CAN'T BE RESOLVED"
    icon.color: "red" // in case fallback is used, icon.color will be gray by default
    name: "Some channel" // used to generated letter identicon as fallback
}
```

Closes #37

### DIFF
--- a/src/StatusQ/Controls/StatusIconTabButton.qml
+++ b/src/StatusQ/Controls/StatusIconTabButton.qml
@@ -40,10 +40,35 @@ TabButton {
 
         Component {
             id: imageIcon
-            StatusRoundedImage {
+            Item {
                 width: 28
                 height: 28
-                image.source: icon.source
+                StatusRoundedImage {
+                    id: statusRoundImage
+                    width: parent.width
+                    height: parent.height
+                    image.source: icon.source
+                }
+                Loader {
+                    sourceComponent: {
+                        if (statusRoundImage.image.status === Image.Loading) {
+                            return statusLoadingIndicator
+                        }
+                        if (statusRoundImage.image.status === Image.Error) {
+                            return letterIdenticon
+                        }
+                    }
+                    anchors.verticalCenter: parent.verticalCenter
+                    anchors.horizontalCenter: parent.horizontalCenter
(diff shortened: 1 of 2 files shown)


src/StatusQ/Controls/StatusIconTabButton.qml
```diff
@@ -40,10 +40,35 @@ TabButton {
...
+                    image.source: icon.source
+                }
+                Loader {
+                    sourceComponent: {
+                        if (statusRoundImage.image.status === Image.Loading) {
+                            return statusLoadingIndicator
+                        }
+                        if (statusRoundImage.image.status === Image.Error) {
+                            return letterIdenticon
+                        }
```
borismelnik, line 59: By the way, `switch` is faster

borismelnik (approved):

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/9b99d8a95791f90feb45387b7d64728c5d5ad0eb

PR #61, opened 2021-05-21
feat(Components): introduce `StatusRoundIcon` component

### LINKED ISSUE #53: Introduce StatusRoundIcon component
The `StatusListItem` component comes with a variation in which it renders an icon that has a background and basically looks like one of our `StatusRoundButton`s without being an actual button:

Screenshot 2021-05-20 at 13 50 43

We need to create such a component first. Most likely, we can rip parts of `StatusRoundButton` apart and extract those into their own component. That component should then also be reused inside `StatusRoundButton`

PascalPrecht:
This commit introduces a `StatusRoundIcon` component which renders
a `StatusIcon` inside a rounded rectangle.

This is similar to `StatusRoundButton`, just that it isn't a clickable
component.

With these changes, we also introduce a new `StatusIconSettings` config
object that can be used across component that need icon configuration.

The configuration includes the following properties:

- `name` - Name of the icon and used to locate the asset
- `width`
- `height`

Closes #53

### DIFF
--- a/sandbox/Others.qml
+++ b/sandbox/Others.qml
@@ -34,4 +34,8 @@ GridLayout {
     StatusBadge {
         value: 100
     }
+
+    StatusRoundIcon {
+        icon.name: "info"
+    }
 }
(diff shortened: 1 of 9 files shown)


PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/8639e8ccba29cb25eae749262be32a7e3940a34d

PR #67, opened 2021-05-25
fix(Core): disable StatusIcon ColorOverlay if no color is supplied

PascalPrecht:
This is needed to have the underlying SVG's color bleed through and
make use of the image's opacity at the same time.

Using `ColorOverlay` together with (half) transparent colors doesn't
work very well as it makes the underlying SVG color come through,
result in wrong colors.

Applying `opacity` on the image itself cause the `ColorOverlay` to
reduce in opacity as well. So in order to work with transparent
colors, we need a way to turn of the ColorOverlay and work with the
Image's opacity, which is what this change enables.

### DIFF
--- a/src/StatusQ/Core/StatusIcon.qml
+++ b/src/StatusQ/Core/StatusIcon.qml
@@ -3,7 +3,7 @@ import QtGraphicalEffects 1.13
 
 Image {
     property string icon: ""
-    property color color
+    property string color: ""
 
     id: statusIcon
     width: 24
@@ -19,7 +19,7 @@ Image {
     }
 
     ColorOverlay {
-        visible: statusIcon.color !== undefined
+        visible: !!statusIcon.color
         anchors.fill: statusIcon
         source: statusIcon
         color: statusIcon.color
--- a/sandbox/ListItems.qml
+++ b/sandbox/ListItems.qml
@@ -1,0 +1,129 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.14
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Components 0.1
+
+GridLayout {
+    columns: 1
+    columnSpacing: 5
+    rowSpacing: 5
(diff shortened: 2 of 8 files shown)


src/StatusQ/Core/StatusIcon.qml
```diff
@@ -3,7 +3,7 @@ import QtGraphicalEffects 1.13
 
 Image {
     property string icon: ""
-    property color color
+    property string color: ""
```
PascalPrecht, line 6: It seems `property color color` does get a default value that is not `undefined` or `""` even when turned into a string using `toString()`

We need an uninitialized color so changing this to `string` helps.

@borismelnik lemme know if there's a better way to check for `!!color` if `color` is of type `color`

src/StatusQ/Core/StatusIcon.qml
```diff
@@ -19,7 +19,7 @@ Image {
     }
 
     ColorOverlay {
-        visible: statusIcon.color !== undefined
+        visible: !!statusIcon.color
```
borismelnik, line 22: i think this maybe more pretty:
`layer.enabled: !!statusIcon.color
  layer.effect: ColorOverlay {
      color: statusIcon.color
  }`

src/StatusQ/Core/StatusIcon.qml
borismelnik: i think we can use "transparent" color by default and check it with `Qt.colorEqual(color, "transaprent")` bc image cant e invisible)))no ways in docs about it :(

PascalPrecht: That worked perfectly - thanks!!

src/StatusQ/Core/StatusIcon.qml
PascalPrecht: Oh nice, yes, will do that

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/f1e34e39b9f6a6e2987367b1c4c02a3ae458bdb7

PR #68, opened 2021-05-25
feat(Components): introduce StatusChatListItem

### LINKED ISSUE #65: Implement `StatusChatListItem`
With #64 in place, we have  a solid foundation to explore other list item variations that could possibly sit on top of `StatusListItem`. 

One of them is the `StatusChatListItem` which looks like this:
Screenshot 2021-05-21 at 15 51 41

One thing to keep in mind here is that it comes with many different channel specific states:
Screenshot 2021-05-21 at 15 52 39

`StatusListItem` isn't capable of these and it might not make sense to add those to it, which would imply that this component here would become a brand new one.

PascalPrecht:
This introduces the brand new `StatusChatListItem` which can be used
to render channels, one-to-one chats and community channels in Status
applications.

The component can be used as follows:

```qml
import Status.Components 0.1

StatusChatListItem {
    name: "channel-name" // name of channel or user
    type: StatusChatListItem.Type.PublicChat // GroupChat | CommunityChat | OneToOneChat
    image.source: "profile/image/source" // uses letter identicon instead of not supplied
    hasUnreadMessages: true // default `false`
    badge.value: 1 // default `0`
    muted: true // `default `false`
    selected: true // default `false`

    onUnmute: ... // signal when unmute icon is clicked
}
```

Screenshot 2021-05-25 at 12 52 52
Screenshot 2021-05-25 at 12 52 57

Closes #65


src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -0,0 +1,195 @@
...
+                    case StatusChatListItem.Type.GroupChat:
+                        return Theme.palette.name == "light" ? "group" : "group-white"
+                        break;
+                    case StatusChatListItem.Type.CommunityChat:
+                        return Theme.palette.name == "light" ? "channel" : "channel-white"
+                        break;
+                    default:
+                        return Theme.palette.name == "light" ? "public-chat" : "public-chat-white"
+                }
+            }
```
PascalPrecht, line 134: This is rather unfortunate, however I couldn't find a better way to make this work. 

Here's the issue:

When using `StatusIcon` with a `color` with use a `ColorOverlay` to render the supplied `color` over the actual SVG (which comes with its own color [black, most of the time]).

This is problematic when supplying colors with alpha channel, such as the designs foresee here. When the color used for the `ColorOverlay` has transparency, the underlying SVG color with shine through, resulting in undesired colors. 

I tried making the `Image` in `StatusIcon` invisible using `opacity: 0` in case a `color` is supplied (so that only the `ColorOverlay` is rendered) but that makes the overlay invisible as well.

So the solution now is to use in deed 100% black and white icons and apply the `opacity` to `StatusIcon` and disabling the `ColorOverlay` when no `color` is supplied (which is the case here).

Issue #54, opened 2021-05-21
Sand box app pops up half off screen


PascalPrecht:
Recently (I believe since https://github.com/status-im/StatusQ/commit/8737c06ab8ac5752c245b40534295f016aa43f7c was introduced), the sandbox app window is rendered half off screen and no longer in the middle of the screen.

That's no show stopper but a little bit annoying when testing things and the first thing one needs to do is dragging the window back into the center of the screen.

Here's what it looks like right after I ran `./bin/sandbox`:

<img width="1440" alt="Screenshot 2021-05-21 at 10 29 44" src="https://user-images.githubusercontent.com/445106/119107379-77b47600-ba1f-11eb-9b9f-93e1d6ee4112.png">

@borismelnik any idea what could cause this?

Issue #55, opened 2021-05-21
Sandbox app spins up multiple screens after reloading more than ones


PascalPrecht:
When hitting the "Reload QML", there's a new app window created for every push of that button after it was pushed the first time.

PascalPrecht:
This was fixed with https://github.com/status-im/StatusQ/commit/26aae6d027b988c45e2a3652484f9f92825250a9

PR #108, opened 2021-06-01
fix: update position of window to center, add traffic lights

borismelnik:
Closes: #54 #55


PascalPrecht (changes requested):
This looks very good to me!

Only thing, as discussed offline:

Let's introduce a `StatusQ.Platform` module where the window title bar buttons are moved. I think `StatusQ.Components` should be used for components/widgets "inside" applications.

Issue #97, opened 2021-05-28
Implement `StatusModal`


PascalPrecht:
A reusable modal component with configurable toolbar action components (buttons etc) and footer.

Can be broken down into

- `StatusModalToolBar`
- `StatusModalFooter`

API to be proposed.

PascalPrecht:
This has landed with https://github.com/status-im/StatusQ/commit/e49b58b94dac7046b2647e3e6d00fed89ce30108

Issue #99, opened 2021-05-28
Implement `StatusModalFooter`


PascalPrecht:
Part of #97

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/9c2a5830f3d38c16ef905174e88643362e85fd33

Issue #98, opened 2021-05-28
Implement `StatusModalToolBar`


PascalPrecht:
Part of #97 

API to be proposed.

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/fa9bb7adc611d401ea85512326919550694d8103

PR #145, opened 2021-06-14
Feat/statusmodal

borismelnik:
Closes: #97 , #99 , #98


sandbox/Popups.qml
```diff
@@ -0,0 +1,38 @@
...
+
+    StatusModal {
+            id: modalExample
+        anchors.centerIn: parent
+        visible: true
+        header.image.source: "https://pbs.twimg.com/profile_images/1369221718338895873/T_5fny6o_400x400.jpg"
+        header.title: "Header"
+        header.subTitle: "SubTitle"
+        header.detailsButtonSettings.name: "info"
+        showBack: true
```
PascalPrecht, line 23: Do we know for sure that there can only be a "back button" in the left part of the footer? How about otherwise, we keep that open so consumers can decide what kind of button(s) there is?

Could be something like:

```
footer.leftButtons: [
   StatusButton {} 
]
footer.rightButtons: [
  StatusButton {}
]
```

On top of that, but just a nice to have, is to give consumers full control over what the footer even looks like by doing:

```
footer: ...
```
^ no big deal if we do that now.

sandbox/Popups.qml
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    StatusModal {
+            id: modalExample
+        anchors.centerIn: parent
+        visible: true
+        header.image.source: "https://pbs.twimg.com/profile_images/1369221718338895873/T_5fny6o_400x400.jpg"
+        header.title: "Header"
+        header.subTitle: "SubTitle"
+        header.detailsButtonSettings.name: "info"
```
PascalPrecht, line 22: Did you verify with Simon that this can be other icons than `info` ?
Is it always going to trigger "details"

src/StatusQ/Components/StatusImageWithTitle.qml
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        StatusBaseText {
+            id: headerSubTitle
+            font.family: Theme.palette.baseFont.name
+            font.pixelSize: 15
+            color:Theme.palette.baseColor1
+        }
+    }
+}
```
PascalPrecht, line 65: I believe we can make this just part of `StatusModalHeader` without introducing this component. Don't think we'll use the same thing anywhere else.

src/StatusQ/Core/StatusModalHeaderSettings.qml
```diff
@@ -0,0 +1,15 @@
+import QtQuick 2.14
+
+QtObject {
+    property string title
+    property string subTitle
+    property StatusIconSettings detailsButtonSettings: StatusIconSettings {
```
PascalPrecht, line 6: As mentioned above: we need to verify whether this is always about "details"

src/StatusQ/Popups/StatusModal.qml
```diff
@@ -0,0 +1,72 @@
...
+
+    property alias showBack: footerImpl.showBack
+
+    property StatusModalHeaderSettings header: StatusModalHeaderSettings {}
+    property alias buttons: footerImpl.buttons
+
+    signal backClicked()
+    signal butttonClicked(var buttonIndex)
+    signal detailsClicked()
+    signal editClicked()
```
PascalPrecht, line 24: In other StatusQ components we've used the `*ButtonClicked()` convention so far. Maybe we can use that here as well?

src/StatusQ/Popups/StatusModal.qml
```diff
@@ -0,0 +1,72 @@
...
+
+    margins: 0
+    padding: 0
+
+    modal: true
+
+
+    background: Rectangle {
+        color: Theme.palette.indirectColor1
+        radius: 6
```
PascalPrecht, line 39: I might be wrong, but this seems to be `8` in the designs no?

src/StatusQ/Popups/StatusModal.qml
```diff
@@ -0,0 +1,72 @@
...
+            anchors.horizontalCenter: parent.horizontalCenter
+            sourceComponent: statusModal.content
+        }
+
+        Spares.StatusModalFooter {
+            id: footerImpl
+            width: 480
+
+            onBack: statusModal.backClicked()
+            onClicked: statusModal.butttonClicked(buttonIndex)
```
PascalPrecht, line 69: So, if we keep the API open, that people can do:

```
(footer).(right)Buttons: [
  StatusButton { ... }
]
```

People can just attach `onClicked` handlers directly on their buttons, so there's no need to emit  an `buttonClicked(index)` event

src/StatusQ/Popups/statusModal/StatusModalFooter.qml
```diff
@@ -0,0 +1,67 @@
...
+
+    property bool showBack: true
+
+    property list<StatusBaseButton> buttons
+
+
+    signal clicked(var buttonIndex)
+    signal back
+
+    radius: 6
```
PascalPrecht, line 21: Should be `8` I guess

src/StatusQ/Popups/statusModal/StatusModalHeader.qml
```diff
@@ -0,0 +1,83 @@
...
+        id: detailsButton
+
+        anchors.right: closeButton.left
+        anchors.rightMargin: 24.5
+        anchors.top: parent.top
+        anchors.topMargin: 18
+
+        type: StatusFlatRoundButton.Type.Secondary
+        width: 20
+        height: 20
```
PascalPrecht, line 52: This seems wrong.  Notice how the hover isn't properly work for this button

src/StatusQ/Popups/statusModal/StatusModalHeader.qml
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    StatusFlatRoundButton {
+        id: closeButton
+        anchors.right: parent.right
+        anchors.rightMargin: 22.5
+        anchors.top: parent.top
+        anchors.topMargin: 22.5
+        width: 11.5
+        height: 11.5
```
PascalPrecht, line 69: Same here, the button is bigger than the icon.

src/assets/img/icons/arrow-left.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="22" height="16" viewBox="0 0 22 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M9.28547 1.61872C9.62717 1.27701 9.62717 0.72299 9.28547 0.381282C8.94376 0.0395728 8.38974 0.0395728 8.04803 0.381282L1.04803 7.38128C0.706321 7.72299 0.706321 8.27701 1.04803 8.61872L8.04803 15.6187C8.38974 15.9604 8.94376 15.9604 9.28547 15.6187C9.62717 15.277 9.62717 14.723 9.28547 14.3813L4.775 9.87081C4.40752 9.50333 4.66778 8.875 5.18748 8.875H20.3334C20.8167 8.875 21.2084 8.48325 21.2084 8C21.2084 7.51675 20.8167 7.125 20.3334 7.125H5.18748C4.66778 7.125 4.40752 6.49667 4.775 6.12919L9.28547 1.61872Z" fill="#88B0FF"/>
+</svg>
```
PascalPrecht, line 3: We have `arrow-right` which can use `icon.rotation: 180`

PascalPrecht (changes requested):
Great works!

There are some small things that need adjustment and probably clarification :)

sandbox/Popups.qml
borismelnik: i just call it details bc i dont know how to call it :) i have some naming problem, if you can give me a great name for this button -  i will be grateful. i will verify.

src/StatusQ/Components/StatusImageWithTitle.qml
borismelnik: but i thinking than we talk about it as independent component :D okay i will fix

src/StatusQ/Popups/StatusModal.qml
borismelnik: sorry, dont understand.

src/StatusQ/Popups/statusModal/StatusModalFooter.qml
borismelnik: right

sandbox/Popups.qml
PascalPrecht: Sure, we can come up with a name, but first we have to know if it's always going to be an "info" icon or if it was just a placeholder in the design

src/StatusQ/Components/StatusImageWithTitle.qml
PascalPrecht: Yea that's true, we talked about this. I thought we might need a similar thing in the `StatusChatInfoToolBar` but we didn't.. Sorry for the inconvenience

src/StatusQ/Popups/StatusModal.qml
PascalPrecht: So for example instead of calling it `editClicked` call it `editButtonClicked`. 

We might not need `buttonClicked` if we keep the buttons fully customizable, because then users can apply click handlers on each button themselves

PascalPrecht: See: https://github.com/status-im/StatusQ/pull/145#discussion_r650936165

src/StatusQ/Popups/StatusModal.qml
PascalPrecht: I think this still needs to be 8

src/StatusQ/Components/StatusImageWithTitle.qml
PascalPrecht: This is still a single component?

src/StatusQ/Popups/StatusModal.qml
PascalPrecht: This still needs to be `8`

src/StatusQ/Popups/StatusModal.qml
```diff
@@ -0,0 +1,66 @@
...
+        Spares.StatusModalHeader {
+            id: headerImpl
+            width: 480
+
+            title: header.title
+            subTitle: header.subTitle
+            image: header.image
+
+            onEditButtonClicked: statusModal.editButtonClicked()
+            onDetailsButtonClicked: statusModal.detailsButtonClicked()
```
PascalPrecht, line 51: There's no such event on this component

src/StatusQ/Popups/statusModal/StatusModalHeader.qml
```diff
@@ -0,0 +1,69 @@
...
+    color: Theme.palette.indirectColor1
+
+
+    StatusImageWithTitle {
+        id: imageWithTitle
+        anchors.verticalCenter: parent.verticalCenter
+        anchors.left: parent.left
+        anchors.leftMargin: 16
+
+        onEditClicked: statusModalHeader.editClicked()
```
PascalPrecht, line 35: This needs to be `editButtonClicked`

src/StatusQ/Popups/statusModal/StatusModalHeader.qml
```diff
@@ -0,0 +1,69 @@
...
+
+    property alias title: imageWithTitle.title
+    property alias subTitle: imageWithTitle.subTitle
+    property alias otherButton: otherButtonLoader.sourceComponent
+
+    property alias image: imageWithTitle.image
+
+    signal editButtonClicked
+    signal close
+    signal detailsButtonClicked
```
PascalPrecht, line 19: This is probably obsolete by now?

src/StatusQ/Popups/statusModal/StatusModalHeader.qml
```diff
@@ -0,0 +1,69 @@
...
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+import StatusQ.Controls 0.1
+
+Rectangle {
+    id: statusModalHeader
+
+    property alias title: imageWithTitle.title
+    property alias subTitle: imageWithTitle.subTitle
+    property alias otherButton: otherButtonLoader.sourceComponent
```
PascalPrecht, line 13: Let's call this `actionButton` instead

PascalPrecht (commented):
Great work, thanks for the update!

Unfortunately, I still found some bugs in there (e.g. the edit button isn't working and there are some left over signals well).

I'm gonna go ahead and apply the needed changes to your commits and merge it into `master` then so we can finally close this :)

PascalPrecht:
These have landed as https://github.com/status-im/StatusQ/commit/fa9bb7adc611d401ea85512326919550694d8103, https://github.com/status-im/StatusQ/commit/9c2a5830f3d38c16ef905174e88643362e85fd33 and https://github.com/status-im/StatusQ/commit/e49b58b94dac7046b2647e3e6d00fed89ce30108

I've also added https://github.com/status-im/StatusQ/commit/148c30b991b93172246b039ccc7db59c29216581

PR #159, opened 2021-06-15
fix(StatusChatListCategoryItem): don't render menubutton with no popup

### LINKED ISSUE #153: `StatusChatListCategory` menu button breaks when no popup menu is provided
This is something that came up while building #133 

Even though, this might not necessarily happen in practice, when a `StatusChatListCategory` is used and doesn't have a `popupMenu` for it's `menuButton`, the underlying `StatusChatListCategoryItem` breaks the menu button's state and also throws and error, because it tries to open a popup menu regardless.

The solution here is to either return early in the menu button trigger when there's no popup menu, or allow to individually hide the menu button (notice that the add button isn't affected by this), similar to #152

PascalPrecht:
Fixes #153

### DIFF
--- a/sandbox/DemoApp.qml
+++ b/sandbox/DemoApp.qml
@@ -157,23 +157,18 @@ Rectangle {
 
         StatusAppTwoPanelLayout {
 
-            leftPanel: Item {
+            leftPanel: StatusChatListAndCategories {
                 anchors.fill: parent
+                anchors.topMargin: 64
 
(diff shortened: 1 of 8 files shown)


borismelnik (approved):

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/05fc97ca2fbbc62ca3f40c60a01c1c1247715375

Issue #106, opened 2021-05-31
Introduce `StatusBaseInput`


PascalPrecht:
An input component with Status look & feel in it's most simple form. Can then be used to build more complex input comnponents.

PR #199, opened 2021-06-28
Introducing StatusBaseInput

borismelnik:
Usage: The same interface like TextField

Closes: #106

### DIFF
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -1,0 +1,146 @@
+import QtQuick 2.14
+
+import QtQuick.Controls 2.14 as QC
+
+import StatusQ.Controls 0.1
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+
+Item {
+    id: statusBaseInput
+
+    property bool multiline: false
+
+    property bool clearable: false
+
+    property alias inputMethodHints: edit.inputMethodHints
+
+    property alias selectedText: edit.selectedText
+    property alias selectedTextColor: edit.selectedTextColor
+    property alias selectionStart: edit.selectionStart
+    property alias selectionEnd: edit.selectionEnd
+
+    property alias color: edit.color
+    property alias font: edit.font
+    property alias verticalAlignmet: edit.verticalAlignment
+    property alias horizontalAlignment: edit.horizontalAlignment
+
+    property alias placeholderText: placeholder.text
+    property alias placeholderTextColor: placeholder.color
+    property alias placeholderFont: placeholder.font
+
+    property real leftPadding: 16
+    property real rightPadding: 16
+    property real topPadding: 11
+    property real bottomPadding: 11
+
+    property real minimumHeight: 0
+    property real maximumHeight: 0
+
+    implicitWidth: 200
+    implicitHeight: multiline ? Math.max(edit.implicitHeight + topPadding + bottomPadding, 40) : 40
+
+    Rectangle {
+        width: parent.width
+        height: maximumHeight != 0 ? Math.min(
+                                        minimumHeight != 0 ? Math.max(statusBaseInput.implicitHeight, minimumHeight)
+                                                          : implicitHeight,
+                                         maximumHeight)
+                                  : parent.height
+        color: Theme.palette.baseColor2
+        radius: 8
+
+        clip: true
+
+        Flickable {
+            id: flick
+
+            anchors.fill: parent
+            anchors.leftMargin: statusBaseInput.leftPadding
+            anchors.rightMargin: statusBaseInput.rightPadding + clearable ? clearButtton.width : 0
+            anchors.topMargin: statusBaseInput.topPadding
+            anchors.bottomMargin: statusBaseInput.bottomPadding
+            contentWidth: edit.paintedWidth
+            contentHeight: edit.paintedHeight
+            clip: true
+
+            QC.ScrollBar.vertical: QC.ScrollBar { interactive: multiline }
+
+
+            function ensureVisible(r) {
+                if (contentX >= r.x)
+                    contentX = r.x;
+                else if (contentX+width <= r.x+r.width)
+                    contentX = r.x+r.width-width;
+                if (contentY >= r.y)
+                    contentY = r.y;
+                else if (contentY+height <= r.y+r.height)
+                    contentY = r.y+r.height-height;
+            }
+
+
+
+            TextEdit {
+                id: edit
+                width: flick.width
+                selectByMouse: true
+                anchors.verticalCenter: parent.verticalCenter
+                focus: true
+
+                font.pixelSize: 15
+                font.family: Theme.palette.baseFont.name
+                color: Theme.palette.directColor1
+
+                onCursorRectangleChanged: flick.ensureVisible(cursorRectangle)
+                wrapMode: TextEdit.NoWrap
+
+                Keys.onReturnPressed: {
+                    if (multiline) {
+                        event.accepted = false
+                    } else {
+                        event.accepted = true
+                    }
+                }
+
+                Keys.onEnterPressed: {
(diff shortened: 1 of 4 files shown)


src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -0,0 +1,113 @@
...
+        id: flick
+
+        anchors.fill: parent
+        anchors.leftMargin: statusBaseInput.leftPadding
+        anchors.rightMargin: statusBaseInput.rightPadding + clearButtton.width
+        anchors.topMargin: statusBaseInput.topPadding
+        anchors.bottomMargin: statusBaseInput.bottomPadding
+        contentWidth: edit.paintedWidth
+        contentHeight: edit.paintedHeight
+        clip: true
```
PascalPrecht, line 53: Couple of things here:

1. Is it possible to turn this off from the outside? Many of the inputs are just single line inputs (search input, password input, etc). Having the base input being scrollable/flickable by default seems off. The default should probably be that it's just a single line input. We can even split this up into separate components if needed (e.g. StatusBaseInput, StatusBaseTextArea).

2. The whole flickable thing doesn't seem to work well. When entering information and hitting SHIFT+Enter and then continuing typing, this is what it looks like:

Screenshot 2021-06-29 at 09 59 29

Notice how the entered text isn't fully visible, users can't see the cursor anymore. either if they just add a few more lines.

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -0,0 +1,113 @@
...
+        type: StatusFlatRoundButton.Type.Secondary
+        width: 14
+        height: 14
+        icon.name: "clear"
+        icon.width: 14
+        icon.height: 14
+        onClicked: {
+            edit.clear()
+        }
+    }
```
PascalPrecht, line 111: This clear button is nice, but I don't think we want this to always show in the status base input. Again, the base input should just be a simple input that can then be used to build more complex inputs.

I'm fine with making this configurable though. We could have a flag `clearable` which will render that button.

Also, we should double check the color of the clear button. Did you take it from any designed component?

PascalPrecht (changes requested):
This is a good start @borismelnik !

I've left some comments, but in a nutshell:

- Flickable doesn't work well and should be turned off by default
- Clear button should be turned off by default as well

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: This seems to behave much better now.
One thing however, is still unexpected:

If it's `multiline: true` and I enter some value it looks like this:

Screenshot 2021-07-05 at 17 41 10

When I then hit Shift+ Enter (or just Enter) to get to the next line, it looks like this:

Screenshot 2021-07-05 at 17 42 01

Basically the first line I've entered disappears because the input value only renders the last line (**this might be expected**).

However, if I do the same thing again, all of a sudden the input starts to grow:
Screenshot 2021-07-05 at 17 43 02

This seems less unexpected to me, but I believe we should be doing either one or the other behaviour (either the height is fixed and text value becomes scrollable, or it always grows [unless a fixed height was supplied from the outside]).

What do you think?

PascalPrecht (commented):
Great update @borismelnik !

src/StatusQ/Controls/StatusBaseInput.qml
borismelnik: sorry that is size bug.

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -0,0 +1,132 @@
...
+
+    clip: true
+
+    Rectangle {
+        anchors.fill: parent
+        color: Theme.palette.baseColor2
+        radius: 8
+    }
+
+    Flickable {
```
pinebit, line 47: we should probably add ScrollBars since we have set `wrapMode: TextEdit.NoWrap`.

src/StatusQ/Controls/StatusBaseInput.qml
pinebit: I support this:
> We can even split this up into separate components if needed (e.g. StatusBaseInput, StatusBaseTextArea).
Based on my prior experience it is the best approach.

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -0,0 +1,132 @@
+import QtQuick 2.14
+
+import StatusQ.Controls 0.1
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+
+Item {
+    id: statusBaseInput
```
pinebit, line 9: as per QML guidelines, it is recommended to name the root entity as "root".

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -0,0 +1,132 @@
...
+
+    property bool multiline: false
+
+    property bool clearable: false
+
+    property alias inputMethodHints: edit.inputMethodHints
+
+    property alias selectedText: edit.selectedText
```
pinebit, line 17: If you would split this control into TextField and TextArea, you could derive from them explicitly and customize the look and feel. In this case you don't need to alias all the properties.

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: We've had this discussion before, in fact most of the components used `root` but eventually we decided to go for component names. Lemme see if I can find that discussion

PascalPrecht: Can't find it anymore, it's somewhere in one of the PRs. Anyways, I think @borismelnik mentioned that not using `root` is preferable.

I personally don't have a strong opinion about that. Question is whether we gain anything from renaming all of them again (other than increasing chances to introduce bugs)

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Yes, I was preferring having a `StatusBaseInput` and something like `StatusTextArea` as well. @borismelnik mentioned something about some `QDocument` or something that we'd need and we'll get access to with the text component we use in this PR.

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Happy to revisit this. For now though, I'd like us to land **some** input component that really just renders an input because so many other components depend on this.

src/StatusQ/Controls/StatusBaseInput.qml
borismelnik: i never see this recommendation - could you pls give me a link?

borismelnik: i have this one https://doc.qt.io/qt-5/qml-codingconventions.html

pinebit: https://github.com/Furkanzmc/QML-Coding-Guide#cs-5-giving-components-ids

> Make sure that the top most component in the file always has root as its id. Qt will make unqualified name look up deprecated in QML 3, so it's better to start giving IDs to your components now and use qualified look up.

PascalPrecht:
Thanks for updating this again @borismelnik !
This has landed as https://github.com/status-im/StatusQ/commit/1321760442c2e1d5ec4c2e66c599ede441460fc8

I've found another usability issue with this though: When the input is made like a textarea, even though it looks like a text area, only the very first line of the input is clickable. This needs to be fixed, or we should really just introduce a `StatusTextArea` component that uses `TextArea` which gives us the right behaviour by default.

Will create an issue for that.

PascalPrecht:
Created an issue here https://github.com/status-im/StatusQ/issues/241 /cc @borismelnik

PR #232, opened 2021-07-06
fix(StatusChatInfoButton): disable hover effects when sensor is disabled

### LINKED ISSUE #231: Disable `StatusChatInfoButton` hover effects when `!sensor.enabled`
There are scenarios in which this component is used and the sensor is turned off. 
In those cases there's no need to show a hover effect and changing the cursor style.

PascalPrecht:
Closes #231


PascalPrecht:

PR #234, opened 2021-07-06
fix(StatusChatInfoButton): prefix chat name with "#" if needed

### LINKED ISSUE #229: `StatusChatInfoButton` should prefix chatname with `#` if needed
This is basically the same as https://github.com/status-im/StatusQ/issues/191

PascalPrecht:
Closes #229


PascalPrecht:

PR #235, opened 2021-07-06
fix(StatusChatInfoButton): make component identicon aware

### LINKED ISSUE #228: `StatusChatInfoButton` needs to take `image.isIdenticon` into account
Similar to what we've done in `StatusChatListItem`, the chat info button component needs to figure out whether the given image source is an identicon and use that information to render a border around the image.

PascalPrecht:
Closes #228


PascalPrecht:

PR #253, opened 2021-07-09
feat(StatusChatToolBar): add tooltips to action buttons

### LINKED ISSUE #244: `StatusChatToolBar` buttons need tooltip
With more action buttons being added as of #243 , buttons should now receive a tooltip as well.

We have a `StatusToolTip` component for that which should be added here for each button:

https://github.com/status-im/StatusQ/blob/22eaf6fa997caaf4799bef2f7c81ac9d89237895/src/StatusQ/Components/StatusChatToolBar.qml#L39-L105

The tool tip text can/should default to the following: 
Screenshot 2021-07-08 at 12 07 47

Apps that use StatusQ however, will most likely have them i18n'ed. They will have to override it accordingly via

```qml
StatusChatToolBar {
  ...
  menuButton.tooltip.text: qsTr("More")
}
```

PascalPrecht:
Closes #244

### DIFF
--- a/sandbox/DemoApp.qml
+++ b/sandbox/DemoApp.qml
@@ -255,6 +255,8 @@ Rectangle {
                     chatInfoButton.type: StatusChatInfoButton.Type.OneToOneChat
                     chatInfoButton.pinnedMessagesCount: 1
 
+                    searchButton.visible: false
+                    membersButton.visible: false
                     notificationCount: 1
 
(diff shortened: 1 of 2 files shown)


PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/86da901e8875cd595a25b95185fba640a82eef34

Issue #259, opened 2021-07-15
Remove visual state for unread messages in `StatusChatListitem` when muted


PascalPrecht:
Right now, whenever there's unread messages in a chat item, the chat item title is `font.weight: bold`. This should not be the case when the item is `muted: true`

PascalPrecht:
This has landed with #260

PR #260, opened 2021-07-15
feat(StatusChatListItem): introduce muted badge visuals

### LINKED ISSUE #258: `StatusChatListItem` needs variation with badge in muted state
This came up after using the new status chat list for a while and realising that muted chat list items don't indicate well when there's mentions in the dedicated chat.

Simon came up with a new design that looks something like this:
Screenshot 2021-07-15 at 13 49 08

`StatusBadge` most likely need a new variation for that. Or we just enforce it's color in that case..

PascalPrecht:
Also ensure title font weight stays `normal` when item is `muted`.

Closes #258, #259

### DIFF
--- a/sandbox/ListItems.qml
+++ b/sandbox/ListItems.qml
@@ -96,16 +96,43 @@ GridLayout {
     }
 
     StatusChatListItem {
+        name: "muted-with-mentions"
+        type: StatusChatListItem.Type.PublicChat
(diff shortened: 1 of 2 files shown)


PascalPrecht:
Screenshot 2021-07-15 at 14 12 04
Screenshot 2021-07-15 at 14 12 15

PR #262, opened 2021-07-15
feat(StatusListItem): add identicon support

### LINKED ISSUE #261: `StatusListItem` image/icon needs to provide API to change its border
Originally, there were no cases in the `StatusListItem` designs that required an image, let alone an image with a border, but only a rounded icon.

We're now using `StatusListItem` to render contact items which may end up having identicons and those need a border. There's already similar mechanisms in place in `StatusChatListItem` and others.

PascalPrecht:
Closes #261

### DIFF
--- a/sandbox/ListItems.qml
+++ b/sandbox/ListItems.qml
@@ -122,11 +122,18 @@ GridLayout {
         icon.name: "info"
     }
 
-
     StatusListItem {
         title: "Title"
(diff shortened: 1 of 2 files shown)


PascalPrecht:
Screenshot 2021-07-15 at 14 59 27
Screenshot 2021-07-15 at 14 59 21

PR #268, opened 2021-07-16
fix(StatusChatInfoToolBar): provide control about action visibility

alaibe:


PascalPrecht (commented):
@alaibe the changes look fine to me!

If you don't mind however, I'd split these things up, because we're seeing a bunch of things here:

1. A new API being exposed in `StatusChatInfoToolBar` (which is a feature)
2. A new API on `StatusChatListAndCategories` which is a feature as well
3. A fix in `StatusChatListCategory` that ensures `showActionButtons` is taken into account

I think these are important changes and shouldn't get lost, so it would be wonderful if they'll show up in our change log (https://github.com/status-im/StatusQ/blob/master/CHANGELOG.md) once released

PR #283, opened 2021-07-22
feat(StatusQ.Layout): introducing StatusAppThreePanelLayout

### LINKED ISSUE #272: Implement `StatusAppThreePanelLayout`
So far, most of the views in Status Desktop came with a two panel layout e.g., chat list on the left panel, chats messages on the right panel. Or, menu items on the left panels, content sections on the right panel.

With the latest "User list" feature, which renders a list of users that joined the currently active chat, there's a third column added to the layout:
Screenshot 2021-07-20 at 10 12 54

To help Status Desktop implement this feature, StatusQ needs to provide a `StatusAppThreePanelLayout` component, analogous to `StatusAppTwoPanelLayout`.

The two panel layout comes with an API that looks like this:

```qml
StatusAppTwoPanelLayout {

  leftPanel: ...
  rightPanel: ...
}
```
Where both panels provide their implicit dimensions. Resizing panels isn't supported yet (#134)

Similarly,  a `StatusAppThreePanelLayout` could have an API that looks like this:

```qml
StatusAppThreePanelLayout {
  leftPanel: ...
  midPanel: ...
  rightPanel: ...
}
```

A more concrete example of `StatusAppTwoPanelLayout` can be found here:

https://github.com/status-im/StatusQ/blob/51a345866a9b31577e76aa97112990d9501844d3/sandbox/main.qml#L100-L204

alexandraB99:
Added new component to support a 3 column view      

Closes #272

### DIFF
--- a/src/StatusQ/Layout/StatusAppThreePanelLayout.qml
+++ b/src/StatusQ/Layout/StatusAppThreePanelLayout.qml
@@ -1,0 +1,51 @@
+import QtQuick 2.13
+import QtQuick.Controls 2.13
+import StatusQ.Core.Theme 0.1
+
+SplitView {
+    id: root
+
+    implicitWidth: 822
+    implicitHeight: 600
+
+    handle: Item { }
+
+    property Item leftPanel
+    property Item centerPanel
+    property Component rightPanel
+    property bool showRightPanel
+
+    Control {
+        SplitView.preferredWidth: 300
+        SplitView.fillHeight: true
+        contentItem: (!!leftPanel) ? leftPanel : null
+        background: Rectangle {
+            anchors.fill: parent
+            color: Theme.palette.baseColor4
+        }
+    }
+
+    Control {
+        SplitView.fillWidth: true
+        SplitView.fillHeight: true
+        contentItem: (!!centerPanel) ? centerPanel : null
+        background: Rectangle {
+            anchors.fill: parent
+            color: Theme.palette.statusAppLayout.rightPanelBackgroundColor
+        }
+    }
+
+    Control {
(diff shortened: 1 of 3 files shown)


src/StatusQ/Layout/StatusAppThreePanelLayout.qml
```diff
@@ -0,0 +1,55 @@
...
+        background: Rectangle {
+            anchors.fill: parent
+            color: Theme.palette.statusAppLayout.rightPanelBackgroundColor
+        }
+    }
+
+    Control {
+        SplitView.preferredWidth: root.showRightPanel ? 250 : 0
+        SplitView.minimumWidth: root.showRightPanel ? 50 : 0
+        Behavior on SplitView.preferredWidth { NumberAnimation {}}
```
PascalPrecht, line 41: Let's remove animations for now until we know exactly what this should look like.

PascalPrecht (changes requested):
Very nicely done @alexandraB99 !

I just left a minor comment there. We don't know yet what the transition should be like, so for now I think it's better if it doesn't animate by default.

Also really really cool that you've updated the `DemoApp`.

Would you mind updating your commit message to follow the conventions of this repo?
Something like:

```
feat(StatusQ.Layout): introduce StatusAppThreePanelLayout 

<body>
```

should do the trick.

PascalPrecht (approved):

PascalPrecht:
Wonderful.

src/StatusQ/Layout/StatusAppThreePanelLayout.qml
alexandraB99: @PascalPrecht all done! :)

PR #292, opened 2021-07-23
Epic: `StatusInput`

PascalPrecht:
This PR adds a bunch of commits that introduce the following:

- Hover and disabled state for `StatusBaseInput`
- Validity state for `StatusBaseInput`
- It introduces `StatusInput` which wraps `StatusBaseInput` for further composition
- Error message, label and charlimit APIs

Usage:

```qml
StatusInput {
  label: "FieldName"
  charLimit: 30
  errorMessage: "Some error occured"

  input.valid: false
  input.text: "Some invalid value"
  input.placeholderText: "Some placeholder"
  input.clearable: true
}
```


PascalPrecht:
Screenshot 2021-07-23 at 13 24 18
Screenshot 2021-07-23 at 13 24 12

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -0,0 +1,85 @@
...
+                  charLimitLabel.height :
+              0) +
+            statusBaseInput.anchors.topMargin +
+            statusBaseInput.height + 
+            (errorMessage.visible ? 
+                  errorMessage.anchors.topMargin +
+                  errorMessage.height :
+                  0) + 8
+
+    color: "transparent"
```
pinebit, line 25: if the color is always transparent, then why not make root an `Item`, not `Rectange`?

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    StatusBaseText {
+        id: charLimitLabel
+        height: visible ? implicitHeight : 0
+        anchors.top: parent.top
+        anchors.right: parent.right
+        anchors.topMargin: visible ? 11 : 0
+        anchors.rightMargin: 16
+        visible: !!root.charLimit
```
pinebit, line 52: what if `charLimit` is a negative value?

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -0,0 +1,85 @@
...
+    StatusBaseText {
+        id: label
+        height: visible ? implicitHeight : 0
+        anchors.top: parent.top
+        anchors.left: parent.left
+        anchors.topMargin: visible ? 8 : 0
+        anchors.leftMargin: 16
+        visible: !!root.label
+
+        text: root.label
```
pinebit, line 40: is elide enabled by default? shall we set the max width for the text then?

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -0,0 +1,85 @@
...
+        anchors.topMargin: 11
+        anchors.right: parent.right
+        anchors.rightMargin: 16
+
+        height: visible ? implicitHeight : 0
+        visible: !!root.errorMessage && !statusBaseInput.valid
+
+        font.pixelSize: 12
+        color: Theme.palette.dangerColor1
+        text: root.errorMessage
```
pinebit, line 83: same as for label - need to make sure the message will be elided.

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: Good catch. This is a leftover from debugging. Whoopsie

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: In that case it'd be truthy. Could be considered a bug. Will update

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: Yea wouldn't hurt! Although, in practice I think chances are very low we'd exceed the width

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: In this case, it's probably better if it wraps. And elided error message could be less useful

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -0,0 +1,85 @@
...
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Controls 0.1
+
+Rectangle {
+    id: root
+
+    implicitWidth: 480
+    height: (label.visible ?
```
pinebit, line 11: Qt Layout system does all this math for you if you were using `ColumnLayout` instead of anchors.
```
FocusScope {
    id: root
    implicitHeight: layout.childrenRect.height

    ColumnLayout {
        id: layout
        
        StatusBaseText {
            Layout.fillWidth: true // for elide to work
        }
        
        StatusBaseInput {
            Layout.fillWidth: true
        } 
        ... etc
    }
}
```

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -54,91 +56,107 @@ Item {
...
-        }
     } // Rectangle
 
     StatusFlatRoundButton {
         id: clearButtton
-        visible: edit.text.length != 0 && statusBaseInput.clearable && !statusBaseInput.multiline
+        visible: edit.text.length != 0 && 
+            statusBaseInput.clearable && 
+            !statusBaseInput.multiline &&
+            edit.activeFocus
```
pinebit, line 159: you probably now better our UX guidelines, but if the field is not empty and has no active focus, then maybe I should still see the clear button? in order to quickly clear it by mouse click? otherwise I need to clicks: one to set focus, the second is to click on this button.

PascalPrecht: Looping in @simonricoo to get his view on this. Design system shows clear button should only be rendered when active focus

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: Ah, that's cool! I'm very bad at this layout stuff haha

src/StatusQ/Controls/StatusBaseInput.qml
pinebit: OK. Also I just tested how this works and found this button too small and too hard to point and click. I would wish it to have larger hotspot area.

src/StatusQ/Controls/StatusInput.qml
pinebit: Agreed

src/StatusQ/Controls/StatusInput.qml
pinebit: In my experience it started to overflow when your app gets translated to other languages... translators (humans) will never consider allowed boundaries.

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -54,91 +56,107 @@ Item {
...
-            TextEdit {
-                id: edit
-                width: flick.width
-                selectByMouse: true
-                anchors.verticalCenter: parent.verticalCenter
-                focus: true
+                TextEdit {
+                    id: edit
+                    width: flick.width
+                    selectByMouse: true
```
pinebit, line 106: I just tested your code and found that I cannot right click and do Cut/Copy/Paste (context menu).
Is this something you will be introducing later?

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -0,0 +1,85 @@
...
+    StatusBaseText {
+        id: charLimitLabel
+        height: visible ? implicitHeight : 0
+        anchors.top: parent.top
+        anchors.right: parent.right
+        anchors.topMargin: visible ? 11 : 0
+        anchors.rightMargin: 16
+        visible: !!root.charLimit
+
+        text: "0 / " + root.charLimit
```
pinebit, line 54: probably not implemented yet, because the characters counter is not updated.
`"%1 / %2".arg(statusBaseInput.text).arg(root.charLimit)`
also it looks like the charLimit is not enforced on `statusBaseInput`.

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: Yeap, very good argument. Updating this.

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Created an issue for this as it's outside of the scope of this PR: https://github.com/status-im/StatusQ/issues/294

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Yea something that I haven't even thought about to be honest. Wasn't area that built-in/native right click menus don't work here. cc'ing @simonricoo to potentially update the design system.

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: > probably not implemented yet, because the characters counter is not updated.
"%1 / %2".arg(statusBaseInput.text.length).arg(root.charLimit)

Totally missed this. Thanks for the quick snippet

The enforcement should be done via a validator. What "enforcement" means though, depends on the application. I could imagine that enforcement of char limit isn't needed, but setting the validity of this input is, so that other controls can be disabled accordingly (like submit buttons).

Either way, validator pipeline still needs to be built.

PR #297, opened 2021-07-26
fix(StatusAppThreePanelLayout): limit right panel width to 300px

alexandraB99:
Also added handle in DemoApp as well as hiding text when in minimum
width for right and left panels

### DIFF
--- a/sandbox/DemoApp.qml
+++ b/sandbox/DemoApp.qml
@@ -408,6 +408,12 @@ Rectangle {
 
         StatusAppThreePanelLayout {
             id: root
+
+            handle: Rectangle {
+                implicitWidth: 5
+                color: SplitHandle.pressed ? Theme.palette.baseColor2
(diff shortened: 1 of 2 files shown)


PascalPrecht:
There's already this https://github.com/status-im/StatusQ/commit/762ff87bcc1aa14ad67b444b6314447c61a1db64 which claims to do the same thing, but clearly this PR has more changes to it.

Anything that should be noted?

alexandraB99:
> There's already this [762ff87](https://github.com/status-im/StatusQ/commit/762ff87bcc1aa14ad67b444b6314447c61a1db64) which claims to do the same thing, but clearly this PR has more changes to it.
> 
> Anything that should be noted?

That one refers to center panel "limit center panel width to 300px", this one is for the right panel. Plus "Also added handle in DemoApp as well as hiding text when in minimum
width for right and left panels"

PascalPrecht:
The devil is in the details. I should read more carefully!

PR #300, opened 2021-07-28
feat(Drag&Drop): Add drag and drop functionality on StatusChatListItem

borismelnik:


PascalPrecht:
I haven't reviewed the code yet, but I checked out the PR to see how it works. Unfortunately this isn't working well.
Here's what I ran into:

1.  Dragged items don't snap properly into position. I doubt we want items to be positioned anywhere. Below is a screenshot of how I was able to arrange some of the list items:
Screenshot 2021-07-28 at 13 16 59

2. It's even worse with categories:
Screenshot 2021-07-28 at 13 18 39

3. It doesn't seem like dragging into other categories is possible (is that intended?)

iurimatias (changes requested):
doesn't work for me see video


https://user-images.githubusercontent.com/176720/128252394-680143c9-c309-4d77-9a97-e3c5f27cb705.mov

src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -207,6 +232,93 @@ Rectangle {
...
+                    }
+
+                    StatusBaseText {
+                        anchors.verticalCenter: parent.verticalCenter
+                        width: cover.width - iconLoader.item.width - 22
+                        text: chatName.text
+                        elide: Text.ElideRight
+                    }
+                }
+            }
```
PascalPrecht, line 308: This is problematic. By rebuilding the item component this way, we're losing a lot of its properties. The result is that e.g. the font weight changes when it's dragged, or that we lose various badges and icons.

Can we make sure that, when a chat list item is dragged, it doesn't lose its properties?

PascalPrecht (changes requested):

PascalPrecht:
One more thing:

As @iurimatias pointed out, this is not working well in the categories view. Can we introduce a flag to disable drag support by default? That way we don't break existing UIs when this lands and StatusQ dependency is updated.

Status Desktop would then have to explicitely turn on drag support in the chat view (therefore not affecting the communities view with categories).

It works great for single chat lists though, so I think once the comments are ironed out, we can land this!

Oh, and please update your commit message to something like

```
feat(StatusChatList): introduce drag & drop support
```

PascalPrecht:
Closing this in favour of https://github.com/status-im/StatusQ/pull/339

@borismelnik I kept you as a committer, but I've moved some things around. Namely, the `DropArea` and `Drag` capabilities I've moved into `StatusChatList`. Two reasons why I've done this:

1. Semantically, chat items are drag'n'droppable inside of a list, not as standalone items
2. We actually need to reuse entire `StatusChatListItem` components for the drag effect, so drag and drop capabilities need to be decoupled from that component.

PR #301, opened 2021-07-28
feat(StatusQ.Popups) Adding SearchPopup component

### LINKED ISSUE #264: Full Search Results component
note: status-desktop contains the messages portion of this component which could potentially be moved here

Screen Shot 2021-07-15 at 1 21 40 PM

https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/💬-Chat⎜Desktop?node-id=6042%3A1

alexandraB99:
Closes #264

### DIFF
--- a/sandbox/DemoApp.qml
+++ b/sandbox/DemoApp.qml
@@ -299,14 +299,14 @@ Rectangle {
                     StatusChatList {
                         anchors.horizontalCenter: parent.horizontalCenter
 
-                        chatListItems.model: demoChatListItems
+                        chatListItems.model: models.demoChatListItems
                         selectedChatId: "0"
                         onChatItemSelected: selectedChatId = id
                         onChatItemUnmuted: {
-                            for (var i = 0; i < demoChatListItems.count; i++) {
-                                let item = demoChatListItems.get(i);
+                            for (var i = 0; i < models.demoChatListItems.count; i++) {
+                                let item = models.demoChatListItems.get(i);
                                 if (item.chatId === id) {
-                                    demoChatListItems.setProperty(i, "muted", false)
+                                    models.demoChatListItems.setProperty(i, "muted", false)
                                 }
                             }
                         }
@@ -412,7 +412,7 @@ Rectangle {
             handle: Rectangle {
                 implicitWidth: 5
                 color: SplitHandle.pressed ? Theme.palette.baseColor2
-                    : (SplitHandle.hovered ? Qt.darker(Theme.palette.baseColor5, 1.1) : "transparent")
+                                           : (SplitHandle.hovered ? Qt.darker(Theme.palette.baseColor5, 1.1) : "transparent")
             }
             leftPanel: Item {
                 id: leftPanel
@@ -467,8 +467,8 @@ Rectangle {
                         width: leftPanel.width
                         height: implicitHeight > (leftPanel.height - 64) ? implicitHeight + 8 : leftPanel.height - 64
 
-                        chatList.model: demoCommunityChatListItems
-                        categoryList.model: demoCommunityCategoryItems
+                        chatList.model: models.demoCommunityChatListItems
+                        categoryList.model: models.demoCommunityCategoryItems
 
                         showCategoryActionButtons: true
                         onChatItemSelected: selectedChatId = id
@@ -557,6 +557,7 @@ Rectangle {
 
             centerPanel: Item {
                 StatusChatToolBar {
+                    id: statusChatToolBar
                     anchors.top: parent.top
                     width: parent.width
 
@@ -564,11 +565,60 @@ Rectangle {
                     chatInfoButton.subTitle: "Community Chat"
                     chatInfoButton.icon.color: Theme.palette.miscColor6
                     chatInfoButton.type: StatusChatInfoButton.Type.CommunityChat
-                    searchButton.onClicked: searchButton.highlighted = !searchButton.highlighted
+                    onSearchButtonClicked: {
+                        searchButton.highlighted = !searchButton.highlighted;
+                        searchPopup.searchSelectionButton.primaryText = demoCommunityDetailModal.header.title;
+                        searchPopup.searchSelectionButton.image.source = demoCommunityDetailModal.header.image.source;
+                        searchPopup.open();
+                    }
                     membersButton.onClicked: membersButton.highlighted = !membersButton.highlighted
                     onMembersButtonClicked: {
                         root.showRightPanel = !root.showRightPanel;
                     }
+                }
+
+                StatusSearchPopup {
+                    id: searchPopup
+                    searchOptionsPopupMenu: searchPopupMenu
+                    onAboutToHide: {
+                        if (searchPopupMenu.visible) {
+                            searchPopupMenu.close();
+                        }
+                        //clear menu
+                        for (var i = 2; i < searchPopupMenu.count; i++) {
+                            searchPopupMenu.removeItem(searchPopupMenu.takeItem(i));
+                        }
+                    }
+                    onClosed: {
+                        statusChatToolBar.searchButton.highlighted = false
+                        searchPopupMenu.dismiss();
+                    }
+                    onSearchTextChanged: {
+                        if (searchPopup.searchText !== "") {
+                            searchPopup.loading = true;
+                            searchModelSimTimer.start();
+                        } else {
+                            searchPopup.searchResults = [];
+                            searchModelSimTimer.stop();
+                        }
+                    }
+                    Timer {
+                        id: searchModelSimTimer
+                        interval: 500
+                        onTriggered: {
+                            if (searchPopup.searchText.startsWith("c")) {
+                                searchPopup.searchResults = models.searchResultsA;
+                            } else {
+                                searchPopup.searchResults = models.searchResultsB;
+                            }
+                            searchPopup.loading = false;
+                        }
+                    }
+                }
+                StatusSearchLocationMenu {
+                    id: searchPopupMenu
+                    searchPopup: searchPopup
+                    locatioModel: models.optionsModel
                 }
             }
 
@@ -654,7 +704,7 @@ Rectangle {
                         spacing: 4
 
                         Repeater {
-                            model: demoProfileGeneralMenuItems
+                            model: models.demoProfileGeneralMenuItems
                             delegate: StatusNavigationListItem {
                                 title: model.title
                                 icon.name: model.icon
@@ -664,7 +714,7 @@ Rectangle {
                         StatusListSectionHeadline { text: "Settings" }
 
                         Repeater {
-                            model: demoProfileSettingsMenuItems
+                            model: models.demoProfileSettingsMenuItems
                             delegate: StatusNavigationListItem {
                                 title: model.title
                                 icon.name: model.icon
@@ -678,7 +728,7 @@ Rectangle {
                         }
 
                         Repeater {
-                            model: demoProfileOtherMenuItems
+                            model: models.demoProfileOtherMenuItems
                             delegate: StatusNavigationListItem {
                                 title: model.title
                                 icon.name: model.icon
@@ -833,194 +883,7 @@ Rectangle {
         }
     }
 
-    ListModel {
-        id: demoChatListItems
-        ListElement {
-            chatId: "0"
-            name: "#status"
-            chatType: StatusChatListItem.Type.PublicChat
-            muted: false
-            unreadMessagesCount: 0
-            mentionsCount: 0
-            color: "blue"
-        }
-        ListElement {
-            chatId: "1"
-            name: "status-desktop"
-            chatType: StatusChatListItem.Type.PublicChat
-            muted: false
-            color: "red"
-            unreadMessagesCount: 1
-            mentionsCount: 1
-        }
-        ListElement {
-            chatId: "2"
-            name: "Amazing Funny Squirrel"
-            chatType: StatusChatListItem.Type.OneToOneChat
-            muted: false
-            color: "green"
-            identicon: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAYAAAAeP4ixAAAAlklEQVR4nOzW0QmDQBAG4SSkl7SUQlJGCrElq9F3QdjjVhh/5nv3cFhY9vUIYQiNITSG0Bh
-CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2ImYgiNITTlTdG1nUZ5a92VITQxITFiJmIIjSE0htAYQrMHAAD//+wwFVpz+yqXAAAAAElFTkSuQmCC"
-            unreadMessagesCount: 0
-        }
-        ListElement {
-            chatId: "3"
-            name: "Black Ops"
-            chatType: StatusChatListItem.Type.GroupChat
-            muted: false
-            color: "purple"
-            unreadMessagesCount: 0
-        }
-        ListElement {
-            chatId: "4"
-            name: "Spectacular Growing Otter"
-            chatType: StatusChatListItem.Type.OneToOneChat
-            muted: true
-            color: "Orange"
-            unreadMessagesCount: 0
-        }
-        ListElement {
-            chatId: "5"
-            name: "channel-with-a-super-duper-long-name"
-            chatType: StatusChatListItem.Type.PublicChat
-            muted: false
-            color: "green"
-            unreadMessagesCount: 0
-        }
-    }
-
-    ListModel {
-        id: demoCommunityChatListItems
-        ListElement {
-            chatId: "0"
-            name: "general"
-            chatType: StatusChatListItem.Type.CommunityChat
-            muted: false
-            unreadMessagesCount: 0
-            color: "orange"
-        }
-        ListElement {
-            chatId: "1"
-            name: "random"
-            chatType: StatusChatListItem.Type.CommunityChat
-            muted: false
-            unreadMessagesCount: 0
-            color: "orange"
-            categoryId: "public"
-        }
-        ListElement {
-            chatId: "2"
-            name: "watercooler"
-            chatType: StatusChatListItem.Type.CommunityChat
-            muted: false
-            unreadMessagesCount: 0
-            color: "orange"
-            categoryId: "public"
-        }
-        ListElement {
-            chatId: "3"
-            name: "language-design"
-            chatType: StatusChatListItem.Type.CommunityChat
-            muted: false
-            unreadMessagesCount: 0
-            color: "orange"
-            categoryId: "dev"
-        }
-    }
-
-    ListModel {
-        id: demoCommunityCategoryItems
-        ListElement {
-            categoryId: "public"
-            name: "Public"
-        }
-        ListElement {
-            categoryId: "dev"
-            name: "Development"
-        }
-    }
-
-    ListModel {
-        id: demoProfileGeneralMenuItems
-
-        ListElement {
-            title: "My Profile"
-            icon: "profile"
-        }
-
-        ListElement {
-            title: "Contacts"
-            icon: "contact"
-        }
-
-        ListElement {
-            title: "ENS Usernames"
-            icon: "username"
-        }
-
-    }
-
-    ListModel {
-        id: demoProfileSettingsMenuItems
-
-        ListElement {
-            title: "Privacy & Security"
-            icon: "security"
-        }
-
-        ListElement {
-            title: "Appearance"
-            icon: "appearance"
-        }
-
-        ListElement {
-            title: "Browser"
-            icon: "browser"
-        }
-
-        ListElement {
-            title: "Sounds"
-            icon: "sound"
-        }
-
-        ListElement {
-            title: "Language"
-            icon: "language"
-        }
-
-        ListElement {
-            title: "Notifications"
-            icon: "notification"
-        }
-
-        ListElement {
-            title: "Sync settings"
-            icon: "mobile"
-        }
-
-        ListElement {
-            title: "Advanced"
-            icon: "settings"
-        }
-
-    }
-
-    ListModel {
-        id: demoProfileOtherMenuItems
-
-        ListElement {
-            title: "Need help?"
-            icon: "help"
-        }
-
-        ListElement {
-            title: "About"
-            icon: "info"
-        }
-
-        ListElement {
-            title: "Sign out & Quit"
-            icon: "logout"
-        }
+    Models {
+        id: models
     }
 }
--- a/sandbox/ListItems.qml
+++ b/sandbox/ListItems.qml
@@ -295,6 +295,27 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
         type: StatusListItem.Type.Danger
     }
 
+    StatusListItem {
+        title: "List Item with Badge"
+        subTitle: "Subtitle"
+        image.source: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAYAAAAeP4ixAAAAlklEQVR4nOzW0QmDQBAG4SSkl7SUQlJGCrElq9F3QdjjVhh/5nv3cFhY9vUIYQiNITSG0Bh
+CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2ImYgiNITTlTdG1nUZ5a92VITQxITFiJmIIjSE0htAYQrMHAAD//+wwFVpz+yqXAAAAAElFTkSuQmCC"
+        image.isIdenticon: true
+        badge.image.source: "https://pbs.twimg.com/profile_images/1369221718338895873/T_5fny6o_400x400.jpg"
+        badge.primaryText: "CryptoKitties"
+        badge.secondaryText: "#test"
+    }
+
+    StatusListItem {
+        title: "List Item with Badge 2"
+        subTitle: "Subtitle"
+        icon.isLetterIdenticon: true
+        badge.primaryText: "CryptoKitties"
+        badge.secondaryText: "#test"
+        badge.icon.color: "orange"
+        badge.icon.isLetterIdenticon: true
+    }
+
     StatusDescriptionListItem {
         title: "Title"
         subTitle: "Subtitle"
--- a/src/StatusQ/Components/StatusListItem.qml
+++ b/src/StatusQ/Components/StatusListItem.qml
@@ -1,4 +1,6 @@
 import QtQuick 2.13
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Components 0.1
@@ -33,8 +35,11 @@ Rectangle {
     radius: 8
 
     property string title: ""
+    property string titleAsideText: ""
     property string subTitle: ""
     property string tertiaryTitle: ""
+    property alias badge: statusListItemBadge
+
     property real leftPadding: 16
     property real rightPadding: 16
     property bool enabled: true
@@ -107,6 +112,7 @@ Rectangle {
                 }
                 return !!statusListItem.icon.name ? statusRoundedIcon : statusRoundedImage
             }
+
             active: statusListItem.icon.isLetterIdenticon || 
                     !!statusListItem.icon.name || 
                     !!statusListItem.image.source.toString()
@@ -158,15 +164,15 @@ Rectangle {
             anchors.leftMargin: statusListItem.leftPadding
             anchors.rightMargin: statusListItem.rightPadding
             anchors.verticalCenter: parent.verticalCenter
-            height: statusListItemTitle.height + 
-              (statusListItemSubTitle.visible ? statusListItemSubTitle.height : 0) +
-              (statusListItemTertiaryTitle.visible ? statusListItemTertiaryTitle.height : 0)
+            height: childrenRect.height
 
             StatusBaseText {
                 id: statusListItemTitle
                 text: statusListItem.title
-                width: parent.width
+                width: contentWidth < (parent.width - statusListItemTitleAsideText.contentWidth) ?
+                       contentWidth : (parent.width - statusListItemTitleAsideText.contentWidth)
                 font.pixelSize: 15
+                height: visible ? contentHeight : 0
                 wrapMode: Text.WrapAtWordBoundaryOrAnywhere
                 color: {
                   if (!statusListItem.enabled) {
@@ -184,12 +190,24 @@ Rectangle {
             }
 
             StatusBaseText {
+                id: statusListItemTitleAsideText
+                anchors.left: statusListItemTitle.right
+                anchors.leftMargin: 4
+                anchors.verticalCenter: statusListItemTitle.verticalCenter
+                text: statusListItem.titleAsideText
+                font.pixelSize: 10
+                color: Theme.palette.baseColor1
+                visible: !!statusListItem.titleAsideText
+            }
+
+            StatusBaseText {
                 id: statusListItemSubTitle
                 anchors.top: statusListItemTitle.bottom
                 width: parent.width
                 text: statusListItem.subTitle
                 font.pixelSize: 15
                 color: !statusListItem.enabled || !statusListItem.tertiaryTitle ? Theme.palette.baseColor1 : Theme.palette.directColor1
+                height: visible ? contentHeight : 0
                 visible: !!statusListItem.subTitle
                 wrapMode: Text.WrapAtWordBoundaryOrAnywhere
             }
@@ -198,11 +216,19 @@ Rectangle {
                 id: statusListItemTertiaryTitle
                 anchors.top: statusListItemSubTitle.bottom
                 width: parent.width
+                height: visible ? contentHeight : 0
                 text: statusListItem.tertiaryTitle
                 color: Theme.palette.baseColor1
                 font.pixelSize: 13
                 visible: !!statusListItem.tertiaryTitle
                 wrapMode: Text.WrapAtWordBoundaryOrAnywhere
+            }
+
+            StatusListItemBadge {
+                id: statusListItemBadge
+                anchors.top: statusListItemTertiaryTitle.bottom
+                width: contentItem.width
+                implicitHeight: visible ? 22 : 0
             }
         }
 
--- a/src/StatusQ/Components/StatusListItemBadge.qml
+++ b/src/StatusQ/Components/StatusListItemBadge.qml
@@ -1,0 +1,80 @@
+import QtQuick 2.13
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+Control {
+    id: root
+    implicitWidth: contentItem.width
+    implicitHeight: visible ? 22 : 0
+    visible: !!root.primaryText
+
+    property string primaryText: ""
+    property string secondaryText: ""
+
+    property StatusImageSettings image: StatusImageSettings {
+        height: 16
+        width: 16
+        isIdenticon: false
+    }
+    property StatusIconSettings icon: StatusIconSettings {
+        height: 16
+        width: 16
+        isLetterIdenticon: false
+        background: StatusIconBackgroundSettings {}
+        color: "transparent"
+    }
+
+    background: Rectangle {
+        anchors.fill: parent
+        color: "transparent"
+        radius: 11
+        border.color: Theme.palette.directColor7
+    }
+
+    contentItem: Item {
+        id: contentItem
+        width: (contentItemRow.width + 10)
+        height: parent.height
+        RowLayout {
+            id: contentItemRow
+            anchors.centerIn: parent
+            anchors.horizontalCenterOffset: -spacing
+            spacing: 2
+            StatusRoundedImage {
+                implicitWidth: root.image.width
+                implicitHeight: root.image.height
+                visible: !root.icon.isLetterIdenticon
+                image.source: root.image.source
+                border.color: Theme.palette.baseColor1
+                border.width: root.image.isIdenticon ? 1 : 0
+            }
+            StatusLetterIdenticon {
+                implicitWidth: root.icon.width
+                implicitHeight: root.icon.width
+                letterSize: 11
+                visible: root.icon.isLetterIdenticon
+                color: root.icon.color
+                name: root.primaryText
+            }
+            StatusBaseText {
+                font.weight: Font.Medium
+                color: Theme.palette.baseColor1
+                text: root.primaryText
+            }
+            StatusIcon {
+                Layout.alignment: Qt.AlignVCenter
+                visible: !!root.secondaryText
+                color: Theme.palette.baseColor1
+                icon: "next"
+            }
+            StatusBaseText {
+                font.weight: Font.Medium
+                color: Theme.palette.baseColor1
+                visible: !!root.secondaryText
+                text: root.secondaryText
+            }
+        }
+    }
+}
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -179,8 +179,8 @@ Item {
 
     StatusFlatRoundButton {
         id: clearButtton
-        visible: edit.text.length != 0 && 
-            statusBaseInput.clearable && 
+        visible: edit.text.length != 0 &&
+            statusBaseInput.clearable &&
             !statusBaseInput.multiline &&
             edit.activeFocus
         anchors.right: parent.right
--- a/src/StatusQ/Popups/StatusModal.qml
+++ b/src/StatusQ/Popups/StatusModal.qml
@@ -19,6 +19,8 @@ QC.Popup {
     property alias contentComponent: contentLoader.item
     property alias rightButtons: footerImpl.rightButtons
     property alias leftButtons: footerImpl.leftButtons
+    property bool showHeader: true
+    property bool showFooter: true
 
     signal editButtonClicked()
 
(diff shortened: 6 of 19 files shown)


PascalPrecht:
Haven't reviewed the code yet, but wanted to try it out. Unfortunately, I'm unable to try your changes in the demo application because this commit seems to break the sandbox app when navigating to the demo app.

All I'm getting is

```
zsh: segmentation fault  ./bin/sandbox
```

Are you experiencing the same?

alexandraB99:
> Haven't reviewed the code yet, but wanted to try it out. Unfortunately, I'm unable to try your changes in the demo application because this commit seems to break the sandbox app when navigating to the demo app.
> 
> All I'm getting is
> 
> ```
> zsh: segmentation fault  ./bin/sandbox
> ```
> 
> Are you experiencing the same?

@PascalPrecht no. I just rebased and tried with a fresh build and still I see no issues.

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -16,35 +24,26 @@ Rectangle {
         Danger
     }
 
-    property StatusIconSettings icon: StatusIconSettings {
+    property StatusIconSettings iconSettings: StatusIconSettings {
```
PascalPrecht, line 27: This and the move to `Button` introduces a breaking change which we should avoid for now. I could see that this was probably done to be consistent with how icon settings work with `StatusMenuItem`. There however, we had to use `iconSettings` over `icon` because the built-in `icon` property couldn't be mutated.

I'm very fine with moving to `iconSettings` altogether in the long run but doubt that this breaking change is needed for this PR.

Maybe you can shine some light onto this :D I'm probably missing something

src/StatusQ/Popups/StatusPopupMenu.qml
```diff
@@ -19,6 +19,8 @@ Menu {
     property var subMenuItemIcons: []
     property var openHandler
     property var closeHandler
+    property string currentImg
+    property string currentIdColor
```
PascalPrecht, line 23: It doesn't seem to be obvious why these ones are around.  Is this fixing some bug?

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,192 @@
...
+                    anchors.leftMargin: 8
+                    anchors.right: parent.right
+                    anchors.rightMargin: 8
+                    anchors.bottom: parent.bottom
+                    input.icon.name: "search"
+                    input.placeholderText: "Search"
+                    input.bgVisible: false
+                }
+            }
+            StatusMenuSeparator { Layout.fillWidth: true }
```
PascalPrecht, line 53: Semantically, this is not a menu, so we might want to consider introducing a general `StatusSeparator` in the near future.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,192 @@
...
+                                image.source: root.selectionBadgeImage
+                            }
+                            StatusLetterIdenticon {
+                                implicitWidth: 16
+                                implicitHeight: 16
+                                letterSize: 11
+                                visible: ((root.selectionBadgeImage === "") && (root.selectionBadgePrimaryText !== "Anywhere"))
+                                color: root.selectionBadgeIdenticonColor
+                                name: root.selectionBadgePrimaryText
+                            }
```
PascalPrecht, line 87: > We're using this mechanism where we first try StatusRoundedImage and fallback to LetterIdenticon if it fails (or go to letter identicon straight away if no image source was provided. Maybe we can use a similar logic here?

> Here's an example of this: https://github.com/status-im/StatusQ/blob/master/src/StatusQ/Components/StatusChatListItem.qml#L59-L103

Maybe this applies here as well?

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,192 @@
...
+                    model: root.searchResults
+                    delegate: StatusListItem {
+                        width: view.width
+                        title: model.name
+                        statusListItemTitle.color: model.name.startsWith("@") ? Theme.palette.primaryColor1 : Theme.palette.directColor1
+                        subTitle: model.content
+                        statusListItemSubTitle.height: model.content !== "" ? 20 : 0
+                        statusListItemSubTitle.elide: Text.ElideRight
+                        statusListItemSubTitle.color: Theme.palette.black
+                        titleAsideText: model.time //Utils.formatTime(timestamp)
```
PascalPrecht, line 155: Left over comment?

PascalPrecht (changes requested):
Massive work! As commented above, I couldn't play around with this yet, so I'll give it another spin once the app crash is resolved.

PascalPrecht:
> @PascalPrecht no. I just rebased and tried with a fresh build and still I see no issues.

@alexandraB99 that's interesting. Did you run `sandbox/scripts/build` before `./bin/sandbox` ?
If not, can you give it another spin after running the build script?  I can't access the demo application without sandbox crashing in this commit...

PascalPrecht:
Sorry, to clarify: the demo app can be opened but opening the community in the demo app crashes. That's what I'm referring to.

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -1,4 +1,6 @@
 import QtQuick 2.13
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
```
borismelnik, line 3: If you do't mind pls use minor version of QtQuick import

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -198,12 +218,69 @@ Rectangle {
...
                 width: parent.width
+                height: visible ? contentHeight : 0
                 text: statusListItem.tertiaryTitle
                 color: Theme.palette.baseColor1
                 font.pixelSize: 13
                 visible: !!statusListItem.tertiaryTitle
                 wrapMode: Text.WrapAtWordBoundaryOrAnywhere
             }
+
+            Control {
```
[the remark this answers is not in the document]
borismelnik: and i think if use some customization we should create file for it `StatusListItemBadge`

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -1,10 +1,18 @@
 import QtQuick 2.14
+import QtQuick.Controls 2.12
```
borismelnik, line 2: The same as before. Qt creator sometimes dont understand which modules is default.

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -1,10 +1,18 @@
...
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Components 0.1
 
-Rectangle {
+Button {
```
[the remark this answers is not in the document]
borismelnik: i agreed. Why `Button`?

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -51,27 +52,29 @@ Item {
     implicitWidth: 448
     implicitHeight: multiline ? Math.max(edit.implicitHeight + topPadding + bottomPadding, 44) : 44
 
-    Rectangle {
+    Item {
```
borismelnik, line 55: Actually i dont understand why you create two components instead of one?

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -179,10 +182,10 @@ Item {
...
     StatusFlatRoundButton {
         id: clearButtton
-        visible: edit.text.length != 0 && 
-            statusBaseInput.clearable && 
-            !statusBaseInput.multiline &&
-            edit.activeFocus
+        visible: edit.text.length != 0 &&
+                 statusBaseInput.clearable &&
+                 !statusBaseInput.multiline &&
+                 edit.activeFocus
```
borismelnik, line 188: Its formatter stuff?

src/StatusQ/Popups/StatusModal.qml
```diff
@@ -43,7 +43,7 @@ QC.Popup {
     }
 
     contentItem: Column {
-        width: parent.width
+        width: parent ? parent.width : 0
```
borismelnik, line 46: sometimes parent can be null?

PascalPrecht:
@alexandraB99 Since you've already figured out, using `contentItem` in `StatusModal` breaks the application. `StatusModal` has a dedicated property for content items which is called `content`. You can see how it's used in other modal examples in the sandbox application.

Using that property, the app no longer breaks. However, it seems that the search component doesn't look quite how it's been designed (not sure why QtCreator shows something differeent but that's kinda what I've been referring to when I said it doesn't seem to be reliable):

Screenshot 2021-07-30 at 14 13 33


Also, notice that, when you want to access components that live inside `content`, they have to be accessed via `contentComponent` because the content is loaded via a `Loader`.

src/StatusQ/Controls/StatusBaseButton.qml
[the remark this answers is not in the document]
alexandraB99: Because I want to re-use the background, which is the same and all other features that might be needed and so I don't like to duplicate code.

src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: see above regarding the background

src/StatusQ/Popups/StatusPopupMenu.qml
alexandraB99: no. those are exposed so that it can be set in the Button that opens the popup menu.

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: Can we just rename this one if there is no difference between those 2? Maybe create a task for the near future?

src/StatusQ/Components/StatusListItem.qml
alexandraB99: You're referring to Controls and Layouts? because QtQuick was there already

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: What do you mean?

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: Yes, ctrl+i

src/StatusQ/Popups/StatusModal.qml
alexandraB99: was when using the contentItem, switched back to use content now

src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: please see above, you mean QtQuick or QtQuick.Controls import?

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: removed

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: please see above

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -16,35 +24,26 @@ Rectangle {
...
-    property StatusIconSettings icon: StatusIconSettings {
+    property StatusIconSettings iconSettings: StatusIconSettings {
         width: 24
         height: 24
     }
 
     property bool loading: false
-
-    property alias hovered: sensor.containsMouse
+    //property bool hovered: sensor.containsMouse
```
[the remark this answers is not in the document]
alexandraB99: removed

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -198,12 +218,69 @@ Rectangle {
...
+                            border.color: Theme.palette.baseColor1
+                        }
+                        StatusLetterIdenticon {
+                            implicitWidth: 16
+                            implicitHeight: 16
+                            letterSize: 11
+                            visible: (statusListItem.badgeImage === "")
+                            color: statusListItem.badgeIdenticonColor
+                            name: statusListItem.badgePrimaryText
+                        }
```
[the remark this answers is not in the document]
alexandraB99: do you mean to use `image.source.toString()` instead of `=== ""`?

src/StatusQ/Components/StatusListItem.qml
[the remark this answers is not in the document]
alexandraB99: done

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -183,13 +191,25 @@ Rectangle {
...
+            StatusBaseText {
+                id: statusListItemTitleAsideText
+                anchors.left: statusListItemTitle.right
+                anchors.leftMargin: 4
+                anchors.verticalCenter: statusListItemTitle.verticalCenter
+                text: statusListItem.titleAsideText
+                font.pixelSize: 10
+                color: Theme.palette.baseColor1
+                visible: !!statusListItem.titleAsideText
+            }
```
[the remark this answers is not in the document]
alexandraB99: done

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -33,8 +35,13 @@ Rectangle {
     radius: 8
 
     property string title: ""
+    property string titleAsideText: ""
     property string subTitle: ""
     property string tertiaryTitle: ""
+    property string badgeImage: ""
+    property string badgePrimaryText: ""
+    property string badgeSecondaryText: ""
+    property string badgeIdenticonColor: ""
```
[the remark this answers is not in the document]
alexandraB99: done

sandbox/DemoApp.qml
```diff
@@ -564,12 +564,180 @@ Rectangle {
...
+                            ListElement { name: "Crocodile Vanilla Bird"; type: "Chat"; time: ""; content: ""; badgeImage: "https://pbs.twimg.com/profile_images/1369221718338895873/T_5fny6o_400x400.jpg"; badgePrimaryText: ""; badgeSecondaryText: ""; badgeIdenticonColor: "" }
+                            ListElement { name: "carmen eth"; type: "Chat"; time: ""; content: ""; badgeImage: ""; badgePrimaryText: ""; badgeSecondaryText: ""; badgeIdenticonColor: "" }
+                            ListElement { name: "CryptoKitties"; type: "Communities"; time: ""; content: ""; badgeImage: ""; badgePrimaryText: ""; badgeSecondaryText: ""; badgeIdenticonColor: "" }
+                            ListElement { name: "MyCommunity"; type: "Communities"; time: ""; content: ""; badgeImage: ""; badgePrimaryText: ""; badgeSecondaryText: ""; badgeIdenticonColor: "" }
+                            ListElement { name: "Foo"; type: "Communities"; time: ""; content: ""; badgeImage: ""; badgePrimaryText: ""; badgeSecondaryText: ""; badgeIdenticonColor: "" }
+                        }
+                        property var searchResultsElse: ListModel {
```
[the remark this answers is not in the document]
alexandraB99: done

sandbox/DemoApp.qml
```diff
@@ -564,12 +564,180 @@ Rectangle {
...
+                        } else {
+                            searchPopup.searchResults.clear();
+                            searchModelSimTimer.stop();
+                        }
+                    }
+                    Timer {
+                        id: searchModelSimTimer
+                        interval: 500
+                        //dummy models
+                        property var searchResultsC: ListModel {
```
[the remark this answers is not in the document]
alexandraB99: done

alexandraB99:
> @alexandraB99 Since you've already figured out, using `contentItem` in `StatusModal` breaks the application. `StatusModal` has a dedicated property for content items which is called `content`. You can see how it's used in other modal examples in the sandbox application.
> 
> Using that property, the app no longer breaks. However, it seems that the search component doesn't look quite how it's been designed (not sure why QtCreator shows something differeent but that's kinda what I've been referring to when I said it doesn't seem to be reliable):
> 
> Screenshot 2021-07-30 at 14 13 33
> 
> Also, notice that, when you want to access components that live inside `content`, they have to be accessed via `contentComponent` because the content is loaded via a `Loader`.

switched to use `content`, crash is gone now, please verify

src/StatusQ/Components/StatusListItemBadge.qml
```diff
@@ -0,0 +1,65 @@
...
+                implicitWidth: 16
+                implicitHeight: 16
+                letterSize: 11
+                visible: (root.badgeImage === "")
+                color: root.badgeIdenticonColor
+                name: root.badgePrimaryText
+            }
+            StatusBaseText {
+                color: Theme.palette.baseColor1
+                text: root.badgePrimaryText
```
PascalPrecht, line 50: This should be `Font.Medium` weight

src/StatusQ/Components/StatusListItemBadge.qml
```diff
@@ -0,0 +1,65 @@
...
+            StatusIcon {
+                Layout.alignment: Qt.AlignVCenter
+                visible: !!root.badgeSecondaryText
+                color: Theme.palette.baseColor1
+                icon: "next"
+            }
+            StatusBaseText {
+                color: Theme.palette.baseColor1
+                visible: !!root.badgeSecondaryText
+                text: root.badgeSecondaryText
```
PascalPrecht, line 61: This should be `Font.Medium` weight

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: I think the reason this was done, was so that the `Rectangle` could be made invisible while keeping the element around.

src/StatusQ/Controls/StatusFlatButton.qml
```diff
@@ -14,7 +14,7 @@ StatusBaseButton {
                                                      : Theme.palette.dangerColor1
     disabledTextColor: Theme.palette.baseColor1
 
-    border.color: type === StatusBaseButton.Type.Normal || hovered ? "transparent"
+    borderColor: type === StatusBaseButton.Type.Normal || hovered ? "transparent"
```
PascalPrecht, line 17: I assume this is a result of moving to `Button` type in `StatusBaseButton` as well?

src/StatusQ/Popups/StatusPopupMenu.qml
PascalPrecht: Got it!

We should come up with better names though. Especially for `currentIdColor`. How about `currentItemColor` and then we can also go with `currentItemImageSource`?

src/StatusQ/Popups/StatusSearchPopup.qml
PascalPrecht: Yea absolutely. Let's just keep this as is and rename this in the future.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                    anchors.left: parent.left
+                    anchors.leftMargin: 8
+                    anchors.right: parent.right
+                    anchors.rightMargin: 8
+                    anchors.bottom: parent.bottom
+                    input.icon.name: "search"
+                    input.placeholderText: "Search"
+                    input.bgVisible: false
+                }
+            }
```
PascalPrecht, line 58: The size of this icon seems wrong. This is what has been designed:

Screenshot 2021-08-03 at 23 06 56

This is what it looks like:

Screenshot 2021-08-03 at 23 07 23


If `StatusBaseInput` is the issue here and not flexible enough, it's perfectly fine to introduce a custom inline input for this scenario.

src/StatusQ/Popups/StatusSearchPopup.qml
PascalPrecht: Not really sure what you're referring to here...

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                            anchors.centerIn: parent
+                            spacing: 2
+                            StatusBaseText {
+                                text: "In: "
+                            }
+                            StatusRoundedImage {
+                                implicitWidth: 16
+                                implicitHeight: 16
+                                visible: !!root.selectionBadgeImage
+                                image.source: root.selectionBadgeImage
```
PascalPrecht, line 84: This one needs a border in case the selected image is an identicon. Just as with the `StatusListItemBadge`, you'd get all the information you need if this was turned into `StatusImageSettings`.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                    implicitHeight: 32
+
+                    contentItem: Item {
+                        anchors.fill: parent
+                        RowLayout {
+                            id: contentItemRowLayout
+                            anchors.centerIn: parent
+                            spacing: 2
+                            StatusBaseText {
+                                text: "In: "
```
PascalPrecht, line 78: We probably should either alias this button and expose it or introduce a property on the search popup, so that this label can be internationalized.

PascalPrecht: Also this needs to use `Theme.palette.directColor1`

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                            StatusLetterIdenticon {
+                                implicitWidth: 16
+                                implicitHeight: 16
+                                letterSize: 11
+                                visible: ((root.selectionBadgeImage === "") && (root.selectionBadgePrimaryText !== "Anywhere"))
+                                color: root.selectionBadgeIdenticonColor
+                                name: root.selectionBadgePrimaryText
+                            }
+                            StatusBaseText {
+                                text: root.selectionBadgePrimaryText
```
PascalPrecht, line 95: This needs to use `Theme.palette.directColor1` as well

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                            }
+                            StatusIcon {
+                                Layout.alignment: Qt.AlignVCenter
+                                visible: !!root.selectionBadgeSecondaryText
+                                color: Theme.palette.baseColor1
+                                icon: "next"
```
PascalPrecht, line 101: The size of this icon is incorrect.
Screenshot 2021-08-03 at 23 15 06

This was designed:

Screenshot 2021-08-03 at 23 15 26

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                            }
+                            StatusBaseText {
+                                visible: !!root.selectionBadgeSecondaryText
+                                text: "#" + root.selectionBadgeSecondaryText
+                            }
+                            StatusIcon {
+                                Layout.alignment: Qt.AlignVCenter
+                                icon: "chevron-down"
```
PascalPrecht, line 109: The size of this icon is off.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                    contentItem: Item {
+                        anchors.fill: parent
+                        RowLayout {
+                            id: contentItemRowLayout
+                            anchors.centerIn: parent
+                            spacing: 2
+                            StatusBaseText {
+                                text: "In: "
+                            }
+                            StatusRoundedImage {
```
PascalPrecht, line 80: In front of this image (or letter identicon) there's a possibility that we need to render another icon, in case we're selecting a private group chat type:

Screenshot 2021-08-03 at 23 18 36

You can check  https://github.com/status-im/StatusQ/blob/master/src/StatusQ/Components/StatusChatListItem.qml#L126-L137 to see what you need for this.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                                implicitWidth: 16
+                                implicitHeight: 16
+                                visible: !!root.selectionBadgeImage
+                                image.source: root.selectionBadgeImage
+                            }
+                            StatusLetterIdenticon {
+                                implicitWidth: 16
+                                implicitHeight: 16
+                                letterSize: 11
+                                visible: ((root.selectionBadgeImage === "") && (root.selectionBadgePrimaryText !== "Anywhere"))
```
PascalPrecht, line 90: Can't rely on `"Anywhere"` here as the text could be in a different language when i18n'ed.

Do we have other heuristics?

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                                text: "#" + root.selectionBadgeSecondaryText
+                            }
+                            StatusIcon {
+                                Layout.alignment: Qt.AlignVCenter
+                                icon: "chevron-down"
+                            }
+                        }
+                    }
+                    onClicked: { searchPopupMenuLoader.item.popup(); }
+                }
```
PascalPrecht, line 114: As seen in previous screenshots attached in this review, the background of this button doesn't match the design. This needs update as well

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                    sourceComponent: root.searchOptionsPopupMenu
+                }
+
+                StatusFlatRoundButton {
+                    id: closeButton
+                    width: 32
+                    height: 32
+                    anchors.left: searchOptionsMenuButton.right
+                    anchors.verticalCenter: parent.verticalCenter
+                    opacity: (root.selectionBadgePrimaryText === "Anywhere") ? 0.0 : 1.0
```
PascalPrecht, line 127: Same as above: Can't rely on `Anywhere` here.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -0,0 +1,198 @@
...
+                    }
+                }
+                StatusLoadingIndicator {
+                    anchors.centerIn: parent
+                    visible: root.loading
+                }
+
+                StatusBaseText {
+                    anchors.centerIn: parent
+                    text: "No results"
```
PascalPrecht, line 184: Needs to be i18n'able

PascalPrecht (changes requested):
@alexandraB99 thank you for updating the PR!

I could now try this out and play with it and it seems to work great, aside from a few bugs I've found. Super sorry for this, don't wanna be the annoying guy, but there's a lot of UI related inconsistencies with the design as well that we need to iron out.

Also the component doesn't work well in darkmode. The comments should address all of these things.

Please let me know if anything is unclear! I've attached screenshots accordingly to make it easier to follow

src/StatusQ/Components/StatusListItem.qml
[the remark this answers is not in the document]
PascalPrecht: Oh I'm seeing this comment just now... So the mechanism I was referring to was:

1. First try and see if there's an `image.source`, then we know we want to render a `RoundedImage`.
2. Now the question is if it's an identicon or not. If yes, we need to render a border around it. 
3. If loading the image fails, we fallback to letter identicon
4. if there was no image.source in the first place, we probably deal with an icon and `isLetterIdenticon` true, in which case we'll render a letter identicon right away.

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: In this case isn't it better to use an Item and set its visibility to false instead? Rectangles are expensive when it comes to performance (due to gl calls) since they suppose to paint on screen and it's recommendable to use only when we need to really paint (eg have some background color)

sandbox/DemoApp.qml
```diff
@@ -564,12 +564,150 @@ Rectangle {
...
+                            id: cryptoKittiesPopupMenu
+                            title: "Cryptokitties"
+
+                            StatusMenuItem {
+                                text: "welcome"
+                                iconSettings.name: "channel"
+                                iconSettings.color: Theme.palette.directColor1
+                                onTriggered: {
+                                    searchPopup.resetSelectionBadge();
+                                    searchPopup.selectionBadgeImage = cryptoKittiesPopupMenu.currentImg;
```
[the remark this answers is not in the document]
alexandraB99: There is a problem with StatusPopupMenu, you can try printing `currentOptionItemImgSource` for both 'Chat' and 'Cryptokitties' in DemoApp.qml  (on `Component.onCompleted`) and you'll get 
```
qml: Cryptokitties 
qml: Chat https://pbs.twimg.com/profile_images/1369221718338895873/T_5fny6o_400x400.jpg
```
so `currentOptionItemImgSource` for 'Cryptokitties' is always empty.

sandbox/DemoApp.qml
```diff
@@ -564,12 +564,150 @@ Rectangle {
                     chatInfoButton.subTitle: "Community Chat"
                     chatInfoButton.icon.color: Theme.palette.miscColor6
                     chatInfoButton.type: StatusChatInfoButton.Type.CommunityChat
-                    searchButton.onClicked: searchButton.highlighted = !searchButton.highlighted
+                    onSearchButtonClicked: {
+                        searchButton.highlighted = !searchButton.highlighted
+                        searchPopup.searchOptionsPopupMenu = searchPopupMenu;
```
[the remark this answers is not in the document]
alexandraB99: not really, just for demonstration :) changed it

sandbox/ListItems.qml
```diff
@@ -295,6 +295,26 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
         type: StatusListItem.Type.Danger
     }
 
+    StatusListItem {
+        title: "List Item with Badge"
+        subTitle: "Subtitle"
+        image.source: "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAYAAAAeP4ixAAAAlklEQVR4nOzW0QmDQBAG4SSkl7SUQlJGCrElq9F3QdjjVhh/5nv3cFhY9vUIYQiNITSG0Bh
+CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2ImYgiNITTlTdG1nUZ5a92VITQxITFiJmIIjSE0htAYQrMHAAD//+wwFVpz+yqXAAAAAElFTkSuQmCC"
+        image.isIdenticon: true
+        badge.badgeImage: "https://pbs.twimg.com/profile_images/1369221718338895873/T_5fny6o_400x400.jpg"
```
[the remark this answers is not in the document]
alexandraB99: not sure I get this, could you please explain?

sandbox/ListItems.qml
```diff
@@ -295,6 +295,26 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
...
+        badge.badgePrimaryText: "CryptoKitties"
+        badge.badgeSecondaryText: "#test"
```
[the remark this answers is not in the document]
alexandraB99: sure thing, updated

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done, added

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: Exactly, does it really make sense to duplicate a whole text input for just making the icon bigger or maybe make the existing StatusInput able to have a bigger icon?

PascalPrecht: I'm honestly happy fine either way. The search input is a super special version of an input. If it can't be easily derived from `StatusBaseInput` that's perfectly fine.

I personally would probably build a new one (that's also dramatically simpler than `StatusBaseInput`. But if you prefer making the base one more flexible, please go ahead!

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: what do you mean?

src/StatusQ/Popups/StatusPopupMenu.qml
alexandraB99: changed to `currentOptionItemColor` and `currentOptionItemImgSource`

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: I was just trying to interpret what this change from `Rectangle` to `Item` + the introduction of another `Rectangle` below was about :D

src/StatusQ/Components/StatusListItemBadge.qml
alexandraB99: done

src/StatusQ/Components/StatusListItemBadge.qml
alexandraB99: done

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: Done, switched to use a TextInput instead

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done. however I can't use 4.5 - 7.5 values from figma because it looks tiny, so I increased to 14.5 - 17.5

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done, same as above

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: Done, updated. Also switched StatusBaseButton as it was (Rectangle) and made this one a QtQuick Controls 2 Button instead

sandbox/DemoApp.qml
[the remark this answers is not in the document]
PascalPrecht: So... the url shown in `Chat` however is the image that belongs to `CryptoKitties` in this case. Probably a off-by-one bug?

src/StatusQ/Components/StatusListItem.qml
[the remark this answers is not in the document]
alexandraB99: not sure if in this case the user can set manually properties like `isLetterIdenticon` or `isIdenticon`. I assume this info will come from some model, plus I don't see any borders surrounding the images looking at the design, maybe I'm missing something?

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: Probably some clipping issue, removed now as I switched to use a TextInput instead

src/StatusQ/Components/StatusListItemBadge.qml
```diff
@@ -0,0 +1,65 @@
...
+            id: contentItemRow
+            anchors.centerIn: parent
+            spacing: 2
+            StatusRoundedImage {
+                implicitWidth: 16
+                implicitHeight: 16
+                visible: !!root.badgeImage
+                image.source: root.badgeImage
+                border.color: Theme.palette.baseColor1
+            }
```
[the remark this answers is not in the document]
alexandraB99: As I already wrote above: not sure if in this case the user will be able to set manually properties like `isIdenticon`. I assume this info will come from some model, does it still make sense to use StatusImageSettings?

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: See https://github.com/status-im/StatusQ/pull/301#discussion_r683337557

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done

src/StatusQ/Popups/StatusSearchPopup.qml
alexandraB99: done

sandbox/DemoApp.qml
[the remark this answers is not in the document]
alexandraB99: Not sure why, however as this is relevant to StatusPopupMenu and not the SearchPopup, should we look fix this in this task or maybe file another bug?

src/StatusQ/Controls/StatusFlatButton.qml
alexandraB99: removed

src/StatusQ/Components/StatusListItem.qml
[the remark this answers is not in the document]
PascalPrecht: The moment the popup gets generated this information should be available. Basically, the rule is, if there's no profile image, but an image source is provided, then it's an identicon. Identicon profile images Always render with a border in the application. Need to check the design if it's missing there. You can see this being done in many other places in StatusQ

src/StatusQ/Popups/StatusPopupMenu.qml
PascalPrecht: Great

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Cool!

sandbox/DemoApp.qml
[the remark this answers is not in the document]
PascalPrecht: The currentOptionItemImgSource and other properties have been introduced in this pr so we should also fix them here, otherwise we introduce this bug in the First place just to fix it later.

sandbox/ListItems.qml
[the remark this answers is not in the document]
PascalPrecht: Instead of having badeImage, it'd be `image` of type `StatusImageSettings` then you can set the `source` and other props

src/StatusQ/Components/StatusListItemBadge.qml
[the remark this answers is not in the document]
PascalPrecht: Yea so the menu is generated in status  desktop where there's access to information that sets these values

saledjenic:
Hey @alexandraB99 @PascalPrecht since I tried to use this PR in the main app and tried to have all that working I faced two issues which I would like you to update:

1. (minor) there is no vertical scroll bar in the search result list.
2. (blocker) the location menu (which determines where the search should be applied) must be created dynamically (using model/instantiator), currently there is no way to create content of the StatusPopupMenu component dynamically. That is a must cause that items/subitems should represent the current state in the app in terms of communities/chats/channels.

alexandraB99:
@saledjenic I have added the Scrollbar and also switched to use Instantiator so that the model can be set dynamically. There are a few TODO on my side, but this should give you a heads up for now for you to work. One question: what is the thing to check on the model in order to determine whether contains a submenu or not?

@PascalPrecht I believe I have addressed all comments now. The bug found with the image is not relevant anymore, since the menu is added dynamically, I have also removed the properties. I will switch to use IconSettings in ListItemBadge tomorrow in a separate commit.

saledjenic:
@alexandraB99 @PascalPrecht two more things:

1. the UI should be able to render a menu based on model structure like this one:
```
ListModel {
        ListElement { text: "Chat"
            imageSource: ""
            iconName: "private-chat"
            identiconColor: ""
            isIdenticon: false
            subItems: [
                ListElement {
                    text: "Pascal"
                    imageSource: "https://pbs.twimg.com/profile_images/1369221718338895873/T_5fny6o_400x400.jpg"
                    iconName: ""
                    identiconColor: ""
                    isIdenticon: false
                },
                ListElement {
                    text: "vitalik.eth"
                    imageSource: ""
                    iconName: ""
                    identiconColor: "red"
                    isIdenticon: true
                }]
        }
}
```
so the menu based on that model should look something like:
```
|
|-> Anywhere
|-> Separator
|-> (private-chat-icon) Chat ->
                               |-> (image-source-icon) Pascal
                               |-> (V-in-a-red-circle-icon) vitalik.eth
```

2. With the current implementation we're able to select only "Anywhere" or channel/chat as a location where to search in, but not able to select whole chat section (with all containing chats) or community section (with all containing channels) as a location for the search. But that's also a location for the search, see [design](https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=6042%3A1).

But for that second thing maybe we need to speak to Simon as selecting whole Chat or Community section is not quite intuitive cause they are menu which contains submenus by design.

PascalPrecht (commented):
Regarding "#" in the popup menu:

Not sure if you're already doing this, but the "#" for community channel names in the popup menu should be created using `iconName: "channel"` and not be `"#"` in concatenated to the name either

PascalPrecht:
@saledjenic @alexandraB99 here you can see what I'm referring to in https://github.com/status-im/StatusQ/pull/301#pullrequestreview-727237770 : https://github.com/status-im/StatusQ/pull/301/commits/9fe6e0a32a9cc0864e4bfecac1f7cff920d5189b#diff-2f615fc01dbb3829146521d69af20abe6ee2e0bc519dde836e087eb2694f4be5R262

saledjenic:
> Regarding "#" in the popup menu:
> 
> Not sure if you're already doing this, but the "#" for community channel names in the popup menu should be created using `iconName: "channel"` and not be `"#"` in concatenated to the name either

@PascalPrecht no I didn't add it as an icon, but as a concatenated character in the model's items. Ok, will update that. Thanks.

PascalPrecht:
🎉 It has landed!

PR #304, opened 2021-07-30
Status form control validator pipeline

PascalPrecht:
Here's what the draft is doing.

### New validators API for StatusInput to apply validators declaratively

There's a new `validators` property that can be used to add validators to `StatusInput` instances, which are executed in the same order:

```qml
StatusInput {
  text: "Some value"
  validators: [...]
}
```

### Built-in validators

For convenience StatusQ provides some common validation methods, such as `StatusMinLengthValidator`, StatusMaxLengthValidator` and could be extended to other (e.g. email validation etc):

```qml
StatusInput {
  text: "Some value"
  validators: [
    StatusMinLengthValidator { minLength: 3 }
  ]
}
```

### Validators execution and errors

Validators are executed every time the text of the input changes. They are executed in the same order they have been applied, which enables users to create cascading conditions like "First make sure the value is at least 3 characters long, then make sure it matches a certain pattern".

When a validation fails, it sets the validity of the input (`valid: false`) accordingly and optionally exposes additional error information on `StatusInput.errors`:

```qml
StatusInput {
  text: "Fo"
  onTextChanged: {
    if (errors) {
      /**
       * errors now has the following structure:
       * errors: {
       *   minLength: { minValue: 3, actual: ... }
       * }
       * Also, `StatusInput` is now `valid = false`
       **/ 
       errorMesssage = "Expected " + errors.minLenght.minValue + " but got: "+ errors.minLength.actual; // i18n'able
    }
  }
  validators: [
    StatusMinLengthValidator { minLength: 3 }
  ]
}
```
There can be any number of error objects on the `errors` property, depending on who many validators have been run that failed validation.

### Custom validators

Custom validators can be implemented by introducing a new `StatusValidator` type that has to implement a `validate()` function and defines the validators name. The `validate()` function has to return either `true` or `false` depending on whether the value is valid.

Alternatively, the function can return an error object which gets exposed on the underlying input's `errors` property, at which point it's considered invalid as well.

Here's a simple custom validator:

```qml
// HelloValidator.qml
import StatusQ.Controls.Validators 0.1

StatusValidator {

  property string name: "hello"

  validate: function (value) { // `value` is the `text` value of the underlying control
    return value === "hello"
  }
}
```

Applying this validators would look like this:

```qml
StatusInput {
  text: "Some value"
  validators: [
    HelloValidator {}
  ]
  onTextChanged: {
    if (errors.hello) {
      errorMessage = "Doesn't say hello!"
    }
  }
}
```

Alternatively, validators can return error objects to provide more information about what went wrong. Here's the implementation of the `StatusMinLengthValidator` as an example:

```qml
StatusValidator {

    property int minLength: 0

    name: "minLength"

    validate: function (value) {
        return value.length >= minLength ? true : {
            min: minLength,
            actual: value.length
        }
    }
}
```

Because validators as components, they can hold any custom properties they need to be configured.

### Maintaining error messages

There has been concern that, with this API, error messages need to be potentially defined in multiple places, given that there could be multiple instances of any validator. This is easily solved by having a centralized function figure out what the error message is, given a certain error object:

```qml
StatusInput {
  validators: [
    StatusMinLengthValidator { minLength: 3 }
  ]
  onTextChanged: {
    if (errors) {
      errorMessage = getErrorMessage(errors) // this function is provided by global or elsewhere
    }
  }
}
```

### DIFF
--- a/sandbox/StatusInputPage.qml
+++ b/sandbox/StatusInputPage.qml
@@ -4,6 +4,7 @@ import QtQuick.Layouts 1.14
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
+import StatusQ.Controls.Validators 0.1
 
 import Sandbox 0.1
 
@@ -64,6 +65,22 @@ Column {
         input.clearable: true
         input.valid: false
         input.placeholderText: "Placeholder"
+    }
+
+    StatusInput {
+        label: "Label"
+        charLimit: 30
+        input.placeholderText: "Input with validator"
+
+        validators: [
+            StatusMinLengthValidator { minLength: 10 }
+        ]
+
+        onTextChanged: {
+            if (errors && errors.minLength) {
+                errorMessage = `Value can't be shorter than ${errors.minLength.min} but got ${errors.minLength.actual}`
+            }
+        }
     }
 
     StatusInput {
--- a/src/StatusQ/Controls/StatusInput.qml
+++ b/src/StatusQ/Controls/StatusInput.qml
@@ -3,6 +3,7 @@ import QtQuick 2.14
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Controls 0.1
+import StatusQ.Controls.Validators 0.1
 
 Item {
     id: root
@@ -22,9 +23,14 @@ Item {
                   0) + 8
 
     property alias input: statusBaseInput
+    property alias valid: statusBaseInput.valid
+    property alias text: statusBaseInput.text
     property string label: ""
     property int charLimit: 0
     property string errorMessage: ""
+    property list<StatusValidator> validators
+
+    property var errors: ({})
 
     StatusBaseText {
         id: label
@@ -68,6 +74,24 @@ Item {
         anchors.rightMargin: 16
 
         maximumLength: root.charLimit
+
+        onTextChanged: {
+            if (root.validators.length) {
+                for (let idx in root.validators) {
+                    let result = root.validators[idx].validate(statusBaseInput.text)
+
+                    if (typeof result === "boolean" && result) {
+                        statusBaseInput.valid = true
+                    } else {
+                        if (!root.errors) {
+                            root.errors = {}
+                        }
+                        root.errors[root.validators[idx].name] = result
+                        statusBaseInput.valid = false
+                    }
+                }
+            }
+        }
     }
 
     StatusBaseText {
--- a/src/StatusQ/Controls/Validators/StatusMinLengthValidator.qml
+++ b/src/StatusQ/Controls/Validators/StatusMinLengthValidator.qml
@@ -1,0 +1,16 @@
+import StatusQ.Controls 0.1
+
+StatusValidator {
+
+    property int minLength: 0
+
+    name: "minLength"
+
+    validate: function (value) {
+        return value.length >= minLength ? true : {
+            min: minLength,
+            actual: value.length
+        }
+    }
+}
+
--- a/src/StatusQ/Controls/Validators/StatusValidator.qml
+++ b/src/StatusQ/Controls/Validators/StatusValidator.qml
@@ -1,0 +1,11 @@
+import QtQuick 2.13
+
+QtObject {
+    id: statusValidator
+
+    property string name: ""
+
+    property var validate: function (value) {
+        return true
+    }
+}
--- a/src/StatusQ/Controls/Validators/qmldir
+++ b/src/StatusQ/Controls/Validators/qmldir
@@ -1,0 +1,5 @@
+module StatusQ.Controls.Validators
+
+StatusValidator 0.1 StatusValidator.qml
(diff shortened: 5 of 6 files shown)


PascalPrecht:
> If we use on multiple places StatusInput component with the same validator StatusMaxLengthValidator { maxLength: 30 } since we need to create error message on the place of caller that may lead us having different messages across the app for the same validation. That's why I suggested on Discord when we discussed that, starting here, to have single file with all possible validators and use "or" of enum values to define set of validations which need to be applied over some input.

@saledjenic thanks again for your feedback. So your concern is that the same error message might need to be implemented in multiple places. Generally this is true, although, this is easily solved if you create such a universal function (as you mentioned) and you pass it the `StatusInput.errors` object. That way you can keep your error messages in a single place.

I'm very open to other solutions for applying validations as long as the following conditions are true:

1. StatusQ form controls come with a validation API
2. Implementing validation on form controls should be as easy as saying: Put this control here and apply *that* validator.
3. StatusQ controls should not know anything about any particular validator, but, as pointed out in 1), provide an API so that any number of validators can be applied to it.
4. StatusQ controls should know when they are valid, and if they are not what errors are attached to them.
5. Particular error messages can't be the concern of the validators because StatusQ doesn't handle i18n. That's the application's responsibility. However, StatusQ ensures that all components are internationalizable (in this case, by exposing error data on inputs which could then be used in error messages)

I see the convenience-side of what you've implemented and I'd probably do it the same way if we wouldn't have to worry about maintaining a reusable component library that aims to be used by various applications, not just Status Desktop by the Status team.  

That being said you can always build a custom validator that does exactly what your proposal is doing, thanks to the API provided in this PR draft.

> What I would change is... when you e.g. set a limit of 30 chars for an input... you're allowed to type more than 30 characters getting an error that only 30 is allowed, but your character's counter looks like "31/30" what is a bit weird I would say? What do you think? If you type more, you may get to like "100/30", so what I would do in this case... when user try to enter 31st character we should not allow him to do that, but keep 30 characters in that input area and display warning message that only 30 is allowed, what do you think? In that case character's counter will go from 0/30 to 30/30 what will be the max.

Yea, that's a fair point. Not really scope of this PR though. The `charlimit` property is first and foremost just a UI related property and renders that counter you're seeing. Now we could say that this should also cause the `StatusInput` to stop receiving any more input if the limit is reached. Another way to look at it is to have the application decide what to do when the limit is reached. Some applications would rather call a validator, which would require the value to get over the limit. This can't be done anymore if `StatusInput` comes with the other behaviour built-in.

I guess I'm fine either way, however, since most of StatusQ is driven by Status Desktop at the moment, I'd probably advocate for the second option because the designs show validation messages on inputs where the char limit is reached.

PascalPrecht:
Updated the draft description to give more insights on how this is supposed to work

sandbox/StatusInputPage.qml
```diff
@@ -66,6 +67,22 @@ Column {
         input.placeholderText: "Placeholder"
     }
 
+    StatusInput {
+        label: "Label"
+        charLimit: 30
+        input.placeholderText: "Input with validator"
+
+        validators: [
+            StatusMaxLengthValidator { maxLength: 30 }
```
borismelnik, line 76: i think it should be done by `maximumLength` property. Looks like user can add text more than 30 chars but its not acceptable. When this option can be helpful?

PascalPrecht: Yes, as explained above, this is to have room for a validator to kick in. Will double check with designs, but happy to roll back on the max length validator (which isn't needed anymore  if we have the limit enforced built-in)

PascalPrecht: This is addressed here: https://github.com/status-im/StatusQ/pull/321

PascalPrecht:
Sent a new PR which will enforce the `charLimit` on `StatusInput` via `StatusBaseInput.maximumLength`: https://github.com/status-im/StatusQ/pull/321

This will make the `StatusMaxLengthValidator` obsolete in this PR so I'm gonna remove it.

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -68,6 +74,25 @@ Item {
...
+
+                    if (typeof result === "boolean" && result) {
+                        statusBaseInput.valid = true
+                    } else {
+                        if (!root.errors) {
+                            root.errors = {}
+                        }
+                        root.errors[root.validators[idx].name] = result
+                        statusBaseInput.valid = false
+                        break;
```
alaibe, line 91: If I understand correctly, it means that as soon as we encounter an error, we stop running subsequent validator?

I think we should let the decision to render all error or only the first one to the Error component (if we built any). But as an input I think he should still run all validators.

What do you think?

alaibe (commented):
API looks clean and easy to understand to me. Nice job. Only 1 question below

src/StatusQ/Controls/StatusInput.qml
PascalPrecht: Ah good catch! We shouldn't break there. Will update this!

PR #305, opened 2021-07-30
fix(StatusChatListItem): don't signal item selection if already selected

### LINKED ISSUE #303: `StatusChatList` invokes signal `chatItemSelect` for already selected item
When you click on the same chat multiple times, it gets reloaded, because the signal is emitted on every click ignoring the fact the item/chat is already selected.

PascalPrecht:
Closes #303


pinebit (approved):

pinebit:
Thank you for fixing this!

borismelnik (approved):

PR #309, opened 2021-08-03
fix(StatusChatToolBar) Fixing more menu not closing

### LINKED ISSUE #308: Clicking the top right kebab button a second time doesn't close the menu
Fixing https://app.zenhub.com/workspaces/nim-status-client--nim-status-5ea70d29c12ee928df5ae848/issues/status-im/status-desktop/3129

alexandraB99:
The menu has a CloseOnReleaseOutside policy and so it
was closing and immediately re-opened when the kebab icon
was clicked since it's outside the menu area and also was
calling the popup function of the menu. Added dummy bool
property to detect whether the menu is already closed and
not open it again

Closes #308

### DIFF
--- a/src/StatusQ/Components/StatusChatToolBar.qml
+++ b/src/StatusQ/Components/StatusChatToolBar.qml
@@ -88,10 +88,17 @@ Rectangle {
             type: StatusFlatRoundButton.Type.Secondary
             visible: !!statusChatToolBar.popupMenu
 
+            property bool showMoreMenu: false
             onClicked: {
-                statusChatToolBar.menuButtonClicked()
+                if (showMoreMenu) {
+                    popupMenuSlot.item.popup(-popupMenuSlot.item.width + menuButton.width, menuButton.height + 4)
+                }
                 highlighted = true
-                popupMenuSlot.item.popup(-popupMenuSlot.item.width + menuButton.width, menuButton.height + 4)
+                statusChatToolBar.menuButtonClicked()
+            }
+
+            onPressed: {
+                showMoreMenu = !showMoreMenu;
             }
 
             StatusToolTip {


src/StatusQ/Components/StatusChatToolBar.qml
```diff
@@ -88,10 +88,17 @@ Rectangle {
...
+                if (showMoreMenu) {
+                    popupMenuSlot.item.popup(-popupMenuSlot.item.width + menuButton.width, menuButton.height + 4)
+                }
                 highlighted = true
-                popupMenuSlot.item.popup(-popupMenuSlot.item.width + menuButton.width, menuButton.height + 4)
+                statusChatToolBar.menuButtonClicked()
+            }
+
+            onPressed: {
+                showMoreMenu = !showMoreMenu;
```
PascalPrecht, line 101: Can't we check for `popupMenuSlot.item.opened` instead of introducing a new `showMoreMenu` property?

PascalPrecht:
Bonus points if this follows the projects commit message conventions so we have the fix picked up in our automatic changelog generation

src/StatusQ/Components/StatusChatToolBar.qml
alexandraB99: It's explained in the commit message, the problem is that popup has a CloseOnReleaseOutside policy so it will anyway close if the user clicks in the 'more' button but since this button calls the popup function again it will re-open, hence something like this
```
                if (!popupMenuSlot.item.opened) {
                    popupMenuSlot.item.popup(-popupMenuSlot.item.width + menuButton.width, menuButton.height + 4)
                }
```
it will always open the popup.

PascalPrecht (approved):

PR #313, opened 2021-08-09
Issue 264 fixes

PascalPrecht:
@alexandraB99 here are some fixes that you can review. We'll squash them into your PR once you're good with these.

There are a few more which I'll add on top of this


PascalPrecht:
Here's another thing I've found. Message results don't render properly in darkmode (font color needs to change)

Screenshot 2021-08-09 at 14 16 51

PR #315, opened 2021-08-11
fix(@desktop/chat) Acitvity tooltip arrow position is incorrect when members list is shown

Khushboo-dev-cpp:
Added a property to check if a tool tip is a corner tool tip and in that case arrow position will be measured differently.

fixes #3102

### DIFF
--- a/src/StatusQ/Components/StatusChatToolBar.qml
+++ b/src/StatusQ/Components/StatusChatToolBar.qml
@@ -17,6 +17,8 @@ Rectangle {
     property alias membersButton: membersButton
     property alias searchButton: searchButton
     property int notificationCount: 0
+    // Used to understand if the tooltip is in the corner of the screen
+    property bool isCorner: true
 
     property Component popupMenu
 
@@ -166,7 +168,7 @@ Rectangle {
                 text: "Activity"
                 orientation: StatusToolTip.Orientation.Bottom
                 y: parent.height + 12
-                arrow.x: width - (parent.width-4)
+                cornerToolTip: isCorner
             }
         }
 
--- a/src/StatusQ/Controls/StatusToolTip.qml
+++ b/src/StatusQ/Controls/StatusToolTip.qml
@@ -13,6 +13,7 @@ ToolTip {
         Right
     }
 
+    property bool cornerToolTip: false
     property int maxWidth: 800
     property int orientation: StatusToolTip.Orientation.Top
     property alias arrow: arrow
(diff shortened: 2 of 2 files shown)


src/StatusQ/Components/StatusChatToolBar.qml
```diff
@@ -166,7 +168,7 @@ Rectangle {
                 text: "Activity"
                 orientation: StatusToolTip.Orientation.Bottom
                 y: parent.height + 12
-                arrow.x: width - (parent.width-4)
+                cornerToolTip: isCorner
```
PascalPrecht, line 171: 2 thoughts here:

1. I think `cornerToolTip` is not a good name for this API. Looking at the tooltip standalone, it doesn't make too much sense. Maybe there are even other use cases where the same visuals are desired and we're not dealing with a window "corner".  How about `offset` which lets you specify the actual offset of the tooltip (while the arrow is always centered relative to the trigger?

2. Instead of introducing an `isCorner` API on the toolbar (which again, is very contextually named), how about we alias the tooltip on the `notificationButton` as `tooltip`? Then consumers of `StatusChatToolBar` can do something like:

```qml
StatusChatToolBar {
  notificationButton.tooltip.offset: ...  // notificationButton is already exposed
}
```

This even opens possibilities to other issues, where right now the tooltips aren't i18n'able.

With this change we can then do:

```qml
StatusChatToolBar {
  notificationButton.tooltip.text = "..."
}
```

Using `offset` as described above would also allow users to offset the tooltip in both directions if needed.

PascalPrecht (changes requested):
Thanks for this fix!

I'd like us to make this API a bit more generic though and introduce it as a feature of `StatusToolTip` first and then applying the fix in `StatusChatToolBar` (2 commits).

Once that is done, a fix can be made in Status Desktop which references the issue you've mentioned in the commit message here.

Also, keep in mind that the scope you're using in your message (`@desktop/chat`) doesn't apply in this project.

Basically what we'd end up with instead would be two changes:

1. `feat(StatusToolTip): introduce offset API`
2. `fix(StatusChatToolBar) expose tooltip on action buttons`

Followed by a commit in Status Desktop that will use a version of StatusQ that includes this feature + fix

src/StatusQ/Components/StatusChatToolBar.qml
PascalPrecht: Oh and lastly: you can then also apply whatever is needed in `DemoApp.qml` so that everything looks natural there (as in, the tooltip doesn't need an offset in the toolbar because it's not super close to the window)

Khushboo-dev-cpp: Hey @PascalPrecht makes sense the points from you. I was only thinking of corner uses cases. But the StatusToolTip as widget definitely needs a more usable API from the outside. Will rework on this.

PR #318, opened 2021-08-12
Feat/StatusToolTip

Khushboo-dev-cpp:
feat(StatusToolTip) adding offset property
fix(StatusChatToolBar) exposing the tooltip form the toolbar

### DIFF
--- a/src/StatusQ/Components/StatusChatToolBar.qml
+++ b/src/StatusQ/Components/StatusChatToolBar.qml
@@ -56,12 +56,10 @@ Rectangle {
             type: StatusFlatRoundButton.Type.Secondary
             onClicked: statusChatToolBar.searchButtonClicked()
 
-            StatusToolTip {
-                visible: !!text && parent.hovered
-                text: "Search"
-                orientation: StatusToolTip.Orientation.Bottom
-                y: parent.height + 12
-            }
+            // initializing the tooltip
+            tooltip.text: "Search"
+            tooltip.orientation: StatusToolTip.Orientation.Bottom
+            tooltip.y: parent.height + 12
         }
 
         StatusFlatRoundButton {
@@ -72,12 +70,10 @@ Rectangle {
             type: StatusFlatRoundButton.Type.Secondary
             onClicked: statusChatToolBar.membersButtonClicked()
 
-            StatusToolTip {
-                visible: !!text && parent.hovered
-                text: "Members"
-                orientation: StatusToolTip.Orientation.Bottom
-                y: parent.height + 12
-            }
+            // initializing the tooltip
+            tooltip.text: "Members"
+            tooltip.orientation: StatusToolTip.Orientation.Bottom
+            tooltip.y: parent.height + 12
         }
 
         StatusFlatRoundButton {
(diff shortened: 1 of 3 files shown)


src/StatusQ/Components/StatusChatToolBar.qml
```diff
@@ -14,6 +14,7 @@ Rectangle {
     property alias chatInfoButton: statusChatInfoButton
     property alias menuButton: menuButton
     property alias notificationButton: notificationButton
+    property alias notificationTooltip: notificationButton.tooltip
```
Khushboo-dev-cpp, line 17: @PascalPrecht  This is something I am not happy with. if any of the other buttons tooltip need modification for future features, then we need to expose each of these individually. if instead all these button came form a common file such as the 
-> "StatusFlatRoundButtonWithTooltip" or 
-> adding the tooltip in the StatusFlatRoundButton and it will only be visible if someone adds text for the tooltip. w

then these could be exposed directly whenw e expose the button from the StatusChatToolBar

PascalPrecht: Okay so I actually thought we should expose the tooltip on the `notificationButton` which is already exposed on the chat toolbar.

That way no new property in chat toolbar is needed. Again, the end result would be:

```
StatusChatToolBar {
  notificationButton.tooltip.[...] = ...
}
```
This can be done for all buttons exposed by the toolbar to be consistent.

As discussed offline, we can explore the idea of adding tooltips to `StatusBaseButton` so all buttons get them for free. But that's a new feature which is not scope of this PR.

src/StatusQ/Components/StatusChatToolBar.qml
```diff
@@ -137,6 +138,8 @@ Rectangle {
             icon.height: 21
             type: StatusFlatRoundButton.Type.Secondary
 
+            property alias tooltip: tooltip
```
PascalPrecht, line 141: This is enough, no need to expose it explicitly on the chat toolbar

Please do the same for all other action buttons as well.

PascalPrecht (changes requested):
Thanks @Khushboo-dev-cpp !

There was a bit of a misunderstanding here. Exposing the tooltip on the button is enough because the buttons are already exposed on the toolbar. No need to introduce a new property on the toolbar.

Also can you please fix your commits to follow the correct convention?

E.g.

```
feat(StatusToolTip): introduce offset property
```

And the second commit is not actually a fix in the `StatusToolTip` component but in `StatusChatToolBar`

We want this to properly be picked up by our changelog generation tool so the format has to be correct.

src/StatusQ/Components/StatusChatToolBar.qml
Khushboo-dev-cpp: This wont work because of the way in which QMl does binding. It does not let you access the property newly created in this component but links it the actual StatusToolTip.qml component. The only way this can work is if it in a file and that file is then aliased!
You could try it out very simply. wont work!

PascalPrecht: Ah, then what if we do `property var tooltip` instead and assign the tooltip in the buttons `onCompleted` handler (for now). Once we've landed tooltips in StatusBaseButton we can refactor this (essentially just removing code).

PascalPrecht: As discussed offline, @Khushboo-dev-cpp pointed out that this solution will result with the same issue as the other approach. We'll roll back and put tooltip instances into flat round button so they are a natural property of the buttons which are then automatically exposed via chat tool bar.

Khushboo-dev-cpp: Done!

src/StatusQ/Components/StatusChatToolBar.qml
PascalPrecht: This property can now be removed :)

Khushboo-dev-cpp: Done!

PR #319, opened 2021-08-13
Status search popup component updates

saledjenic:
Changes needed for the desktop part of issue-2934 are added.

### DIFF
--- a/sandbox/DemoApp.qml
+++ b/sandbox/DemoApp.qml
@@ -567,8 +567,9 @@ Rectangle {
                     chatInfoButton.type: StatusChatInfoButton.Type.CommunityChat
                     onSearchButtonClicked: {
                         searchButton.highlighted = !searchButton.highlighted;
-                        searchPopup.searchSelectionButton.primaryText = demoCommunityDetailModal.header.title;
-                        searchPopup.searchSelectionButton.image.source = demoCommunityDetailModal.header.image.source;
+                        searchPopup.setSearchSelection(demoCommunityDetailModal.header.title,
+                                                       "",
+                                                       demoCommunityDetailModal.header.image.source);
                         searchPopup.open();
                     }
                     membersButton.onClicked: membersButton.highlighted = !membersButton.highlighted
--- a/sandbox/DemoApp.qml
+++ b/sandbox/DemoApp.qml
@@ -618,7 +618,7 @@ Rectangle {
                 StatusSearchLocationMenu {
                     id: searchPopupMenu
                     searchPopup: searchPopup
-                    locatioModel: models.optionsModel
+                    locationModel: models.optionsModel
                 }
             }
 
--- a/src/StatusQ/Components/StatusChatListCategoryItem.qml
+++ b/src/StatusQ/Components/StatusChatListCategoryItem.qml
@@ -23,14 +23,11 @@ StatusListItem {
     property alias menuButton: menuButton
     property alias toggleButton: toggleButton
 
-    signal clicked(var mouse)
     signal addButtonClicked(var mouse)
     signal menuButtonClicked(var mouse)
     signal toggleButtonClicked(var mouse)
 
     color: sensor.containsMouse || highlighted ? Theme.palette.baseColor2 : "transparent"
-
-    sensor.onClicked: statusChatListCategoryItem.clicked(mouse)
 
     statusListItemTitle.color: Theme.palette.directColor4
     statusListItemTitle.font.weight: Font.Medium
--- a/src/StatusQ/Components/StatusListItem.qml
+++ b/src/StatusQ/Components/StatusListItem.qml
@@ -34,6 +34,9 @@ Rectangle {
 
     radius: 8
 
+    property string itemId: ""
+    property string titleId: ""
+
     property string title: ""
     property string titleAsideText: ""
     property string subTitle: ""
@@ -83,6 +86,9 @@ Rectangle {
 
     property list<Item> components
 
+    signal clicked(string itemId)
+    signal titleClicked(string titleId)
+
     onComponentsChanged: {
         if (components.length) {
             for (let idx in components) {
@@ -99,6 +105,10 @@ Rectangle {
         cursorShape: enabled ? Qt.PointingHandCursor : Qt.ArrowCursor
         acceptedButtons: Qt.LeftButton | Qt.RightButton
         hoverEnabled: true
+
+        onClicked: {
+            statusListItem.clicked(statusListItem.itemId)
+        }
 
         Loader {
             id: iconOrImage
@@ -186,6 +196,15 @@ Rectangle {
                       case StatusListItem.Type.Danger:
                           return Theme.palette.dangerColor1
                   }
+                }
+
+                MouseArea {
+                    anchors.fill: parent
+                    cursorShape: containsMouse? Qt.PointingHandCursor : Qt.ArrowCursor
+                    hoverEnabled: true
+                    onClicked: {
+                        statusListItem.titleClicked(statusListItem.titleId)
+                    }
                 }
             }
 
--- a/src/StatusQ/Popups/StatusSearchPopup.qml
+++ b/src/StatusQ/Popups/StatusSearchPopup.qml
@@ -19,21 +19,32 @@ StatusModal {
 
     property string searchText: contentComponent.searchText
     property string noResultsLabel: "No results"
+    property string defaultSearchLocationText: "Anywhere"
     property bool loading
     property Menu searchOptionsPopupMenu: Menu { }
     property var searchResults: [ ]
     property var searchSelectionButton
-    function resetSelectionBadge() {
-        searchSelectionButton.iconSettings.name = ""
-        searchSelectionButton.iconSettings.isLetterIdenticon = false
-        searchSelectionButton.iconSettings.color = "transparent"
-        searchSelectionButton.image.source = ""
-        searchSelectionButton.image.isIdenticon = false
-        searchSelectionButton.primaryText = qsTr("Anywhere");
-        searchSelectionButton.secondaryText = "";
 
     signal resultItemClicked(string itemId)
     signal resultItemTitleClicked(string titleId)
+
+    function setSearchSelection(text = "",
+                                secondaryText = "",
+                                imageSource = "",
+                                isIdenticon = "",
+                                iconName = "",
+                                iconColor = "") {
+        searchSelectionButton.primaryText = text
+        searchSelectionButton.secondaryText = secondaryText
+        searchSelectionButton.image.source = imageSource
+        searchSelectionButton.image.isIdenticon = isIdenticon
+        searchSelectionButton.iconSettings.name = iconName
+        searchSelectionButton.iconSettings.color = iconColor !== ""? iconColor : Theme.palette.primaryColor1
+        searchSelectionButton.iconSettings.isLetterIdenticon = !iconName && !imageSource
+    }
+
+    function resetSearchSelection() {
+        setSearchSelection(defaultSearchLocationText, "", "", false, "", "transparent")
     }
 
     content: Item {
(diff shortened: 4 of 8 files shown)


src/StatusQ/Components/StatusListItem.qml
```diff
@@ -34,6 +34,9 @@ Rectangle {
 
     radius: 8
 
+    property string itemId: ""
```
PascalPrecht, line 37: So this property only really makes sense in the search results. It shouldn't be part of `StatusListItem` by default because there's no natural id for any list item.

Can we move that inline to the `StatusListItem` used in `StatusSearchPopup` instead?
I'd also be fine with extracting that one into its own `StatusSearchResultItem` component.

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -34,6 +34,9 @@ Rectangle {
...
+    property string titleId: ""
```
PascalPrecht, line 38: What's this one about? Can you provide some context on why this is needed? Just by looking at it without context, it doesn't make a lot of sense but I'm sure you have your reasons

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -83,6 +86,9 @@ Rectangle {
 
     property list<Item> components
 
+    signal itemClicked(string itemId)
```
PascalPrecht, line 89: If your goal is to receive an `onClicked` signal, you can rely on the exposed `sensor` property. You can then propagate `itemId` once its moved to `StatusListItem` in `StatusSearchPopup`.

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -187,6 +197,15 @@ Rectangle {
...
                 }
+
+                MouseArea {
+                    anchors.fill: parent
+                    cursorShape: containsMouse? Qt.PointingHandCursor : Qt.ArrowCursor
+                    hoverEnabled: true
+                    onClicked: {
+                        statusListItem.titleClicked(statusListItem.titleId)
+                    }
+                }
```
PascalPrecht, line 208: I see why this is useful, just need to clarify `titleId`, maybe there's a cleaner way to get the same results.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -18,19 +18,33 @@ StatusModal {
     showFooter: false
 
     property string searchText: contentComponent.searchText
-    property string noResultsLabel: "No results"
+    property string noResultsLabel: qsTr("No results")
+    property string firstItemText: qsTr("Anywhere")
```
PascalPrecht, line 22: How about calling this `defaultSearchLocationText`?

PascalPrecht: Also, no need to `qsTr()` them here because it's the applications reponsibility to i18n app messages.

src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -272,6 +289,14 @@ StatusModal {
                         badge.image.source: model.badgeImage
                         badge.icon.isLetterIdenticon: model.isLetterIdenticon
                         badge.icon.color: model.badgeIdenticonColor
+
+                        onItemClicked: {
+                            root.resultItemClicked(itemId)
+                        }
```
PascalPrecht, line 295: As mentioned in the other comment, this would be `sensor.onClicked()` and `itemId` should be a property on the component here, inline'd

PascalPrecht (changes requested):
These changes make sense - thank you!

Just a few minor things to adjust. And then we need to clean up the commits

saledjenic:
Hey @PascalPrecht thanks for the comments, so let's firstly discuss your translation related comments

> How about calling this defaultSearchLocationText?

I definitely agree on this, that is more proper naming for this property (will update this).

> Also, no need to qsTr() them here because it's the applications reponsibility to i18n app messages.

Are you sure about this? That string property refers to the value we're displaying to the user, so we must translate it if we want to have correctly translated all text labels in UI when user switches the language. Translation script we're using will surely translate it correctly.

Now about all other comments, as all them refer to the same thing, so I am going to explain them:
- Known requests (so far, may be more later):
1. if user clicks a message sender's name of the item listed in the search result we need to display sender's profile popup (check that in this branch or check that in the current master)
2. is user clicks an item listed in the search result we need to jump to that community/channel/message (depends what item was clicked on)

The app to be able to do that, the search model (in our case `searchResults`) must have exposed all visible properties needed for creation a UI of the item, and also some "hidden" values which we will use in the background in order to perform mentioned functionalities (mentioned above points 1. and 2.). 

For this particular case properties I added `itemId` and `titleId` refer to the id of the displayed item and id of the sender in that item respectively. So we definitely do need them. The question is whether they need to be in the `StatusListItem` component or not?

What was my logic to put them there... `StatusListItem` component refers to the single item displayed in the search result, so hidden id of it (id used in the background logic) is belonging to id, logically, and that's why I put `itemId` there and named it like that. We may have more places where we may or may not need it, but it belongs to that component. Same logic for the `titleId`, it refer to the component which is used in `StatusListItem` component with id `statusListItemTitle` and that's why I used it and put it there.

Now in order to have some kind of dynamic I needed to add signals so a component which is using `StatusSearchPopup` is aware when the item's title is clicked and when the whole item is clicked. I agree that I might use exposed `sensor` (alias to the mouse area) instead of adding `itemClicked` signal, but there is no something similar for click on the title, and I decided to use same approach for both and added those two signals.

I don't see a point extracting `StatusListItem` into `StatusSearchResultItem` because with `itemId` and `titleId` I am referring to the whole status item and title where both are part of `StatusListItem` I am not extending it. So `StatusListItem` should be able to notify the component which is using it about changes/events it performs.

I think that I explained why do we need them, why I named those added signals and properties like that and when I use them. I hope it's clear now, but everything is a matter of style.

PascalPrecht:
Thanks for the thorough feedback!

> Are you sure about this? That string property refers to the value we're displaying to the user, so we must translate it if we want to have correctly translated all text labels in UI when user switches the language. Translation script we're using will surely translate it correctly.

We can verify with Desktop the exact details, but AFAIK all `qsTr()` are being replaced with `qsTrId()` by Desktop's i18n scripts. This means, the changes of that script will bleed into StatusQ which should not be the case. StatusQ has to provide APIs for consumers to change these things without changing the inner workings of StatusQ. Also, it's theoretically possible that apps wanna use their own i18n solution, in which case `qsTr()` isn't even meaningful. 

That's why all places where text occurs in StatusQ components, we've ensured that they are properly exposed so applications that use StatusQ get full control over how they want to i18n they're app messages.

So to summarize: no doubt desktop is gonna pick it up, but we don't want desktop to mutate the source of this third party library.

> For this particular case properties I added itemId and titleId refer to the id of the displayed item and id of the sender in that item respectively. So we definitely do need them. The question is whether they need to be in the StatusListItem component or not?

That makes sense.

>  I agree that I might use exposed sensor (alias to the mouse area) instead of adding itemClicked signal, but there is no something similar for click on the title, and I decided to use same approach for both and added those two signals.

Okay, I mean it's really not a big deal after all. The reason so far I've avoided a dedicated click handler on the item itself, is because there's use cases with StatusListItem where its sensor is turned off (because it shouldn't be clickable). Coming from that perspective, we'd end up with a click handler that' rendered useless in case the sensor is turned off. To ensure we have APIs that always exist and are valid (are functioning), I went with exposing `sensor` and have the click handler being attached there.

Now I see that this isn't as easy with the title unless we expose its sensor as well (in fact, 99% of the cases we use StatusListItem the title is not clickable, do we really want an API that's 99% of the time not relevant?). 

I'll let this be your call. If we do introduce these sugar APIs can we at least rename `onItemClicked` to `onClicked` ?

saledjenic:
> The reason so far I've avoided a dedicated click handler on the item itself, is because there's use cases with StatusListItem where its sensor is turned off (because it shouldn't be clickable).

> in fact, 99% of the cases we use StatusListItem the title is not clickable, do we really want an API that's 99% of the time not relevant?

It's not a big deal if we have something but not use that all the time. There are many properties/signals attached to built in qml components which we're almost never using. So that's pretty fine.

> If we do introduce these sugar APIs can we at least rename onItemClicked to onClicked ?

This is totally fine, as that refer to click on the whole `StatusListItem` component.

To summarize (what I am going to change here):
-  will rename `firstItemText` to `defaultSearchLocationText`
- will remove `qsTr` from the properties in `StatusListItem` (just to follow the principle of statusq components), but also will set them from outside (place which is using it, but that is `StatusSearchPopup` which is also part of the statusq, so I need to expose those two from the `StatusSearchPopup` component again) again and that way we will be sure that translation will be performed.
- will rename `onItemClicked` to `onClicked`

That's it, are you ok with that?

PascalPrecht:
Awesome, let's do that!

sandbox/DemoApp.qml
```diff
@@ -618,7 +619,12 @@ Rectangle {
                 StatusSearchLocationMenu {
                     id: searchPopupMenu
                     searchPopup: searchPopup
-                    locatioModel: models.optionsModel
+                    locationModel: models.optionsModel
+
+                    onItemClicked: {
+                        console.warn("ClickedFirstLevelItemValue:", firstLevelItemValue,
+                                     " ClickedSecondLevelItemValue:", secondLevelItemValue)
+                    }
```
PascalPrecht, line 627: This should be removed.

src/StatusQ/Components/StatusChatListCategoryItem.qml
```diff
@@ -23,15 +23,12 @@ StatusListItem {
...
 
-    signal clicked(var mouse)
     signal addButtonClicked(var mouse)
     signal menuButtonClicked(var mouse)
     signal toggleButtonClicked(var mouse)
 
     color: sensor.containsMouse || highlighted ? Theme.palette.baseColor2 : "transparent"
 
-    sensor.onClicked: statusChatListCategoryItem.clicked(mouse)
-
```
PascalPrecht, line 34: Why are these removed? This is not related to the scope of this PR, plusm we need those.

PascalPrecht (changes requested):
Alright to wrap this up:

1. We need to reintroduce the click handlers for the category list item components
2. Ideally we end up with commits for each significant change (which are):

   1. `itemId`, `titleId` plus their dedicated handlers have been introduced (feature)
   2. Menu item click handlers have been replaced with `itemClicked`. Technically that's a breaking change, but I'm fine with making this a refactor commit
   3. There's a typo fix in `StatusSearchLocationMenu.locationModel`
   4. New APIs `resetSearchSelection()` and `setSearchSelection()` where introduced 
   5. Result item (and title) clicked signals have been added to search popup
   6. Status Menu item received new `value` and `parentValue` APIs

This is what I see at a quick glance. Would be wonderful if you could clean this up accordingly. Let me know if you need help with this.

Just make sure that no commit breaks anything haha

src/StatusQ/Components/StatusChatListCategoryItem.qml
saledjenic: > Why are these removed? This is not related to the scope of this PR, plusm we need those.

Since we have now "clicked" singla in `StatusListItem` and `StatusChatListCategoryItem` is inherited from `StatusListItem` we don't need it in `StatusChatListCategoryItem`, that works the same way now as it worked before.
that's because we have them now in StatusListItem

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -83,6 +86,9 @@ Rectangle {
 
     property list<Item> components
 
+    signal clicked(string itemId)
```
PascalPrecht, line 89: This breaks `StatusNavigationListItem` because it's a `StatusListItem` as well and introduces its own `clicked` handler.

PascalPrecht (changes requested):
Thanks for cleaning everything up @saledjenic !
This looks good to me and I've tested this locally as well.

Only thing is that there's a component left that's introducing its own `clicked` handler but is also a `StatusListItem`

This should be addressed in the commit that introduces the `clicked` handler on `StatusListItem`.

PascalPrecht (approved):

PascalPrecht:
Landed! Awesome work @saledjenic !

PR #322, opened 2021-08-13
Adding grouping sections in search results model

alexandraB99:
ListView doesn't sort by default all items per section, thus this should be done manually

### DIFF
--- a/src/StatusQ/Popups/StatusSearchPopup.qml
+++ b/src/StatusQ/Popups/StatusSearchPopup.qml
@@ -33,6 +33,23 @@ StatusModal {
         searchSelectionButton.secondaryText = "";
     }
 
+    onSearchResultsChanged: {
+        var curIndexes = [...Array(searchResults.count)].map((v,i) => i);
+        curIndexes.sort((a,b) => searchResults.get(a).sectionName.localeCompare(searchResults.get(b).sectionName));
+        var sorted = 0;
+        while ((sorted < curIndexes.length) && (sorted === curIndexes[sorted])) {
+            sorted++;
+        }
+        if (sorted === curIndexes.length) {
+            return;
+        }
+        for (let i = sorted; i < curIndexes.length; i++) {
+            searchResults.move(curIndexes[i], searchResults.count - 1, 1);
+            searchResults.insert(curIndexes[i], { } );
+        }
+        searchResults.remove(sorted, curIndexes.length - sorted);
+    }
+
     content: Item {
         width: parent.width
         height: root.height
--- a/sandbox/Models.qml
+++ b/sandbox/Models.qml
@@ -202,29 +202,29 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
             badgeSecondaryText: "";
             badgeIdenticonColor: "";
             isLetterIdenticon: false }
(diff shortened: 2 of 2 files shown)


src/StatusQ/Popups/StatusSearchPopup.qml
```diff
@@ -33,6 +33,23 @@ StatusModal {
...
+        }
+        if (sorted === curIndexes.length) {
+            return;
+        }
+        for (let i = sorted; i < curIndexes.length; i++) {
+            searchResults.move(curIndexes[i], searchResults.count - 1, 1);
+            searchResults.insert(curIndexes[i], { } );
+        }
+        searchResults.remove(sorted, curIndexes.length - sorted);
+    }
```
PascalPrecht, line 51: My brain is having a very hard time parsing this function.
Maybe we can add some comments here to help understand what's going on?

PascalPrecht (changes requested):
Thanks for this @alexandraB99 !
I'd be nice to have some helpful comment son that handler.

The commit itself is actually a fix in the search popup component. So let's make sure the commit message states that accordingly.

Since this also depends on `get()` API, let's first confirm with @saledjenic that we can provide it.

src/StatusQ/Popups/StatusSearchPopup.qml
saledjenic: @PascalPrecht @alexandraB99 no need at all for this part on the side of qml, cause I already handle that correctly on the nim side, and works good. 
There are 2 main reasons:
1. We should always avoid applying logic on the side of qml, we need qml part only for the UI and it should know only to render (display) what backend says to be rendered
2. This is too complex method for the qml as that search result may have even thousands of items, and there are too many passes through entire list, too many removal/insertions/shifts so it may have a performance impact at the end

PascalPrecht: > no need at all for this part on the side of qml, cause I already handle that correctly on the nim side, and works good.

I believe this was introduced because the sections aren't handled well if they aren't properly pre-ordered. Can we verify that it "works good"?

saledjenic: @PascalPrecht that's what I did, I handled them in a correct way on the side of nim and ordered sections as requested by the design, and also ordered messages by their time within "Messages" section. That's definitelly a job for the backend.

PascalPrecht: Just recalling here what @alexandraB99 mentioned (which resulted in this PR):

Apparently, when the data looks like:

```
Section: messages > ....., Section: contacts > ...., Section: somethingelse > ...., Section messages > ........
```

The search results end up as:
```
Messages
-> Data
Contacts
-> Data
SomethingElse
-> Data
Messages
-> Data
```

While it should be:

```
Messages
-> Data
Contacts
-> Data
SomethingElse
-> Data
```

I haven't experienced myself yet. If you can both verify that this works as expected without this change in the front-end then we can close this.

saledjenic: @PascalPrecht correct, if you set the model as you described we will end up having same sections twice, but that's the case using manually created model, which is by the way here just for the demo up. So all we need here is to update model for the demo app and avoid duplicate sections that way in the demo app.

But as I already mentioned we don't need any logic in the qml. A real model we're using will be well prepared in the backend and take care about ordering and sections.

alexandraB99: @PascalPrecht, after @saledjenic confirmation, if this is taken care already in the backend then there is nothing to be done, we can ignore this PR. The dummy models in sandbox are already 'sorted' manually.

PR #334, opened 2021-08-17
feat(StatusLetterIdenticon): Expose the text component

Khushboo-dev-cpp:
Aliased the text compoenent so that its color, text value can be set from outside


PascalPrecht (commented):
Changes look generally good to me, just wondering: what's the use case for this? When do we have letter identicons that derive from what's implemented?

Khushboo-dev-cpp:
> Changes look generally good to me, just wondering: what's the use case for this? When do we have letter identicons that derive from what's implemented?

Needed it for https://github.com/status-im/status-desktop/issues/2009. Check the design suggested by Simon for websites that dont have an icon

PascalPrecht:
Gotcha!

PR #336, opened 2021-08-18
fix(statuschatinfoButton) text was not elided on the right side

### LINKED ISSUE #335: StatusChatInfoButton text not elided
Relates to [Desktop #3242](https://app.zenhub.com/workspaces/nim-status-client--nim-status-5ea70d29c12ee928df5ae848/issues/status-im/status-desktop/3242)

alexandraB99:
Closes #335

### DIFF
--- a/src/StatusQ/Components/StatusChatToolBar.qml
+++ b/src/StatusQ/Components/StatusChatToolBar.qml
@@ -37,6 +37,7 @@ Rectangle {
         anchors.verticalCenter: parent.verticalCenter
         anchors.left: parent.left
         anchors.leftMargin: 12
+        anchors.right: actionButtons.left
         onClicked: statusChatToolBar.chatInfoButtonClicked()
     }
 
@@ -45,7 +46,7 @@ Rectangle {
         anchors.right: parent.right
         anchors.rightMargin: 8
         anchors.verticalCenter: parent.verticalCenter
-
+        width: childrenRect.width
         spacing: 8
 
         StatusFlatRoundButton {
--- a/src/StatusQ/Controls/StatusChatInfoButton.qml
+++ b/src/StatusQ/Controls/StatusChatInfoButton.qml
@@ -102,8 +102,7 @@ Rectangle {
             anchors.topMargin: statusChatInfoButtonSubTitle.visible ? 0 : 8
             anchors.left: identicon.right
             anchors.leftMargin: 8
-
-            width: statusIcon.width + chatName.anchors.leftMargin + chatName.width + (mutedIcon.visible ? mutedIcon.width + mutedIcon.anchors.leftMargin : 0)
+            anchors.right: parent.right
             height: chatName.height
 
             StatusIcon {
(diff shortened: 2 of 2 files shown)


PascalPrecht (approved):
Changes LGTM!

Can you please update your commit message to have the proper component name (casing is wrong right now). And it'd also be cool if the commit message says the fix (e.g. "ensure button text elides")

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/38b0207055f81c853830320e8b20e9532e84973b

PascalPrecht:
Had to revert the commit and reopen this PR.

Turns out that it actually breaks the UI. Need to investigate what exactly is going on but #338 is related to that.

PascalPrecht (changes requested):
Thanks for the additional changes @borismelnik !
This already looks much better. But it's still buggy...

I think we need to clarify how this should behave:

1. The goal is to make sure that the button's title is elided right
2. This requires the component to have some "max" width
3. That max width depends on its context (it's used in the `StatusChatToolBar`, but also in the above the community chat lists)
4. The component should grow with its text (like it does prior to this PR), but once it reaches the "max" width, the text should elide

There are at least two more problems with this PR as of right now:

1. The button now takes the entire available space of the chat toolbar, even though it's not needed. This looks especially off on bigger screens when users increase the window size of the app.
Screenshot 2021-08-25 at 10 16 47
2. It seems the icon for pinned messages is pushed out of the button (look closely under the menu button below)
Screenshot 2021-08-25 at 10 18 10

borismelnik:
> Thanks for the additional changes @borismelnik !
> This already looks much better. But it's still buggy...
> 
> I think we need to clarify how this should behave:
> 
> 1. The goal is to make sure that the button's title is elided right
> 2. This requires the component to have some "max" width
> 3. That max width depends on its context (it's used in the `StatusChatToolBar`, but also in the above the community chat lists)
> 4. The component should grow with its text (like it does prior to this PR), but once it reaches the "max" width, the text should elide
> 
> There are at least two more problems with this PR as of right now:
> 
> 1. The button now takes the entire available space of the chat toolbar, even though it's not needed. This looks especially off on bigger screens when users increase the window size of the app.
> 
> Screenshot 2021-08-25 at 10 16 47
> 
> 2. It seems the icon for pinned messages is pushed out of the button (look closely under the menu button below)
> Screenshot 2021-08-25 at 10 18 10

Hey @PascalPrecht ! Sorry i was focused on binding loop bug and did not touch other implementation. Binding loop  happens when texts started using parent sizes but parent sizes based on texts sizes. 
If you can give me a link on UI i can fix all.

PascalPrecht:
No worries @borismelnik !

There's no link. There's just these rules:

> The goal is to make sure that the button's title is elided right when it's too long
This requires the component to have some "max" width
That max width depends on its context (it's used in the StatusChatToolBar, but also in the above the community chat lists)
The component should grow with its text (like it does prior to this PR), but once it reaches the "max" width, the text should elide

PR #341, opened 2021-08-20
fix(StatusChatListCategory): emit original mouse event data in clicke…

### LINKED ISSUE #333: `StatusChatListCategory` throws reference error when clicked
```
c:/StatusQ/src/StatusQ/Components/StatusChatListCategory.qml:38: ReferenceError: mouse is not defined
```

PascalPrecht:
…d signal

This fixes a bug that was introduced in https://github.com/status-im/StatusQ/commit/01da7508995de15e45703262e97b60f84fe64c03 with a breaking change
where `StatusChatListCategoryItem`'s `clicked` signal would no longer receive
a `mouse` event object.

The reason this is happening is because we've introduced a new `clicked` signal
on `StatusListItem`. The idea was to rely on that within `StatusChatListCategoryItem`,
however, we didn't take into account that the new `clicked` signal in `StatusListItem`
doesn't emit a `mouse` event object.

To fix this issue could either reintroduce the custom `clicked` handler on
`StatusChatListCategoryItem`, or listen to the original `sensor.onClicked` and
`onTitleClicked` signals exposed by `StatusListItem` instead, ensuring the required
`mouse` event data exists.

Fixes #333

### DIFF
--- a/src/StatusQ/Components/StatusChatListCategory.qml
+++ b/src/StatusQ/Components/StatusChatListCategory.qml
@@ -34,7 +34,7 @@ Column {
 
         showMenuButton: showActionButtons && !!statusChatListCategory.popupMenu
 
-        onClicked: {
+        sensor.onClicked: {
             if (mouse.button === Qt.RightButton && showActionButtons && !!statusChatListCategory.popupMenu) {
                 highlighted = true
                 popupMenuSlot.item.popup(mouse.x + 4, mouse.y + 6)
@@ -42,6 +42,8 @@ Column {
             }
             statusChatListCategory.opened = !opened
         }
+
+        onTitleClicked: statusChatListCategory.opened = !opened
         onToggleButtonClicked: statusChatListCategory.opened = !opened
         onMenuButtonClicked: {
             highlighted = true


PascalPrecht:
Another option would be to introduce the `mouse` event object as first parameter of `StatusListItem.onClicked` but that would be yet another breaking change. /cc @saledjenic

saledjenic:
> Another option would be to introduce the `mouse` event object as first parameter of `StatusListItem.onClicked` but that would be yet another breaking change. /cc @saledjenic

@PascalPrecht if you add as you proposed just a new parameter `mouse` event to `signal clicked(string itemId)` and have now new signal like `signal clicked(MouseEvent event, string itemId)` that shouldn't have any impact to the slots which are currently  using only `itemId` (as long as we don't change the name of the parameter `itemId`). And if all corresponding slots to that `clicked` signal are connected in this way:
```
onClicked: {
    // do something with itemId
}
```
So won't be a breaking change in that case.

PascalPrecht:
@saledjenic all this is true. Thing though is, when you deal with an `onClicked` event, what do developers expect the (first) supplied parameter to be? Certainly not an arbitrary `itemId`.

The reason I brought this up was because I'd rather aim for making `onClicked` behave like any built-in `onClicked` event and have `mouseEvent` be the first parameter.

We can defer this for now though and rely on `sensor.onClicked` (as done in this PR).

Issue #226, opened 2021-07-02
Tooltips in `StatusChatListAndCategories` are not internationalizable


PascalPrecht:
Rendered categories come with some action buttons, one that's supposed to let users add new channels, and another one that opens a context menu for the category.

These buttons come with a tooltip that have a default text respectively and can usually be altered via `StatusChatListCategory` like this:

```
StatusChatListCategory {
  addButton.tooltip.text = qsTr("Add channel inside category")
  menuButton.tooltip.text = qsTr("More")
}
```

At this point however, status chat list categories that are rendered inside a `Repeater` in `StatusChatListAndCategories`, don't expose an API for this.

In other words, when using that component, there's no way to change the tooltip text for the categories from the outisde.

PR #343, opened 2021-08-25
feat(QmlTranslations): Add dynamic translator to qml

borismelnik:
This ccommit introduce way of dynamic translations in qml.

Object of class QmlTranslator should be provided in qml engine as contenxt propery. Signal `languageChanged` should be connected to language system.

Example usage:
```qml

    StatusChatListCategory {
       addButton.tooltip.text = translator.emptyString + qsTr("Add channel inside category")
```

So, if we need to translate some string in JS code or in other places without bindings we should add `translator.emptyString` concatination to target string.

Closes: #226


src/StatusQ/Core/StatusTooltipSettings.qml
```diff
@@ -0,0 +1,5 @@
+import QtQuick 2.14
+
+QtObject {
+    property string text: ""
```
PascalPrecht, line 4: We might as well introduce `orientation` and `offset` here as well

src/StatusQ/Components/StatusChatListCategoryItem.qml
```diff
@@ -53,14 +59,15 @@ StatusListItem {
...
-            tooltip.text: "More"
+            tooltip.text: statusChatListCategoryItem.tooltips.menu.text
         },
         StatusChatListCategoryItemButton {
             id: toggleButton
             icon.name: "chevron-down"
             icon.width: 18
             icon.rotation: statusChatListCategoryItem.opened ? 0 : 270
             onClicked: statusChatListCategoryItem.toggleButtonClicked(mouse)
+            tooltip.text: statusChatListCategoryItem.tooltips.toggle.text
```
PascalPrecht, line 70: This one is not supposed to have a tooltip

src/StatusQ/Components/StatusChatListCategoryItem.qml
```diff
@@ -14,6 +14,12 @@ StatusListItem {
     leftPadding: 8
     rightPadding: 8
 
+    property StatusChatListCategoryItemTooltipSettings tooltips: StatusChatListCategoryItemTooltipSettings {
```
PascalPrecht, line 17: We don't need to expose them here because we already expose `addButton` and `menuButton`.

As mentioned in #226 this is about **`StatusChatListAndCategories`**. Please add an API to configure tooltips there.

src/StatusQ/Core/StatusChatListCategoryItemTooltipSettings.qml
```diff
@@ -0,0 +1,7 @@
+import QtQuick 2.14
+
+QtObject {
+    property StatusTooltipSettings add: StatusTooltipSettings {}
+    property StatusTooltipSettings menu: StatusTooltipSettings {}
+    property StatusTooltipSettings toggle: StatusTooltipSettings {}
```
PascalPrecht, line 6: These will live on `StatusChatListAndCategories`, so we should probably name them:

```
StatusChatListCategories {
  categoryAddButton.tooltip: ...
  categoryMenuButton.tooltip: ..
}
```

PascalPrecht (changes requested):

Issue #350, opened 2021-08-27
Menu button in `StatusChatToolBar` needs two clicks to be opened


PascalPrecht:
@alexandraB99 This is probably related to https://github.com/status-im/StatusQ/commit/a8e830f76caca16718353b84dfe77b5eac363943 but haven't verified this yet.

PR #357, opened 2021-08-30
fix(StatusChatToolBar): Fix mouse event catching after menu closing

borismelnik:
Closes: #350

### DIFF
--- a/src/StatusQ/Components/StatusChatToolBar.qml
+++ b/src/StatusQ/Components/StatusChatToolBar.qml
@@ -93,15 +93,19 @@ Rectangle {
 
             property bool showMoreMenu: false
             onClicked: {
-                if (showMoreMenu) {
+                if (!showMoreMenu) {
                     popupMenuSlot.item.popup(-popupMenuSlot.item.width + menuButton.width, menuButton.height + 4)
+                    statusChatToolBar.menuButtonClicked()
                 }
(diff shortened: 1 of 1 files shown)


src/StatusQ/Components/StatusChatToolBar.qml
```diff
@@ -93,15 +93,19 @@ Rectangle {
...
 
-            onPressed: {
-                showMoreMenu = !showMoreMenu;
+            Timer {
+                id: menuClosedUpdater
+                interval: 100
+                repeat: false
+                onTriggered: {
+                    menuButton.showMoreMenu = false
+                }
```
PascalPrecht, line 108: Just for me to understand:

Why is this timer necessary?

PascalPrecht (commented):
Haven't tested this yet, but I've got a question

src/StatusQ/Components/StatusChatToolBar.qml
borismelnik: i dont understand reason but menu dont have "invisible clickable area" or outside event blocker. So focused on "double clicking menu button" if you try to push on button with opened menu, the menu will close before the button receives mouse clicks. Without timer its impossible process two cases: 
-  Menu closed by button click
-  Menu closed by outside click
Thats why i add this ugly solution.

PR #359, opened 2021-08-30
feat(StatusDescriptionListItem): introduce support for `value`

PascalPrecht:
This adds a new `value` property to the component to set label for a
selected value and also rendering a chevron as an indicator that the list
item becomes clickable.


PascalPrecht:
Screenshot 2021-08-30 at 15 32 53
This is what it looks like

PR #361, opened 2021-08-30
fix(@desktop/chat): no identicon/profile picture shown and wrong time stamp format in search results

saledjenic:
This change follows model update in the `status-desktop` repo.
Relates to the issue #3259.

### DIFF
--- a/sandbox/Models.qml
+++ b/sandbox/Models.qml
@@ -209,34 +209,38 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
     //dummy search popup models
     property var searchResultsA: ListModel {
         ListElement { itemId: "i1"; titleId: "t1"; title: "@Flea"; sectionName: "Messages"; time: "18:55 AM"; content: "lorem ipsum <font color='#4360DF'>@Nick</font> dolor sit amet";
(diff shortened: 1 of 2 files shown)


PascalPrecht (changes requested):
Changes LGTM as those are primarily renaming properties in the demo models.

Would you mind updating the commit message accordingly?
There's no `@desktop/chat` scope in StatusQ. Also the changes don't actually fix anything related to identicon/profile images not showing or time stamp format.

Lemme know if you needs help with this!

saledjenic:
@PascalPrecht I looked at this as part of the main issue I fixed in `status-desktop` and that's why I added the branch with that issue number and put comment like I did, but you're right, this is separate repo (stand alone thing) and I updated the comment for this commit, check and let me know if you want something else to add there as description.

PascalPrecht:
Hey @saledjenic 

so what needs change is the commit message. Essentially what this is only doing is renaming demo model properties. There's also a change in the popup component though, which now expects a different model shape, so theoretically that's a breaking change (as in, ppl can't just upgrade StatusQ and stuff would still work. If they don't update their model the update breaks their app).

So I'd suggest something like:

```
refactor(StatusSearchPopup): change expected search model shape
```

Then also mentioning in the message that it's a breaking change so it shows up in our changelog.
Have a look at 3) here to see how this can be done: https://www.conventionalcommits.org/en/v1.0.0/#summary

PascalPrecht (approved):

PascalPrecht:
This is perfect, thank you @saledjenic !

PR #362, opened 2021-08-31
feat: introduce bigger versions of navbar icons

PascalPrecht:
Turns out the icons used in the navigation bar of the application actually
are designed to be bigger than the usual icons.

We can't just use the original source and scale them up in QML because that
will impact the stroke width of the SVGs as well, hence we need to introduce
a special set of icons that are design bigger but presever the feature ratios.


PascalPrecht:
@pinebit this also introduces bigger version of the `play` and `pause` buttons that you've requested in #320 

Once this PR lands, make sure to use `bigger/play` and `bigger/pause` instead :)

PR #366, opened 2021-08-31
feat(StatusQ.Controls): introduce StatusSwitchTabBar and StatusSwitchTab

### LINKED ISSUE #365: Create `StatusTabSwitch` components
There's a new wallet design being implemented in Status Desktop which requires a new control component that looks like this:

Screenshot 2021-08-31 at 15 07 16

In a way this is just a tabs component in a switch-style fashion.

PascalPrecht:
This commit adds two new components to the StatusQ.Controls module:

- StatusSwitchTabBar
- StatusSwitchTab

Usage:

```qml
StatusSwitchTabBar {

    StatusSwitchTab {
        text: "Tab 1"
    }

    StatusSwitchTab {
        text: "Tab 2"
    }

    StatusSwitchTab {
        text: "Tab 3"
    }
}
```

Closes #365

### DIFF
--- a/src/StatusQ/Controls/StatusSwitchTab.qml
+++ b/src/StatusQ/Controls/StatusSwitchTab.qml
@@ -1,0 +1,53 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtGraphicalEffects 1.13
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+TabButton {
+    id: control
+    contentItem: Item {
+        height: 36
+        MouseArea {
+            id: sensor
+            hoverEnabled: true
+            anchors.fill: parent
+            cursorShape: control.hovered ? Qt.PointingHandCursor : Qt.ArrowCursor
+            onClicked: control.checked = true
+
+            StatusBaseText {
+                id: label
+                text: control.text
+                color: Theme.palette.primaryColor1
+                font.weight: Font.Medium
+                font.pixelSize: 15
+                horizontalAlignment: Text.AlignHCenter
+                anchors.centerIn: parent
+            }
+        }
+    }
+
+    background: Rectangle {
+        id: controlBackground
+        implicitHeight: 36
+        implicitWidth: 148
+        color: control.checked ? 
+            Theme.palette.statusSwitchTab.backgroundColor :
+            "transparent"
+        radius: 8
+        layer.enabled: true
+        layer.effect: DropShadow {
+            width: controlBackground.width
+            height: controlBackground.height
+            x: controlBackground.x
+            source: controlBackground
+            horizontalOffset: 0
+            verticalOffset: 0
+            radius: 10
+            samples: 25
+            spread: 0
+            color: Theme.palette.dropShadow
+        }
+    }
+}
--- a/src/StatusQ/Controls/StatusSwitchTabBar.qml
+++ b/src/StatusQ/Controls/StatusSwitchTabBar.qml
@@ -3,14 +3,12 @@ import QtQuick.Controls 2.14
 import StatusQ.Core.Theme 0.1
 
 TabBar {
-    id: control
+    id: statusSwitchTabBar
     padding: 1
 
     background: Rectangle {
         implicitHeight: 36
         color: Theme.palette.directColor7
         radius: 8
-        border.width: 1
-        border.color: color
     }
 }
--- a/src/StatusQ/Controls/StatusSwitchTabBar.qml
+++ b/src/StatusQ/Controls/StatusSwitchTabBar.qml
@@ -1,0 +1,16 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import StatusQ.Core.Theme 0.1
+
+TabBar {
+    id: control
+    padding: 1
+
+    background: Rectangle {
+        implicitHeight: 36
+        color: Theme.palette.directColor7
+        radius: 8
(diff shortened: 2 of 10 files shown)


PascalPrecht:
Screenshot 2021-08-31 at 17 14 34
Screenshot 2021-08-31 at 17 14 28

@carlbennetts There haven't been designs for darkmode so please double check if this looks alright.
Also, I don't think there's an easy way to get this dropshadow without overlapping the tab bar. 
I've now made it so that you do have a drop shadow but it's subtle enough to not jump right out of the tab bar.

src/StatusQ/Controls/StatusSwitchTab.qml
```diff
@@ -0,0 +1,53 @@
...
+        implicitWidth: 148
+        color: control.checked ? 
+            Theme.palette.statusSwitchTab.backgroundColor :
+            "transparent"
+        radius: 8
+        layer.enabled: true
+        layer.effect: DropShadow {
+            width: controlBackground.width
+            height: controlBackground.height
+            x: controlBackground.x
```
borismelnik, line 43: I think than previous 3 line is not necessary bc its by default

src/StatusQ/Controls/StatusSwitchTab.qml
```diff
@@ -0,0 +1,53 @@
...
+            source: controlBackground
```
borismelnik, line 44: This too

src/StatusQ/Controls/StatusSwitchTabBar.qml
```diff
@@ -0,0 +1,16 @@
...
+TabBar {
+    id: control
+    padding: 1
+
+    background: Rectangle {
+        implicitHeight: 36
+        color: Theme.palette.directColor7
+        radius: 8
+        border.width: 1
+        border.color: color
```
borismelnik, line 14: border have a color the same as whole rectangle?

src/StatusQ/Controls/StatusSwitchTab.qml
```diff
@@ -0,0 +1,53 @@
...
+
+TabButton {
+    id: control
+    contentItem: Item {
+        height: 36
+        MouseArea {
+            id: sensor
+            hoverEnabled: true
+            anchors.fill: parent
+            cursorShape: control.hovered ? Qt.PointingHandCursor : Qt.ArrowCursor
```
borismelnik, line 16: if its `pressed` its arrow?

src/StatusQ/Controls/StatusSwitchTab.qml
```diff
@@ -0,0 +1,53 @@
...
+            onClicked: control.checked = true
```
borismelnik, line 17: why? if you set `checkable: true` then `checked` changed automatically on press

src/StatusQ/Controls/StatusSwitchTab.qml
```diff
@@ -0,0 +1,53 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+import QtGraphicalEffects 1.13
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+TabButton {
```
borismelnik, line 8: if its based on `TabButton` its should be `StatusSwitchTabButton` bc inherits `AbstractButton` functionality. `Tab` sounds like a `Page`.

borismelnik (approved):

src/StatusQ/Controls/StatusSwitchTabBar.qml
PascalPrecht: Yes. There's a padding there as well, so we'll see the border around the tabs

src/StatusQ/Controls/StatusSwitchTab.qml
PascalPrecht: Awesome

src/StatusQ/Controls/StatusSwitchTab.qml
PascalPrecht: Not sure I understand the question. It's not right now.

src/StatusQ/Controls/StatusSwitchTab.qml
PascalPrecht: > why? 
Why what exactly? 
I'm setting `control.checked` when the mouseArea emits `clicked` so that the control gets checked. The reason I've done this is because the mouse area shadows the clickable area of the tab itself, making it uncheckable otherwise.

You're saying setting `MouseArea `checkable: true` will make this work as well?

src/StatusQ/Controls/StatusSwitchTab.qml
PascalPrecht: Fair enough, will rename!

src/StatusQ/Controls/StatusSwitchTab.qml
PascalPrecht: Tried it, there's no `checkable` on `MouseArea`

carlfairclough:
@PascalPrecht thank you for covering dark mode. There might be some minor tweaks in the future but this shouldn't block us. Looks good :)

PascalPrecht:
Nice, thanks @borismelnik for you changes!

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/d449f0e903474e69bbe0e6c2acb28390ac1cf0ec

Issue #306, opened 2021-07-30
Refactor `StatusModal` to not introduce custom `content` API


PascalPrecht:
`StatusModal` comes with its own `content `property to allow consumers to implement their modal content. Usually, this is done via `contentItem` in built-in modal types. 

The reason this is not the case in `StatusModal` is because we needed to ensure that modal header and footer are always in the correct position, also taking the unknown height of the potential content into account.

After recent discussion with @alexandraB99 and her work on #301 she pointed out that we can/should make the header and footer part of the modal `background` so we can stick with `contentItem` and don't need to rely on a `Loader`.

Not relying on a `Loader` for the content also means that we no longer need the `contentComponent` introduced in: https://github.com/status-im/StatusQ/issues/237

PR #367, opened 2021-09-01
refactor(StatusModal): Remove custom content property

borismelnik:
BREAKING CHANGES:
- `content` property removed
- `Loader` inside StatusModal removed
- default `contentItem` property should be used now

Closes: #306

### DIFF
--- a/sandbox/DemoApp.qml
+++ b/sandbox/DemoApp.qml
@@ -752,7 +752,7 @@ Rectangle {
             icon.name: "notification"
         }
 
-        content: StatusBaseText {
+        contentItem: StatusBaseText {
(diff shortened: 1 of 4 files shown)


PascalPrecht (changes requested):
It seems that this change is breaking the UI for some modals (notices how the content of the modal below is shifted to the right). Looks like `contentItem` comes with a default padding?

Screenshot 2021-09-01 at 12 48 47

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/4a6800ed77c272cec02abcd8734dc28c23831e42
 
 Great job!

PR #370, opened 2021-09-01
feat(StatusExpandableSettingsItem):  Added new component for wallet s…

Khushboo-dev-cpp:
…ettings


PascalPrecht:
@Khushboo-dev-cpp would you mind adding a new page for this component in the sandbox app where the component is demonstrated so we can see how it behaves?

Khushboo-dev-cpp:
> @Khushboo-dev-cpp would you mind adding a new page for this component in the sandbox app where the component is demonstrated so we can see how it behaves?

Sure @PascalPrecht ! thanks for reminding!

Khushboo-dev-cpp:
@Pascal Done! :)

sandbox/main.qml
```diff
@@ -164,6 +164,11 @@ StatusWindow {
                             selected: page.sourceComponent == othersComponent
                             onClicked: page.sourceComponent = othersComponent
                         }
+                        StatusNavigationListItem {
+                            title: "Settings"
```
PascalPrecht, line 168: Can we call this `StatusExpandableSettingsItem`? The idea is, that eventually every component gets its own page (I know we have pages like "Others" and "Buttons" but we gotta start somewhere :D)

sandbox/qml.qrc
```diff
@@ -13,5 +13,6 @@
         <file>ThemeSwitch.qml</file>
         <file>Layout.qml</file>
         <file>Popups.qml</file>
+        <file>Settings.qml</file>
```
PascalPrecht, line 16: Can also rename this to `StatusExpandableSettingsItemPage`

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
+import QtQuick 2.14
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+
+Rectangle {
+    id: root
+
+    property alias icon: icon
```
PascalPrecht, line 10: As you can see in the usage of this component, because it's aliasing a `StatusIcon`, the effect of this is that we endup with something like:

```
Cmp {
  icon.icon:
}
```

This gets especially hairy the more component are being nested.

To solve this, we actually have `StatusIconSettings` and `StatusBackgroundIconSettings` which expose a similar shape, but instead of having `icon.icon` we'd end up with `icon.name` etc.

Could you update this please?

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+    property alias primaryText: primaryText
+    property alias secondaryText: secondaryText
```
PascalPrecht, line 12: Unless there's a very good reason, we should not expose the entire component right away. So far it looks like the title and subtitle of this component always looks the same.

Can we start with `string` properties first and then go further if necessary?

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+    property alias button: button
+    property alias expandableComponent: expandableRegion.sourceComponent
+
+    property bool isExpandable: true
+    property bool hasButton: false
+    property  bool isExpanded: false
+    property bool hasDropDownArrow: false
```
PascalPrecht, line 19: I think we can fine-tune a bunch of these:

1. `hasDropDownArrow` is something we ca probably avoid using some heuristics. For example, if the comonent is exapandable and there's no `button` we know we want to render an arrow. No need to introduce a property for this
2. `hasButton` is probably not necessary in the same vein. If there's no `button.text` we probably don't want to render it.
3. Lastly, I personally prefer `expandable` and `expanded` over `isExpandable` and `isExpanded` (also to adhere to QML built-in properties) but this is no blocker, more a personal preference. Your call :)

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+
+    anchors.left: parent.left
+    anchors.leftMargin: 20
+    anchors.right: parent.right
+    anchors.rightMargin: 20
```
PascalPrecht, line 24: Please remove the `parent` dependency and rather give it an `implicitWidth`. Consumers of this component can then decide to anchors the component if they need to. The margins would go away as well.

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+
+    radius: 8
+    color: Theme.palette.indirectColor1
```
PascalPrecht, line 27: I don't think this is the correct color. In darkmode this has a way too high contrast.
Screenshot 2021-09-02 at 10 49 17

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    StatusBaseText {
+        id: primaryText
+        anchors.top: parent.top
+        anchors.topMargin: 17
+        anchors.left: icon.right
+        anchors.leftMargin: 10
+        anchors.right: parent.right
+        anchors.rightMargin: 10
```
PascalPrecht, line 52: I might be wrong, but sticking this to `parent.right` is problematic when the button is visible as well. The text will probably go behind it.

We should make sure that it doesn't bleed into the button in case it exists.

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+        color: Theme.palette.directColor3
+    }
+
+    StatusButton {
+        id: button
+        anchors.top: parent.top
+        anchors.topMargin: 19
+        anchors.right: parent.right
+        anchors.rightMargin: 17
+        visible: root.hasButton
```
PascalPrecht, line 82: As mentioned above, we can change this to `!!text` (and the right margin should probably be rather `16` to be consistent with the rest of the components)

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    StatusIcon {
+        id: expandImage
+        anchors.top: parent.top
+        anchors.topMargin: 36
+        anchors.right: parent.right
+        anchors.rightMargin: 23
+        visible: hasDropDownArrow && isExpandable
```
PascalPrecht, line 91:
```
visible: !button.visible && expandable
```

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
...
+        anchors.top: secondaryText.bottom
+        anchors.topMargin: 16
+        anchors.left: parent.left
+        anchors.leftMargin: 48
+        width: parent.width - 64
+    }
+
+    MouseArea {
+        anchors.fill: parent
+        enabled: !hasButton
```
PascalPrecht, line 107: It looks like this isn't working properly.

When there's only a button and the component is not expandable, most likely only the button should be clickable, not the entire component. However, we still see the cursor turning into a pointing hand cursor.

PascalPrecht (changes requested):
Amazing work @Khushboo-dev-cpp !
There are just a few things to get it over the finished line.

See inline comments. Let me know if anything is unclear!

sandbox/Settings.qml
```diff
@@ -0,0 +1,67 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.14
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+
+Column {
+    spacing: 12
+    width: 800
+    anchors.top:parent.top
```
borismelnik, line 10: thats not good. If it possible use `anchors, width and height` when you use this object. Here should be just `natural sizes` like `implicitWidth` :)

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,150 @@
+import QtQuick 2.14
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+
+Rectangle {
+    id: root
```
borismelnik, line 8: I know we talk about it but i remember than we agreed use filename id in StatusQ. But maybe im wrong :)

src/StatusQ/Components/StatusExpandableSettingsItem.qml
borismelnik: Agreed with Pascal. For example, now its impossible to use it with `Row`

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: aah this really makes sense. I didnt know about this. Will update.

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: in the design - https://www.figma.com/file/0m5JUlP8dv9szwQiGvP9gb/Wallet-Redesign-Sketchpad-%5BWIP%5D?node-id=357%3A33670 the value mentioned is -
background: #FFFFFF;

How do I get the right value for this? @PascalPrecht

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: thanks for noticing this!

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: old habits die hard @borismelnik :D will change it!

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: I just took that value from the design file. but i can make it 16

sandbox/Settings.qml
Khushboo-dev-cpp: Hey @borismelnik  this is only a demonstration of how to use the StatusExpandableSettingsItem component, Should be okay right?

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: image
 This definitely looks better! @PascalPrecht

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: no makes sense. I should limit it to the area before the button in case it is visible

Khushboo-dev-cpp:
Thanks for the amazing inputs @PascalPrecht  @borismelnik  Have fixed them all

sandbox/StatusExpandableSettingsItemPage.qml
```diff
@@ -0,0 +1,69 @@
...
+        icon.name: "seed-phrase"
+        primaryText: qsTr("Back up seed phrase")
+        secondaryText: qsTr("Back up your seed phrase now to secure this account ajhaDH SDHSAHDLSADBSA,DLISAHDLASD ADASDHASLDHALSDHAS DAS,DASJDGLIASGD")
+        button.text: qsTr("Back up seed phrase")
+    }
+
+    StatusExpandableSettingsItem {
+        anchors.horizontalCenter: parent.horizontalCenter
+
+        expandable: true
```
PascalPrecht, line 27: Here's another idea, curious what you think:

We could potentially get rid off this property as well, if we rely on the heuristic that the component is always expandable when there's a `expandableComponent`

Wdyt?

src/StatusQ/Components/StatusExpandableSettingsItem.qml
```diff
@@ -0,0 +1,155 @@
...
+            width: 32
+            height: 32
+            color: Theme.palette.primaryColor2
+        }
+    }
+
+    implicitWidth: 718
+
+    radius: 8
+    color: "transparent"//Theme.palette.indirectColor1
```
PascalPrecht, line 29: The comment we should remove.

I think going with "transparent" is fine for now. If we ever run into cases where this is problematic, we can create a new `Theme.palette.statusExpandableSettingsItem` property

PascalPrecht (commented):
This LGTM!

Just one more comment to remove and another idea to get your feedback on! :)

src/StatusQ/Components/StatusExpandableSettingsItem.qml
PascalPrecht: Just to follow up here:

Sometimes, the the corresponding color of the dark theme doesn't really match what's needed in a design. E.g. `Theme.palette.indirectColor1` is white in light mode and black in darkmode. This works fine when the color is used for things like fonts.

When either mode's color doesn't match, we usually introduce a specific component-based mapping in `Theme.palette`. You can see a bunch of them for example here https://github.com/status-im/StatusQ/blob/master/src/StatusQ/Core/Theme/StatusDarkTheme.qml#L139-L180

sandbox/StatusExpandableSettingsItemPage.qml
Khushboo-dev-cpp: As discussed, lets keep this explicit for the user of this component

src/StatusQ/Components/StatusExpandableSettingsItem.qml
Khushboo-dev-cpp: aah thanks.

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/f3ab4ce9c91ba1e654bf5460dffbbea454225be5

Great work @Khushboo-dev-cpp !

PR #371, opened 2021-09-01
 feat(StatusInput): Introduced secondaryLabel property

### LINKED ISSUE #383: Update StatusInput to support desktop #3310 & #3311 tasks
[#3310](https://app.zenhub.com/workspaces/nim-status-client--nim-status-5ea70d29c12ee928df5ae848/issues/status-im/status-desktop/3310)
[#3311](https://app.zenhub.com/workspaces/nim-status-client--nim-status-5ea70d29c12ee928df5ae848/issues/status-im/status-desktop/3311)

alexandraB99:
Due to design updates in AddAccount modal, updates are needed in StatusBaseInput and StatusInput

    * Added the possibility of having the icon on the right
      side
    * Added secondaryLabel for title
    * Added examples in StatusInputPage

Closes #383

### DIFF
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -45,23 +45,23 @@ Item {
     property bool valid: true
     property bool pristine: true
     property bool dirty: false
+    property bool leftIcon: true
 
     property StatusIconSettings icon: StatusIconSettings {
         width: 24
         height: 24
         name: ""
+        color: Theme.palette.baseColor1
     }
 
     implicitWidth: 448
-    implicitHeight: multiline ? Math.max(edit.implicitHeight + topPadding + bottomPadding, 44) : 44
+    implicitHeight: multiline ? Math.max((edit.implicitHeight + topPadding + bottomPadding), 44) : 44
 
     Rectangle {
         width: parent.width
         height: maximumHeight != 0 ? Math.min(
                                          minimumHeight != 0 ? Math.max(statusBaseInput.implicitHeight, minimumHeight)
-                                                            : implicitHeight,
-                                         maximumHeight)
-                                   : parent.height
+                                         : statusBaseInput.implicitHeight, maximumHeight) : parent.height
         color: Theme.palette.baseColor2
         radius: 8
 
@@ -80,36 +80,37 @@ Item {
 
         StatusIcon {
             id: statusIcon
-            anchors.verticalCenter: parent.verticalCenter
-            anchors.left: parent.left
-            anchors.leftMargin: 10
             anchors.topMargin: 10
-
+            anchors.left: statusBaseInput.leftIcon ? parent.left : undefined
+            anchors.right: !statusBaseInput.leftIcon ? parent.right : undefined
+            anchors.leftMargin: 10
+            anchors.rightMargin: 10
+            anchors.verticalCenter: parent.verticalCenter
             icon: statusBaseInput.icon.name
             width: statusBaseInput.icon.width
             height: statusBaseInput.icon.height
-            color: Theme.palette.baseColor1
+            color: statusBaseInput.icon.color
             visible: !!statusBaseInput.icon.name
         }
 
         Flickable {
             id: flick
-
             anchors.top: parent.top
             anchors.bottom: parent.bottom
-            anchors.left: statusIcon.visible ? statusIcon.right : parent.left
-            anchors.right: parent.right
-            anchors.leftMargin: statusIcon.visible ? 8 : statusBaseInput.leftPadding
-            anchors.rightMargin: statusBaseInput.rightPadding + clearable ? clearButton.width : 0
+            anchors.left: (statusIcon.visible && statusBaseInput.leftIcon) ?
+                          statusIcon.right : parent.left
+            anchors.right: (statusIcon.visible && !statusBaseInput.leftIcon) ?
+                            statusIcon.left : parent.right
+            anchors.leftMargin: statusIcon.visible && statusBaseInput.leftIcon ? 8
+                                : statusBaseInput.leftPadding
+            anchors.rightMargin: statusBaseInput.rightPadding + clearable ? clearButton.width
+                                 : statusIcon.visible && !leftIcon ? 8: 0
             anchors.topMargin: statusBaseInput.topPadding
             anchors.bottomMargin: statusBaseInput.bottomPadding
             contentWidth: edit.paintedWidth
             contentHeight: edit.paintedHeight
-            clip: true
             boundsBehavior: Flickable.StopAtBounds
-
             QC.ScrollBar.vertical: QC.ScrollBar { interactive: multiline; enabled: multiline }
-
             function ensureVisible(r) {
                 if (contentX >= r.x)
                     contentX = r.x;
@@ -120,25 +121,21 @@ Item {
                 else if (contentY+height <= r.y+r.height)
                     contentY = r.y+r.height-height;
             }
-
-
-
             TextEdit {
                 id: edit
                 property string previousText: text
-
                 width: flick.width
+                height: flick.height
+                verticalAlignment: Text.AlignVCenter
                 selectByMouse: true
                 selectionColor: Theme.palette.primaryColor2
                 selectedTextColor: color
-                anchors.verticalCenter: parent.verticalCenter
                 focus: true
                 font.pixelSize: 15
                 font.family: Theme.palette.baseFont.name
                 color: Theme.palette.directColor1
-                onCursorRectangleChanged: flick.ensureVisible(cursorRectangle)
+                onCursorRectangleChanged: { flick.ensureVisible(cursorRectangle); }
                 wrapMode: statusBaseInput.multiline ? Text.WrapAtWordBoundaryOrAnywhere : TextEdit.NoWrap
-
                 onActiveFocusChanged: {
                     if (statusBaseInput.pristine) {
                         statusBaseInput.pristine = false
@@ -178,15 +175,14 @@ Item {
                         previousText = text
                     }
                 }
-
                 StatusBaseText {
                     id: placeholder
-                    visible: edit.text.length === 0
+                    visible: (edit.text.length === 0)
                     anchors.left: parent.left
                     anchors.right: parent.right
+                    anchors.rightMargin: statusBaseInput.rightPadding
                     anchors.verticalCenter: parent.verticalCenter
                     font.pixelSize: 15
-
                     elide: StatusBaseText.ElideRight
                     font.family: Theme.palette.baseFont.name
                     color: statusBaseInput.enabled ? Theme.palette.baseColor1 :
--- a/sandbox/StatusInputPage.qml
+++ b/sandbox/StatusInputPage.qml
@@ -68,6 +68,22 @@ Column {
     }
 
     StatusInput {
+        label: "StatusInput"
+        secondaryLabel: "with right icon"
+        input.icon.width: 15
+        input.icon.height: 11
+        input.icon.name: text !== "" ? "checkmark" : ""
+        input.leftIcon: false
+    }
+
+    StatusInput {
+        label: "Label"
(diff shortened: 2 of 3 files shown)


src/assets/img/icons/plusSign.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M8 6V1.00685C8 0.449949 7.55228 0 7 0C6.44386 0 6 0.45078 6 1.00685V6H1.00685C0.449949 6 0 6.44772 0 7C0 7.55614 0.45078 8 1.00685 8H6V12.9932C6 13.5501 6.44772 14 7 14C7.55614 14 8 13.5492 8 12.9932V8H12.9932C13.5501 8 14 7.55228 14 7C14 6.44386 13.5492 6 12.9932 6H8Z" fill="white"/>
+</svg>
```
PascalPrecht, line 3: We already have an icon `add` which is that plus sign

src/assets/img/icons/enter_private_key.svg
```diff
@@ -0,0 +1,4 @@
...
+<path fill-rule="evenodd" clip-rule="evenodd" d="M5.2 2.80005C3.43269 2.80005 2 4.23274 2 6.00005V14.0001C2 15.7674 3.43269 17.2001 5.2 17.2001H6C7.76731 17.2001 9.2 15.7674 9.2 14.0001V6.00005C9.2 4.23274 7.76731 2.80005 6 2.80005H5.2ZM6 4.00005H5.2C4.09543 4.00005 3.2 4.89548 3.2 6.00005V11.0687C3.2 11.2469 3.41543 11.3361 3.54142 11.2101L7.97574 6.77579C7.99077 6.76075 8 6.74066 8 6.7194V6.00005C8 4.89548 7.10457 4.00005 6 4.00005ZM3.2 13.2807C3.2 13.2594 3.20923 13.2394 3.22426 13.2243L7.65858 8.79C7.78457 8.66401 8 8.75324 8 8.93142V14.0001C8 15.1046 7.10457 16.0001 6 16.0001H5.2C4.09543 16.0001 3.2 15.1046 3.2 14.0001V13.2807Z" fill="#4360DF"/>
+</svg>
```
PascalPrecht, line 4: All icons living in `img/icons` need to have a 24x24 container around it. Can you update this asset accordingly please?

PascalPrecht (changes requested):

src/assets/img/icons/plusSign.svg
borismelnik: And if you don't mind maybe name of icon should be `plus_sign.svg` like others

PascalPrecht:
@alexandraB99 sorry to only point this out now, but I've realised that other icon also already exists in StatusQ as `address`.

Here's how it goes:

If you're looking for an icon, check the name in https://www.figma.com/file/Gk13uZiVwGLHMPvQjaICAMWM/Status-icons?node-id=65%3A16 and see if there's an equivalent in src/assets/img/icons

Only if it doesn't exist we'll have to add it.

src/assets/img/icons/enter_private_key.svg
alexandraB99: done

alexandraB99:
> @alexandraB99 sorry to only point this out now, but I've realised that other icon also already exists in StatusQ as `address`.
> 
> Here's how it goes:
> 
> If you're looking for an icon, check the name in https://www.figma.com/file/Gk13uZiVwGLHMPvQjaICAMWM/Status-icons?node-id=65%3A16 and see if there's an equivalent in src/assets/img/icons
> 
> Only if it doesn't exist we'll have to add it.

done! thanks for letting me know

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -45,11 +47,13 @@ Item {
     property bool valid: true
     property bool pristine: true
     property bool dirty: false
+    property bool leftIcon: true
```
PascalPrecht, line 50: As mentioned offline, it's not clear to me why this is needed having an `icon.name` should indicate that there's an icon, no?

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -122,9 +128,7 @@ Item {
                     contentY = r.y+r.height-height;
             }
 
-
-
-            TextEdit {
+            QC.TextField {
```
PascalPrecht, line 131: This is needed for the echoMode?

PascalPrecht (changes requested):
Thanks for the PR!

As discussed offline, it'd be cool to get a variation of the new features being introduced here in sandbox. The `leftIcon` property isn't obvioous to me why it's necessary. Maybe you can clarify that?

re: commits: There's no `desktop/walletv2` in StatusQ. For new features/APIs introduced in components, please use the correct scope. You can have a look at existing commits for reference :)

Also, all existing status inputs in the sandbox app seem broken. This needs to be fixed

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -35,8 +37,8 @@ Item {
 
     property real leftPadding: 16
     property real rightPadding: 16
-    property real topPadding: 12
-    property real bottomPadding: 12
+    property real topPadding: 11
+    property real bottomPadding: 11
```
PascalPrecht, line 41: This still seems to position text inside a status input weirdly... (notice how it's not vertically centered anymore). The `bottomPadding` change may not be necessary
Screenshot 2021-09-06 at 09 01 05

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: I see, so this is to tell the component to not render the `icon` on the left side when a right side icon is needed.

There's indeed various other components that can potentially be rendered on the right side (e.g. there's a "Paste" button in some input components that need to be rendered in the same place).

I think ideally, we'd introduce a similar API to the one we have in `StatusListItem`, which is the `components` property (or just `component` in this case), which lets you render any kind of component in a particular slot:

```qml
StatusInput {
  ...
  component: StatusFlatRoundButton { ... } // Or StatusIcon or whatever else is needed
}
```
Having an API like that we:

1. Don't need to introduce a `leftIcon` property as this can be done based on heuristics
2. In fact, it's now possible to have both, the left icon **and** any component on the right hand side
3. Allow for other components than just icons (which we need anyways)

Do you feel like updating your PR to reflect such an API?
If it's too much of a hassle, we can do it afterwards as well, but I personally prefer to not introduce APIs that are known to be removed in the next step anyways.

PascalPrecht (changes requested):

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Created https://github.com/status-im/StatusQ/issues/380 accordingly

PascalPrecht: > no, it's to define whether the icon should go in the left or right side. Should I rename it to isLeftIcon or something like this? Would that be better?

Figured it out eventually :D : https://github.com/status-im/StatusQ/pull/371#discussion_r702643686

You can rename it if you want, but as mentioned in the comment above, we'll probably remove it very soon again anyways. I'm working on stuff that requires a comopnent on the right hand side as well, so might as well create a proper API for that.

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: done

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: Well, not sure if I should work on this on this occasion. I have already delayed my other task as you suggested to make use of StatusInput and the component was not supporting various requirements of it so I had to work on that as well. This is basically because when adding a new account we have to show a check mark when a seed phrase or private is entered correctly, so since it's about the same icon but different positions I thought to re-use the existing one instead of duplicating code. I am not aware of any other use cases of the StatusInput.

PascalPrecht: Okay, that's fair. Will do it in the next iteration then.

>  I am not aware of any other use cases of the StatusInput.

See https://github.com/status-im/StatusQ/issues/380 for more information

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Okay, for some reason this is still behaving weirdly. Notice how the textfield is "scrollable" even though it shouldn't.

Since `topPadding` and `bottomPadding` have been reverted to what they were before, I believe it has to do something with the introduction of `TextField`. Any idea what could cause this?

Another thing I've noticed: the text of the field itself has quite a big `leftMargin`. What's the reason this was introduced? 

Here's what inputs look like in `master`:
Screenshot 2021-09-06 at 11 16 39

Here's what they look like in this PR:

Screenshot 2021-09-06 at 11 16 57

src/StatusQ/Controls/StatusBaseInput.qml
borismelnik: Its should broke spellchecking....

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: Scrolling fixed, it seems that the `contentHeight` was not returning the real width and height of text so the flickable was already smaller that the TextField and thus it was flickering a bit, so TextMetrics is used now. Also the TextField sets the cursor x to 10, that's why a left margin of 10 was introduced, however I moved the placeholder out of the TextField and set a 6 left margin it (the difference between 16 and 10) and it should be fine now.

src/StatusQ/Controls/StatusBaseInput.qml
alexandraB99: @borismelnik in what sense?

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -178,23 +182,26 @@ Item {
...
-                    font.family: Theme.palette.baseFont.name
-                    color: statusBaseInput.enabled ? Theme.palette.baseColor1 :
-                                                     Theme.palette.directColor6
-                }
             }
 
         } // Flickable
+
+        StatusBaseText {
+            id: placeholder
```
borismelnik, line 190: if you used `TextField` you no need it anymore

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -80,36 +82,36 @@ Item {
...
             icon: statusBaseInput.icon.name
             width: statusBaseInput.icon.width
             height: statusBaseInput.icon.height
-            color: Theme.palette.baseColor1
+            color: statusBaseInput.icon.color
             visible: !!statusBaseInput.icon.name
         }
 
         Flickable {
             id: flick
```
borismelnik, line 99: no need `Flickable` anymore :)

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -80,36 +82,36 @@ Item {
 
         StatusIcon {
             id: statusIcon
```
borismelnik, line 84: it should be moved in `background` property of `TextField`

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -22,7 +22,9 @@ Item {
     property alias selectionEnd: edit.selectionEnd
     property alias cursorPosition: edit.cursorPosition
 
+    property alias edit: edit
```
borismelnik, line 25: Whole component can be based on `TextField`

PascalPrecht:
Nice, this has landed!

PR #379, opened 2021-09-03
feat(StatusModal): Add popup menu support for StatusModal

Khushboo-dev-cpp:
Added logic to support a popup menu to be launched from the StatusModal header.
Added an example in sandbox to demonstrate its usage.


statusq.qrc
```diff
@@ -264,8 +264,12 @@
         <file>src/assets/img/icons/windows_titlebar/maximize.svg</file>
         <file>src/assets/img/icons/windows_titlebar/minimise.svg</file>
         <file>src/assets/img/icons/windows_titlebar/status.svg</file>
+<<<<<<< Updated upstream
         <file>src/StatusQ/Controls/StatusSwitchTabButton.qml</file>
         <file>src/StatusQ/Controls/StatusSwitchTabBar.qml</file>
         <file>src/StatusQ/Components/StatusExpandableSettingsItem.qml</file>
+=======
+>>>>>>> Stashed changes
+        <file>src/assets/img/icons/snt.svg</file>
```
PascalPrecht, line 273: We have merge conflicts here

src/assets/img/icons/snt.svg
```diff
@@ -0,0 +1,6 @@
...
+<path d="M12.4512 14.354H68.5677V0.354004H12.4512V14.354ZM66.5677 12.354V68.3888H80.5677V12.354H66.5677ZM68.5677 66.3888H12.4512V80.3888H68.5677V66.3888ZM14.4512 68.3888V12.354H0.451172V68.3888H14.4512ZM12.4512 66.3888C13.5557 66.3888 14.4512 67.2842 14.4512 68.3888H0.451172C0.451172 75.0162 5.82376 80.3888 12.4512 80.3888V66.3888ZM66.5677 68.3888C66.5677 67.2842 67.4631 66.3888 68.5677 66.3888V80.3888C75.1951 80.3888 80.5677 75.0162 80.5677 68.3888H66.5677ZM68.5677 14.354C67.4631 14.354 66.5677 13.4586 66.5677 12.354H80.5677C80.5677 5.72658 75.1951 0.354004 68.5677 0.354004V14.354ZM12.4512 0.354004C5.82376 0.354004 0.451172 5.72658 0.451172 12.354H14.4512C14.4512 13.4586 13.5557 14.354 12.4512 14.354V0.354004Z" fill="white"/>
+</svg>
```
PascalPrecht, line 6: This probably doesn't belong here?

PascalPrecht (changes requested):
Great job!
This looks good to me except for a few minor things that probably went in here by accident.

src/assets/img/icons/snt.svg
Khushboo-dev-cpp: Hey this is an icon needed for the share modal. Is the location not right?

statusq.qrc
Khushboo-dev-cpp: oh no! I will change this
 asap!

src/assets/img/icons/snt.svg
PascalPrecht: How is this related to this PR?

Khushboo-dev-cpp: aaah you mean I should make 2 separate commits? this is not linked with popup support but with the PR on status desktop for Share Modal

Khushboo-dev-cpp: Will create 2 commits

PascalPrecht: Right :D

So this repo effectively doesn't know about a Share Modal, so having this SVG slide into the changes that introduce a popup Menu API confused me for a sec.

Khushboo-dev-cpp: You are right! Added another commit :)

src/StatusQ/Popups/statusModal/StatusModalHeader.qml
```diff
@@ -42,6 +49,15 @@ Rectangle {
...
     }
 
+    MouseArea {
+        anchors.fill: imageWithTitle
+        visible: !!statusModalHeader.popupMenu
+        cursorShape: Qt.PointingHandCursor
+        onClicked: {
+            popupMenuSlot.item.popup(imageWithTitle.x, imageWithTitle.y + imageWithTitle.height + 8)
+        }
+    }
```
PascalPrecht, line 59: I think we need to move this into `StatusImageWithTitle` and ensure that it doesn't take the entire with of the header (which the component at the moment does).

This results in the popup opening when it may not be expected.

Also, we should probably discuss with design if there should be a hover indicator (apart form the cursor) when the title is clickable

Khushboo-dev-cpp: Yes I actually agree with that. currently there is no way to know if the area is clickable except for the mouse cursor

Khushboo-dev-cpp: image


Check comments from carl. We can do this later

PascalPrecht: Gotcha!

Do we still wanna move the mouse area into that other component though and ensure it's not taking the entire width of the header?

Wdyt?

Khushboo-dev-cpp: I actually kept it that way so that user be easily access it as there is no indicator. Eventually it should be there though

PascalPrecht:
Landed - great work!

PR #384, opened 2021-09-06
feat(StatusInput) exposed edit component

alexandraB99:
We need to be able to call forceActiveFocus, so
TextEdit component should be somehow accessible

Relates to desktop #3310


PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/73c77c29c2db7918c6fd1654ecd8949c7ce65cbe
(Needed to fix the commit message a tiny bit)

PR #386, opened 2021-09-07
feat(Status.Core.Theme): add RobotoMono font

### LINKED ISSUE #342: Add RobotoMono-Regular font face
This is needed for <code> tag as per our design.

PascalPrecht:
Closes #342


src/StatusQ/Core/Theme/StatusDarkTheme.qml
```diff
@@ -68,6 +68,30 @@ ThemePalette {
         source: "../../../assets/fonts/InterStatus/InterStatus-Black.otf"
     }
 
+    property QtObject codeFont: FontLoader {
+        source: "../../../assets/fonts/RobotoMono/RobotoMono-Regular.ttf"
```
pinebit, line 72: I needed just Regular. Adding others would increase deployment size. I hope that's Okay.

pinebit (approved):

src/StatusQ/Core/Theme/StatusDarkTheme.qml
PascalPrecht: I'll just keep those and if we feel this is a bottle neck it terms of size we can still remove unused font sources.

PR #387, opened 2021-09-07
feat(StatusBaseInput): introduce `component` property

### LINKED ISSUE #380: Add `component` support to `StatusInput`
There are various cases where we need to render an input component that has some other component rendered inside of it on the right hand side:

There's the `clearable: true` option:
Screenshot 2021-09-06 at 09 36 37

Here's a new variation of validation indicators:
Screenshot 2021-09-06 at 09 35 29

Screenshot 2021-09-06 at 09 35 38

And lastly there can also be buttons:
Screenshot 2021-09-06 at 09 35 49

We have similar use cases in `StatusListItem` where even multiple components can be rendered on the right hand side of the item. There, we've introduced a `components` property to give users full control over what components they want to render inside the list item and what they can do.

Since we're no longer dealing with just icons (clearable or validity) but also buttons that need to be clickable (see paste button), it probably makes sense to introduce a similar API in StatusInput.

```qml
StatusInput {
  ...
  component: Item { ... }
}
```
Another reason this is necessary, is that currently, these components are rendered "on top of" the input, resulting in the value of the input to go "behind" because it doesn't know about the existance of that other component.

### To be discussed
How do we handle the `clearable: true` case?
I suggest that we keep `clearable` in order to not break anything and then have an either/or check. If it's `clearable`, `compoennt` will be ignored. Otherwise `component` is used.

This would also allow to "rebuild" the `clearable` case with `component` if necessary.

PascalPrecht:
This property enables users to load any component into the input field.
This is useful for rendering a "clearable" icon button, simple icons or
even more complex buttons.

Usage:

```qml
StatusBaseInput {
    ...
    component: StatusIcon {
        name: "cancel"
        color: Theme.palette.dangerColor1
        width: 16
    }
}
```

The `clearable` property of `StatusBaseInput` also renders and icon button
on the right hand side. With this new feature, `clearable` is just a short-hand
for:

```qml
StatusBaseInput {
    ...
    component: StatusFlatRoundButton {
        visible: edit.text.length != 0 &&
                statusBaseInput.clearable &&
                !statusBaseInput.multiline &&
                edit.activeFocus
        width: 24
        height: 24
        icon.name: "clear"
        icon.width: 16
        icon.height: 16
        icon.color: Theme.palette.baseColor1
        onClicked: {
            edit.clear()
        }
    }
}
```

Closes #380

### DIFF
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -56,6 +56,23 @@ Item {
         color: Theme.palette.baseColor1
     }
 
+    property Item component
+
+    onClearableChanged: {
+        if (clearable && !component) {
+            clearButtonLoader.active = true
+            clearButtonLoader.parent = statusBaseInputComponentSlot
+        } else {
+            clearButtonLoader.active = false
+        }
+    }
+
+    onComponentChanged: {
+        if (!!component) {
+            component.parent = statusBaseInputComponentSlot
+        }
+    }
+
     implicitWidth: 448
     implicitHeight: multiline ? Math.max((edit.implicitHeight + topPadding + bottomPadding), 44) : 44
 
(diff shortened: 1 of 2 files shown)


src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -80,149 +97,168 @@ Item {
...
-        height: 24
-        icon.name: "clear"
-        icon.width: 16
-        icon.height: 16
-        icon.color: Theme.palette.baseColor1
-        onClicked: {
-            edit.clear()
+    Loader {
+        id: clearButtonLoader
+        sourceComponent: StatusFlatRoundButton {
```
borismelnik, line 244: This button in the expected position? I ask bc i don't see anchors or layout

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -56,6 +56,23 @@ Item {
...
 
+    property Item component
+
+    onClearableChanged: {
+        if (clearable && !component) {
+            clearButtonLoader.active = true
+            clearButtonLoader.parent = statusBaseInputComponentSlot
+        } else {
+            clearButtonLoader.active = false
+        }
```
borismelnik, line 67: Soo, if user set up component and clearable - component have a high priority?

borismelnik: If yes its not clear.

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Yes, this is because it gets loaded into an `Item` that has the correct possible. Check `onClearableChanged` handler

src/StatusQ/Controls/StatusBaseInput.qml
PascalPrecht: Yes, if a `component` is given, `clearable` is ignored (because then there's a conflict, two components try to get rendered in the same spot)

PR #388, opened 2021-09-07
feat(StatusExpandableItem): Refactored the StatusExpandableSettingsIt…

Khushboo-dev-cpp:
…em to support different types

Renamed StatusExpandableSettingsItem to StatusExpandableItem.
Added support for dofferent types of styles for the item.
Type Primary: Relates to Settings Design
Type Secondary: Relates to Collectibles Design
Type Tertiary: Relates to the Collectibles detailed view design


PascalPrecht (approved):
Amazing. Nothing to comment!

Only noticed that the tertiary variant seems to not have super vertically centered text:
Screenshot 2021-09-07 at 16 22 04

Is that desired?

Bonus points: Can you update your commit message to include a "breaking change" section?
Have a look at https://www.conventionalcommits.org/en/v1.0.0/#examples

Khushboo-dev-cpp:
> Amazing. Nothing to comment!
> 
> Only noticed that the tertiary variant seems to not have super vertically centered text:
> Screenshot 2021-09-07 at 16 22 04
> 
> Is that desired?
> 
> Bonus points: Can you update your commit message to include a "breaking change" section?
> Have a look at https://www.conventionalcommits.org/en/v1.0.0/#examples

image

He if you notice the design I think the text should be on top, it is inconsistent in between "Properties" and "Date Group", however once the area the is expanded it can be in the center, likewise for the arrow. Thus kept it this way. 

About the the Breaking change. Will do it now

Khushboo-dev-cpp:
> > Amazing. Nothing to comment!
> > Only noticed that the tertiary variant seems to not have super vertically centered text:
> > Screenshot 2021-09-07 at 16 22 04
> > Is that desired?
> > Bonus points: Can you update your commit message to include a "breaking change" section?
> > Have a look at https://www.conventionalcommits.org/en/v1.0.0/#examples
> 
> image
> 
> He if you notice the design I think the text should be on top, it is inconsistent in between "Properties" and "Date Group", however once the area the is expanded it can be in the center, likewise for the arrow. Thus kept it this way.
> 
> About the the Breaking change. Will do it now

Btw the breaking change has been linked with this commit now as it still pending and not merged, so it shouldn't break anything :)

PascalPrecht:
> Btw the breaking change has been linked with this commit now as it still pending and not merged, so it shouldn't break anything :)

Yea that's true in Status Desktop (which, to be fair, is likely the only consumer of StatusQ atm), but that other version was already released in 0.8.0, so we should put at least a note about a breaking change in the next release.

Khushboo-dev-cpp:
> > Btw the breaking change has been linked with this commit now as it still pending and not merged, so it shouldn't break anything :)
> 
> Yea that's true in Status Desktop (which, to be fair, is likely the only consumer of StatusQ atm), but that other version was already released in 0.8.0, so we should put at least a note about a breaking change in the next release.

I agree with what you said. I just wanted to mention that luckily this won't break anything. Will keep in mind about this BREAKING CHANGE, I saw the last release notes and understand why it helps :) it's a really good practice

Khushboo-dev-cpp:
Seems like I might need to make some changes

PascalPrecht:
Merged!

Issue #307, opened 2021-08-03
Remove margin between chat messages and left pane


iurimatias:
original issue at https://github.com/status-im/status-desktop/issues/3123

![127674421-ba2cdc32-f3fe-4f1f-ac3e-61e0e9683194](https://user-images.githubusercontent.com/176720/127943033-22889c6d-c921-4371-9eea-23ee301bc17b.png)

PR #390, opened 2021-09-08
fix(StatusAppThreePanelLayout): Fix margin between left and center panels

borismelnik:
This margin shows because of `handle` width incuded in `SplitView` width. Handle have `Component` type and can't accessable for getting sizes. I add `magic constant` as hotfix.

Closes: #307


PascalPrecht:
Sweet!!

PR #394, opened 2021-09-08
feat: update validation for StatusInput and StatusValidator

emizzle:
`StatusInput` can now support multiple validators.

`StatusValidators` can now specify an error message directly in the validator component, which will appear when invalid.

`StatusInput` now has a validation mode that allows it to be validated when dirty, or always (regardless of dirty).

### DIFF
--- a/sandbox/StatusInputPage.qml
+++ b/sandbox/StatusInputPage.qml
@@ -89,14 +89,15 @@ Column {
         input.placeholderText: "Input with validator"
 
         validators: [
-            StatusMinLengthValidator { minLength: 10 }
-        ]
-
-        onTextChanged: {
-            if (errors && errors.minLength) {
-                errorMessage = `Value can't be shorter than ${errors.minLength.min} but got ${errors.minLength.actual}`
+            StatusMinLengthValidator {
+                minLength: 10
+                error: {
+                    if (errors && errors.minLength) {
+                        errorMessage = `Value can't be shorter than ${errors.minLength.min} but got ${errors.minLength.actual}`
+                    }
+                }
             }
-        }
+        ]
     }
 
     StatusInput {
--- a/src/StatusQ/Controls/StatusInput.qml
+++ b/src/StatusQ/Controls/StatusInput.qml
@@ -30,22 +30,43 @@ Item {
     property int charLimit: 0
     property string errorMessage: ""
     property list<StatusValidator> validators
+    property bool dirty: false
+    property int validationMode: StatusInput.ValidationMode.OnlyWhenDirty
+    enum ValidationMode {
+        OnlyWhenDirty, // validates input only after it has become dirty
+        Always // validates input even before it has become dirty
+    }
 
     property var errors: ({})
 
     function validate() {
+        if (!dirty && validationMode === StatusInput.ValidationMode.OnlyWhenDirty) {
+            return
+        }
+        statusBaseInput.valid = true
         if (validators.length) {
             for (let idx in validators) {
-                let result = validators[idx].validate(statusBaseInput.text)
+                let validator = validators[idx]
+                let result = validator.validate(statusBaseInput.text)
 
                 if (typeof result === "boolean" && result) {
-                    statusBaseInput.valid = true
+                    statusBaseInput.valid = statusBaseInput.valid && true
+                    delete errors[validator.name]
                 } else {
                     if (!errors) {
                         errors = {}
                     }
-                    errors[validators[idx].name] = result
-                    statusBaseInput.valid = false
+                    result.error = validator.error
+                    errors[validator.name] = result
+                    statusBaseInput.valid = statusBaseInput.valid && false
+                }
+            }
+            if (errors){
+                let errs = Object.values(errors)
+                if (errs && errs[0]) {
+                    errorMessage.text = errs[0].error || root.errorMessage;
+                } else {
+                    errorMessage.text = ""
                 }
             }
         }
@@ -107,7 +128,10 @@ Item {
         anchors.leftMargin: 16
         anchors.rightMargin: 16
         maximumLength: root.charLimit
-        onTextChanged: root.validate()
+        onTextChanged: {
+            root.dirty = true
+            root.validate()
+        }
 
         Keys.forwardTo: [root]
     }
@@ -123,11 +147,11 @@ Item {
         anchors.leftMargin: 16
 
         height: visible ? implicitHeight : 0
-        visible: !!root.errorMessage && !statusBaseInput.valid
+        visible: !!text && !statusBaseInput.valid
 
         font.pixelSize: 12
         color: Theme.palette.dangerColor1
-        text: root.errorMessage
+
 
         horizontalAlignment: Text.AlignRight
         wrapMode: Text.WordWrap
--- a/src/StatusQ/Controls/StatusInput.qml
+++ b/src/StatusQ/Controls/StatusInput.qml
@@ -30,7 +30,6 @@ Item {
     property int charLimit: 0
     property string errorMessage: ""
     property list<StatusValidator> validators
-    property bool dirty: false
     property int validationMode: StatusInput.ValidationMode.OnlyWhenDirty
     enum ValidationMode {
         OnlyWhenDirty, // validates input only after it has become dirty
@@ -40,7 +39,7 @@ Item {
     property var errors: ({})
 
     function validate() {
-        if (!dirty && validationMode === StatusInput.ValidationMode.OnlyWhenDirty) {
+        if (!statusBaseInput.dirty && validationMode === StatusInput.ValidationMode.OnlyWhenDirty) {
             return
         }
         statusBaseInput.valid = true
@@ -128,10 +127,7 @@ Item {
         anchors.leftMargin: 16
         anchors.rightMargin: 16
         maximumLength: root.charLimit
-        onTextChanged: {
-            root.dirty = true
-            root.validate()
-        }
+        onTextChanged: root.validate()
 
         Keys.forwardTo: [root]
     }
--- a/src/StatusQ/Controls/Validators/StatusAddressValidator.qml
+++ b/src/StatusQ/Controls/Validators/StatusAddressValidator.qml
@@ -1,0 +1,15 @@
+import StatusQ.Controls 0.1
+import StatusQ.Core.Utils 0.1
+
+StatusValidator {
+    name: "address"
+
+    error: qsTr("Please enter a valid address")
+
+    validate: function (t) {
+        // TODO: Add ENS name resolution and validation a la
+        // `ui/shared/AddressInput.qml`
+        return Utils.isValidAddress(t) ? true : { actual: t }
+    }
+}
+
--- a/src/StatusQ/Controls/Validators/StatusAddressValidator.qml
+++ b/src/StatusQ/Controls/Validators/StatusAddressValidator.qml
@@ -4,7 +4,7 @@ import StatusQ.Core.Utils 0.1
 StatusValidator {
     name: "address"
 
-    error: qsTr("Please enter a valid address")
+    errorMessage: "Please enter a valid address"
 
     validate: function (t) {
         // TODO: Add ENS name resolution and validation a la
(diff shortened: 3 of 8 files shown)


sandbox/StatusInputPage.qml
```diff
@@ -89,14 +89,15 @@ Column {
...
-        onTextChanged: {
-            if (errors && errors.minLength) {
-                errorMessage = `Value can't be shorter than ${errors.minLength.min} but got ${errors.minLength.actual}`
+            StatusMinLengthValidator {
+                minLength: 10
+                error: {
+                    if (errors && errors.minLength) {
+                        errorMessage = `Value can't be shorter than ${errors.minLength.min} but got ${errors.minLength.actual}`
+                    }
+                }
```
PascalPrecht, line 98: This is interesting.
As mentioned in: https://github.com/status-im/StatusQ/commit/ba4f27f9ba6e957908d743c571ebc15f2399cd71, error messages have been pushed out of the validator pipeline, so applications can decide how they want to render them (especially needed when there's custom i18n solutions).

But it seems you've provided an API so that validators **do** provide a default error message, but applications still get a way to override it (as seen here).

I like that.

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -30,22 +30,43 @@ Item {
     property int charLimit: 0
     property string errorMessage: ""
     property list<StatusValidator> validators
+    property bool dirty: false
+    property int validationMode: StatusInput.ValidationMode.OnlyWhenDirty
+    enum ValidationMode {
+        OnlyWhenDirty, // validates input only after it has become dirty
+        Always // validates input even before it has become dirty
+    }
```
PascalPrecht, line 38: I wonder if `OnlyWhenDirty` could lead to undesired behaviour but I'm probably missing something..

At the moment, validation happens `onCompleted` which is needed because, if a field is initialized with a value and its valid, the input will still be `valid: false` until `onTextChanged` was emitted.

Using `OnlyWhenDirty` we re-introduce this ill-behaviour, no?

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -30,22 +30,43 @@ Item {
...
+                    delete errors[validator.name]
                 } else {
                     if (!errors) {
                         errors = {}
                     }
-                    errors[validators[idx].name] = result
-                    statusBaseInput.valid = false
+                    result.error = validator.error
+                    errors[validator.name] = result
+                    statusBaseInput.valid = statusBaseInput.valid && false
```
PascalPrecht, line 61: Oh, nice catch!!

src/StatusQ/Controls/Validators/StatusAddressValidator.qml
```diff
@@ -0,0 +1,15 @@
+import StatusQ.Controls 0.1
+import StatusQ.Core.Utils 0.1
+
+StatusValidator {
+    name: "address"
+
+    error: qsTr("Please enter a valid address")
```
PascalPrecht, line 7: So far we've avoided `qsTr` inside StatusQ because Status Desktop will replace `qsTr` with `qsTrId`.

We probably don't want Status Desktop to change StatusQ's source. So giving it plain text would be better here. Applications can then still i18n the message in whatever way they prefer (`qsTr` or other solutions)

src/StatusQ/Controls/Validators/StatusAddressValidator.qml
```diff
@@ -0,0 +1,15 @@
...
+
+    validate: function (t) {
+        // TODO: Add ENS name resolution and validation a la
+        // `ui/shared/AddressInput.qml`
```
PascalPrecht, line 11: You can remove this comment.

ENS name resolution requires `asyncValidators` which would be a different pipeline (triggered **after** synchronous pipeline. But this doesn't exist yet.

src/StatusQ/Controls/Validators/StatusMinLengthValidator.qml
```diff
@@ -3,9 +3,16 @@ import StatusQ.Controls 0.1
 StatusValidator {
 
     property int minLength: 0
+    property string fieldName: "value"
```
PascalPrecht, line 6: Is this necessary? Wouldn't be a general message using "value" be enough?

PascalPrecht (changes requested):
Awesome work!

Thank for catching those bugs in the validation.
I've left some comments that may need to be addressed.
Also would you mind splitting this commit up semantically?

There's fixes and features in here, I'd love to see those show up in our changelog.

If it's too much of a hassle I'm happy to do it for you as well!

src/StatusQ/Controls/Validators/StatusAddressValidator.qml
PascalPrecht: One more thought here:

How would you feel about changing this to `errorMessage` instead?

src/StatusQ/Controls/Validators/StatusAddressValidator.qml
PascalPrecht: Created an issue here: https://github.com/status-im/StatusQ/issues/395

I could imagine we'll end up with a separate validator (`StatusENSValidator`?) which will be a `StatusAsyncValidator` and has to be applied in addition to `StatusAddressValidator` to get the same effect as what we have in legacy `AddressInput` at the moment.

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -30,22 +30,43 @@ Item {
     property int charLimit: 0
     property string errorMessage: ""
     property list<StatusValidator> validators
+    property bool dirty: false
```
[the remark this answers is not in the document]
emizzle: Removed.

src/StatusQ/Controls/StatusInput.qml
emizzle: Note: this semantically the same as `statusBaseInput.valid = false`, but figured this was more descriptive.

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -107,7 +128,10 @@ Item {
         anchors.leftMargin: 16
         anchors.rightMargin: 16
         maximumLength: root.charLimit
-        onTextChanged: root.validate()
+        onTextChanged: {
+            root.dirty = true
```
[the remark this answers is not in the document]
emizzle: Removed.

src/StatusQ/Controls/Validators/StatusAddressValidator.qml
emizzle: Unrelated to asynchronous validators, introducing a separate validator for ENS means the input would have to be a valid address *AND* a valid ENS address (whether or not the resolution to an address happens). But what we really want is validation for `is it an address` *OR* `is it a valid ENS name`. Maybe it would be good to keep both the static (synchronous) checks for address OR ENS in the same validator, but change the name to `StatusAddressOrEnsValidator`.

As mentioned in #395, if `StatusAddressOrEnsValidator` validates, then the asynchronous resolution and validation can happen.

What do you think?

src/StatusQ/Controls/Validators/StatusMinLengthValidator.qml
emizzle: It was a way to retain existing functionality in desktop where error messages specify the name of the field that has an invalid value. Due to the API that allows customisable error messages (via `error`, now `errorMessage`), I will remove this and update on the desktop side where needed.

src/StatusQ/Controls/Validators/StatusAddressValidator.qml
emizzle: Updated.

src/StatusQ/Controls/Validators/StatusAddressValidator.qml
emizzle: I've gone ahead and added a `StatusAddressValidator` and a `StatusAddressOrEnsValidator` which are synchronous, and allow room for async validators discussed in #395.

src/StatusQ/Controls/StatusInput.qml
emizzle: Let's think of the case where we want to validate a field with a min length of 1. If the input loads (ie `Component.onCompleted` is fired) with a blank input, validation will run, and the validator will see that the input which has text of `""`, and it'll say "oh, this isn't valid". However, the input should remain with `valid` = false From a UX perspective, we really don't want that input validated yet as the user hasn't had the chance to add their input. However, if the input was part of an edit form, it would contain pre-populated values, and in that case, it would be useful to the user to see a validation error message.

So we are presented with two cases:
1. a new clean form with inputs that haven't been edited in which we don't want validation to run `onCompleted`
2. a form with pre-populated values that we do want validation to run `onCompleted`

This allows consumers of `StatusInput` to dictate when the validation will occur, which is what I have done in https://github.com/status-im/status-desktop/pull/3425/files#diff-374f8aee4fa4150e3039b4cbfccc3114cfdb704c03045ec0b60a00cb0bbf2e3fR34-R36:
```qml
property int validationMode: edit ?
        StatusInput.ValidationMode.Always :
        StatusInput.ValidationMode.OnlyWhenDirty
```

PascalPrecht: > However, the input should remain with valid = false From a UX perspective, we really don't want that input validated yet as the user hasn't had the chance to add their input.

I see, so this is really just to save computation then? Because all inputs are `valid: false` by default. Having them validate with an empty value cause them to stay `valid: false` (which is the desired behaviour). If I understand correctly, all we're doing with this flag is save the computation because we already "know" the field is invalid (because it hasn't been touched by the user yet).

If the above is true, then there's still a possibility that users won't use the "correct" `mode` (e.g. simply not setting `OnlyWhenDirty` would still work "as expected" [`valid: false` remains]).

Should probably be guided via documentation then.

Anyways, thanks for clarifying. I think we can keep this then.

emizzle: It seems that the base input is `valid: true` by default:
https://github.com/status-im/StatusQ/blob/48309d3040dcaa5a6ee955f8a95153b18316edbc/src/StatusQ/Controls/StatusBaseInput.qml#L47

PascalPrecht: Oh dang. You're right. Wonder if we should change this..

sandbox/StatusInputPage.qml
```diff
@@ -89,14 +89,15 @@ Column {
...
         validators: [
-            StatusMinLengthValidator { minLength: 10 }
-        ]
-
-        onTextChanged: {
-            if (errors && errors.minLength) {
-                errorMessage = `Value can't be shorter than ${errors.minLength.min} but got ${errors.minLength.actual}`
+            StatusMinLengthValidator {
+                minLength: 10
+                error: {
```
PascalPrecht, line 94: Need to update this to `errorMessage` as well

sandbox/StatusInputPage.qml
```diff
@@ -89,14 +89,15 @@ Column {
...
+                    if (errors && errors.minLength) {
```
PascalPrecht, line 95: `errors` is a property on `StatusInput`, so this expression will fail

src/StatusQ/Controls/Validators/StatusMinLengthValidator.qml
```diff
@@ -6,6 +6,12 @@ StatusValidator {
 
     name: "minLength"
 
+    errorMessage: {
+        minLength === 1 ?
+            "Please enter a value" :
+            `Value is ${errors.address.actual} characters, but must be at least ${minLength}.`
```
PascalPrecht, line 12: Just realised that this won't work as we don't have access to `errors` here. And it makes sense, because only the input knows about all of its validators and which ones have failed. Probably the easiest to get "around" it is to just remove the `actual` value from the default message.

Will think about it.

PascalPrecht:
@emizzle As discussed offline, I've split up your commits. Feel free to review here: https://github.com/status-im/StatusQ/pull/402

PascalPrecht:
Closing this in favour of #402

PR #408, opened 2021-09-14
fix(components) Right anchored title in StatusChatInfoToolBar

alexandraB99:
So that elide mode in text can be activated

Relates to https://github.com/status-im/status-desktop/issues/3242

### DIFF
--- a/src/StatusQ/Components/StatusChatInfoToolBar.qml
+++ b/src/StatusQ/Components/StatusChatInfoToolBar.qml
@@ -28,6 +28,8 @@ Item {
         anchors.verticalCenter: parent.verticalCenter
         anchors.left: parent.left
         anchors.leftMargin: 5
+        anchors.right: (implicitWidth > parent.width - 50) ? statusMenuButton.left : undefined
+        anchors.rightMargin: 5
         type: StatusChatInfoButton.Type.OneToOneChat
         onClicked: statusChatInfoToolBar.chatInfoButtonClicked()
     }
@@ -39,7 +41,7 @@ Item {
         anchors.verticalCenter: parent.verticalCenter
 
         visible: popupMenuSlot.active
-        width: 32
+        width: visible ? 32 : 0
         height: 32
 
         type: StatusRoundButton.Type.Secondary


alaibe (approved):

src/StatusQ/Components/StatusChatInfoToolBar.qml
```diff
@@ -28,6 +28,8 @@ Item {
         anchors.verticalCenter: parent.verticalCenter
         anchors.left: parent.left
         anchors.leftMargin: 5
+        anchors.right: statusMenuButton.left
+        anchors.rightMargin: 5
```
PascalPrecht, line 32: There's a typo here so the margin is broken.

In addition:

1. This was "fixed" in https://github.com/status-im/StatusQ/commit/7ef61ed3f897d8b06c1185377de46dc173077edb but it looks like it requires the `width` of `StatusInfoButton` to be calculated. Here's how it was done in `StatusChatToolBar` which uses the same component: https://github.com/status-im/StatusQ/commit/7ef61ed3f897d8b06c1185377de46dc173077edb#diff-561ecc8f8bd7aa15fd3928afcad98931b5ceb20b570739485458be78dce8dd19R40
2. We should probably double check with designers if it's okay to have the button take the entire width here (which is what happens due to `anchors.right`). As per design, it should more behave like it was done in `StatusChatToolBar`.
2. If we go with this solution, we need to make sure this also works when no `popupMenu` is supplied. Because right now the button still bleeds out of the left panel
Screenshot 2021-09-14 at 14 31 11

PascalPrecht (changes requested):
Requesting changes due to typo and clarification with design.

src/StatusQ/Components/StatusChatInfoToolBar.qml
alexandraB99: @PascalPrecht not sure if I get it, this is what I see on my end:

With popup menu:
withMenu

Without popup menu:
withoutMenu

PascalPrecht: Ah that's weird. I'm seeing the same as you do now. Probably forgot to reload the front-end.
This then looks good to me!

Just need to double check with design (@simonricoo are we good with having that button to take more space than needed? Designs usually show a width of button content + a tiny padding)

PascalPrecht: @simonricoo specifically referring to this:

Screenshot 2021-09-14 at 18 00 10

PascalPrecht: > There's a typo here so the margin is broken.

Also this isn't true, misread the code. Sorry for that!

PascalPrecht (approved):

simonricoo:
hey @alexandraB99 & @PascalPrecht we want that entire container to adopt on the length of the content so we want to see this as a result of the hover state :) 
Screen Shot 2021-09-15 at 09 34 23
https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=5197%3A51614

alexandraB99:
> hey @alexandraB99 & @PascalPrecht we want that entire container to adopt on the length of the content so we want to see this as a result of the hover state :)
> Screen Shot 2021-09-15 at 09 34 23
> https://www.figma.com/file/Mr3rqxxgKJ2zMQ06UAKiWL/%F0%9F%92%AC-Chat%E2%8E%9CDesktop?node-id=5197%3A51614

Ok updated. This is what we got:

short
normal
long

PascalPrecht:
NICE! 🎉

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/6e8a36be865aa327068f54f9de3a385a86571134 
FYI: fixed the commit message

PR #410, opened 2021-09-14
feat(StatusInput): introduce `leftPadding` and `rightPadding` properties

PascalPrecht:


alexandraB99 (approved):
We have in desktop in Theme.qml the following:

```
    property int xlPadding: 32
    property int bigPadding: 24
    property int padding: 16
    property int halfPadding: 8
    property int smallPadding: 10
```

Is this something we should use instead?

PascalPrecht:
@alexandraB99 Was thinking about that a lot too... This was done a lot in Status Desktop as well, but then my conclusion was,  that there's no real benefit to put these values in variables and reuse them. Here's why:

When various places/components use things like `Style.current.padding`, the only win you get is that you can change the value in a single place for all of the usages. This is nice in theory but in practice I don't think it makes a lot of sense here. Especially because the variable is used in different semantics as well. Not just margins and paddings. If you now decide to change the value of that variable, it probably breaks the UI in many places.

Then I wondered, how much is that really better than just specifying those values inline and also saving some keystrokes?

However, if there's consensus in the team that we're better off using such variables, then I'm happy to go down than path with you together! :D

alexandraB99:
> @alexandraB99 Was thinking about that a lot too... This was done a lot in Status Desktop as well, but then my conclusion was, that there's no real benefit to put these values in variables and reuse them. Here's why:
> 
> When various places/components use things like `Style.current.padding`, the only win you get is that you can change the value in a single place for all of the usages. This is nice in theory but in practice I don't think it makes a lot of sense here. Especially because the variable is used in different semantics as well. Not just margins and paddings. If you now decide to change the value of that variable, it probably breaks the UI in many places.
> 
> Then I wondered, how much is that really better than just specifying those values inline and also saving some keystrokes?
> 
> However, if there's consensus in the team that we're better off using such variables, then I'm happy to go down that path with you together! :D

@PascalPrecht At the end this is a design matter, if design plans to maintain the same values all over the application, then it makes sense, if not then probably not.

PascalPrecht:
Yes I agree. And I think I'd be fine with moving to variables for those margins etc, we just need to make sure that they are used correctly (e.g. if we end up using them for, say vertical margin between components, then, if we want to change the inner margin for some components, we could end up changing the vertical margins as well which might not be desired).

If you want, you can create an issue for that so we can address this soon.

PR #411, opened 2021-09-14
feat(StatusInput): add support for asynchronous validators

### LINKED ISSUE #395: Add support for `asyncValidators` in `StatusInput`
Something that needs to be explored. There are a bunch of use cases for async validators. In Status Desktop specifically, there's a case where an input lets users enter ENS names. Those names need to successfully resolve. If they don't they should be considered invalid.

We need to look into how this can be done in a nice way, but from a usage point of view, this should look similar to using synchronous validators:

```qml
StatusInput {
  ...
  validators: [...]
  asyncValidators: [...]
}
```

We probably need to leverage `Connections` here. Need to play around with this to see how this can work.
We definitely need to introduce a new `pending` state.

One requirement that is known for sure: Asynchronous validators are only executed when synchronous validation was successful.

PascalPrecht:
This adds support for asynchronous validators which are also just `StatusValidator`.

There are a few differences to syncronous validators though:

1. `validate` function doesn't return, it's only used to trigger the async
   validation. Ideally it would return an async object like Promise or Observable
   but since we have to rely on QML `Connections`, this won't work.
2. As mentioned, `Connections` are needed to listen to async events so the
  validation can be resolved.
3. Validation has to be resolved by calling `input.updateValidity` inside the validator.
   That API notifies `StatusInput` that validation is has been performed and expects
   the name of the validator as well as either a boolean (true), if validation was successful,
   or, if validation has failed, a boolean (false) or an error object.

Here's a `StatusENSValidator` as an example:

```qml
StatusValidator {

    id: root

    name: "ensValidator"
    errorMessage: "Couldn't resolve ENS name."

    readonly property string uuid: Utils.uuid()

    property int debounceTime: 600

    signal ensResolved(string address)

    validate: Backpressure.debounce(root, root.debounceTime, function (name) {
        name = name.startsWith("@") ? name.substring(1) : name
        walletModel.ensView.resolveENS(name, uuid)
    })

    Connections {
        target: walletModel.ensView
        onEnsWasResolved: {
            if (uuid !== root.uuid) {
                return
            }
            root.ensResolved(resolvedAddress)
            input.updateValidityAndPendingState(root.name, resolvedAddress !== "")
        }
    }
}
```

Closes #395

### DIFF
--- a/src/StatusQ/Controls/Validators/StatusValidator.qml
+++ b/src/StatusQ/Controls/Validators/StatusValidator.qml
@@ -1,10 +1,12 @@
 import QtQuick 2.13
+import StatusQ.Controls 0.1
 
 QtObject {
     id: statusValidator
 
     property string name: ""
     property string errorMessage: "invalid input"
+    property StatusInput input
 
     property var validate: function (value) {
         return true
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -47,6 +47,7 @@ Item {
     property bool valid: true
     property bool pristine: true
     property bool dirty: false
+    property bool pending: false
     property bool leftIcon: true
 
     property StatusIconSettings icon: StatusIconSettings {
--- a/src/StatusQ/Controls/StatusInput.qml
+++ b/src/StatusQ/Controls/StatusInput.qml
@@ -29,6 +29,8 @@ Item {
     property string secondaryLabel: ""
     property int charLimit: 0
     property string errorMessage: ""
+    property real leftPadding: 16
+    property real rightPadding: 16
     property list<StatusValidator> validators
     property int validationMode: StatusInput.ValidationMode.OnlyWhenDirty
     enum ValidationMode {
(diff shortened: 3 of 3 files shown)


PascalPrecht:
/CC'ing @emizzle as he had thoughts on how else we could implement this.

PascalPrecht:
Here's a branch where it can be tested: https://github.com/status-im/status-desktop/pull/3377

src/StatusQ/Controls/Validators/StatusValidator.qml
```diff
@@ -1,10 +1,12 @@
 import QtQuick 2.13
+import StatusQ.Controls 0.1
 
 QtObject {
     id: statusValidator
 
     property string name: ""
     property string errorMessage: "invalid input"
+    property StatusInput input
```
emizzle, line 9: This prevents the entire app from running 😱  I wonder if it's some kind of circular dependency, but nothing was reported in the output. Took a while to figure that one out.

PascalPrecht: Oh damn! I've actually noticed that too. Made this change last minute without testing it 🙈 
Then realised that Status wasn't rendering a window any more.

Let's change this back to `var` then

emizzle:
@PascalPrecht should we close this in favour of https://github.com/status-im/StatusQ/pull/416?

PR #412, opened 2021-09-14
fix(StatusListItem) Disable sensor when chevron clicked

alexandraB99:
List items' mouse area was also triggered when chevron (sub menu) button was clicked so it was unsetting the opened variable which is responsible for opening/closing the chats list.

Relates to https://github.com/status-im/status-desktop/issues/3484

### DIFF
--- a/src/StatusQ/Components/StatusChatListCategory.qml
+++ b/src/StatusQ/Components/StatusChatListCategory.qml
@@ -42,14 +42,15 @@ Column {
 
         highlighted: statusChatListCategory.dragged
         sensor.onClicked: {
-            if (mouse.button === Qt.RightButton && showActionButtons && !!statusChatListCategory.popupMenu) {
-                highlighted = true
-                popupMenuSlot.item.popup(mouse.x + 4, mouse.y + 6)
-                return
+            if (sensor.enabled) {
+                if (mouse.button === Qt.RightButton && showActionButtons && !!statusChatListCategory.popupMenu) {
+                    highlighted = true;
+                    popupMenuSlot.item.popup(mouse.x + 4, mouse.y + 6);
+                    return
+                }
+                statusChatListCategory.opened = !opened;
             }
-            statusChatListCategory.opened = !opened
         }
-
         onTitleClicked: statusChatListCategory.opened = !opened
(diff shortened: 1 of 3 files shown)


PascalPrecht (changes requested):
@alexandraB99 Marking as "Request Changes" to ensure we address issues below:

Would love to merge this but it turns out that this change is problematic, so we need to think about solutions for these things first:

1. First of all, having the `MouseArea` not wrapping the entire content of the component results in the cursor leaving the mouse area when hovering over "transcluded" components. You can see that in the "List Items" page in the sandbox application. If there's no easy way to fix it, then users of the component have to explicitly set the `hovered` state of the list item when the cursor hovers a `component`.
2. This causes a weird flicker in `StatusCategoryListItem` when hovering over any of the action buttons. Probably related to 1).

Do you see any solutions to these?

alexandraB99:
> @alexandraB99 Marking as "Request Changes" to ensure we address issues below:
> 
> Would love to merge this but it turns out that this change is problematic, so we need to think about solutions for these things first:
> 
> 1. First of all, having the `MouseArea` not wrapping the entire content of the component results in the cursor leaving the mouse area when hovering over "transcluded" components. You can see that in the "List Items" page in the sandbox application. If there's no easy way to fix it, then users of the component have to explicitly set the `hovered` state of the list item when the cursor hovers a `component`.
> 2. This causes a weird flicker in `StatusCategoryListItem` when hovering over any of the action buttons. Probably related to 1).
> 
> Do you see any solutions to these?

Thanks @PascalPrecht for pointing this our. I have updated with an alternative. The problem was that the sensor's mouse area was also triggered when the chevron button was clicked, so while the chevron was setting `opened` to true the sensor was unsetting to false :) Since in the chat list case the sensor and the chevron should both open the chats list, the sensor is temporary disabled when the chevron is clicked to prevent this situation. The changes in StatusListItem, are just indentation that I did when reseting the file to what it was.

PascalPrecht:
Nice, this is much better!
Thanks for coming up with a different solution!

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/c83641d2f017cdd6623bfeaa3f2a5c32c33a4118

PR #414, opened 2021-09-15
feat(StatusSpellcheckingMenuItems): Add spellchecking menu

borismelnik:
Closes: #398

### DIFF
--- a/src/StatusQ/Popups/StatusSpellcheckingMenuItems.qml
+++ b/src/StatusQ/Popups/StatusSpellcheckingMenuItems.qml
@@ -1,0 +1,84 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+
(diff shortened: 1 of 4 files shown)


src/StatusQ/Popups/StatusSpellcheckingMenuItems.qml
```diff
@@ -0,0 +1,84 @@
...
+
+        background: Rectangle {
+            color: ignoreWord.hovered ? Theme.palette.statusPopupMenu.hoverBackgroundColor
+                           : "transparent"
+        }
+
+        onTriggered: {
+            root.addToUserDict()
+        }
+    }
```
PascalPrecht, line 54: Any reason we're not using `StatusMenuItem` here?

borismelnik: StatusMenuItem is an Action, not MenuItem and i dont know why but i have wrong order if use Action with Repeater in menu.

PR #419, opened 2021-09-22
fix: track category opened state

emizzle:
Category `opened` state is tracked and restored on model change.

Previously, when the model data for categories and channels was changed, all unexpanded, or collapsed categories would automatically expand. This was due to the default state of the category `opened` property being restored (which was set to true) when the model data changed.

With this change in place, the opened state of each category is tracked in the parent component (`StatusChatListAndCategories`), and on each model change, the opened state is restored.

NOTE: it was tempting to the put the changes inside of the `StatusChatListCategory` component, however this would not work as the component is used as a delegate, and all state is wiped on each model change.

### DIFF
--- a/src/StatusQ/Components/StatusChatListAndCategories.qml
+++ b/src/StatusQ/Components/StatusChatListAndCategories.qml
@@ -28,6 +28,9 @@ Item {
     property alias sensor: sensor
     property bool draggableItems: false
     property bool draggableCategories: false
+    // Keeps track of expanded category state. Should only be modified
+    // internally at runtime.
+    property var openedCategoryState: new Object({})
 
     property Component categoryPopupMenu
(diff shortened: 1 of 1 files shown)


src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -28,6 +28,9 @@ Item {
     property alias sensor: sensor
     property bool draggableItems: false
     property bool draggableCategories: false
+    // Keeps track of expanded category state. Should only be modified
+    // internally at runtime.
+    property var openedCategoryState: new Object({})
```
PascalPrecht, line 33: We can probably mark this as `internal`

PascalPrecht (approved):
LGTM!

Nice changes @emizzle !

PR #420, opened 2021-09-22
fix(StatusSpellcheckingMenuItems): Exact  menu items order

borismelnik:
Sometimes menu items is lost order.

### DIFF
--- a/src/StatusQ/Popups/StatusSpellcheckingMenuItems.qml
+++ b/src/StatusQ/Popups/StatusSpellcheckingMenuItems.qml
@@ -13,71 +13,94 @@ StatusPopupMenu {
     signal addToUserDict()
     signal disableSpellchecking()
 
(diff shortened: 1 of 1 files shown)


src/StatusQ/Popups/StatusSpellcheckingMenuItems.qml
```diff
@@ -13,71 +13,94 @@ StatusPopupMenu {
...
+                }
+
+                background: Rectangle {
+                    color: variants.hovered ? Theme.palette.statusPopupMenu.hoverBackgroundColor
+                                              : "transparent"
+                }
+
+                onTriggered: {
+                    root.menuItemClicked(index)
+                }
```
PascalPrecht, line 45: Do we have sandbox demo variations of these so we can test them?

PR #422, opened 2021-09-23
feat: add star icons

emizzle:


src/assets/img/icons/star-icon-outline.svg
```diff
@@ -0,0 +1,3 @@
...
+<path fill-rule="evenodd" clip-rule="evenodd" d="M8.85145 4.58361L7.3077 0.871991C7.19383 0.598227 6.80602 0.598227 6.69215 0.871991L5.1484 4.58361L1.1414 4.90485C0.845844 4.92854 0.726003 5.29738 0.951182 5.49027L4.0041 8.10542L3.07138 12.0156C3.00259 12.304 3.31634 12.5319 3.56937 12.3774L6.99993 10.282L10.4305 12.3774C10.6835 12.5319 10.9973 12.304 10.9285 12.0156L9.99575 8.10542L13.0487 5.49027C13.2738 5.29738 13.154 4.92854 12.8585 4.90485L8.85145 4.58361ZM10.8765 6.03425C10.9891 5.93781 10.9292 5.75339 10.7814 5.74154L8.77154 5.58041C8.39775 5.55045 8.07214 5.31388 7.92813 4.96764L7.15381 3.10596C7.09688 2.96908 6.90297 2.96908 6.84604 3.10596L6.07172 4.96764C5.92771 5.31388 5.6021 5.55045 5.22831 5.58041L3.21848 5.74154C3.0707 5.75339 3.01078 5.93781 3.12337 6.03425L4.65465 7.34596C4.93944 7.58991 5.06381 7.97269 4.97681 8.33744L4.50897 10.2987C4.47458 10.4429 4.63145 10.5569 4.75797 10.4796L6.47867 9.42861C6.79869 9.23315 7.20116 9.23315 7.52118 9.42861L9.24188 10.4796C9.3684 10.5569 9.52527 10.4429 9.49088 10.2987L9.02304 8.33744C8.93604 7.97269 9.06041 7.58991 9.3452 7.34596L10.8765 6.03425Z" fill="#FFCA0F"/>
+</svg>
```
PascalPrecht, line 3: If this is what I think it is, we already have the icon named as `favourite`. You can find all supported icons in this Figma: https://www.figma.com/file/Gk13uZiVwGLHMPvQjaICAMWM/Status-icons?node-id=1078%3A0

The names of the objects in the figma file map to the names of the icons inside StatusQ.

Also imporant: every icon has a size of 24x24 (basically having a square box around them), this is needed so they all take the same space.

If your icon doesn't exist there, we need to check with design. But I think `favourite` is the one you're looking for.

src/assets/img/icons/star-icon.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="13" viewBox="0 0 14 13" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M8.85169 4.58361L7.30794 0.871991C7.19408 0.598227 6.80626 0.598227 6.6924 0.871991L5.14865 4.58361L1.14164 4.90485C0.846088 4.92854 0.726247 5.29738 0.951426 5.49027L4.00434 8.10542L3.07163 12.0156C3.00283 12.304 3.31658 12.5319 3.56961 12.3774L7.00017 10.282L10.4307 12.3774C10.6838 12.5319 10.9975 12.304 10.9287 12.0156L9.996 8.10542L13.0489 5.49027C13.2741 5.29738 13.1543 4.92854 12.8587 4.90485L8.85169 4.58361Z" fill="#FFCA0F"/>
+</svg>
```
PascalPrecht, line 3: Based on the name of the file, I assume this is the same start icon, but filled?

PascalPrecht:
@emizzle can you double check my comments here when you get a chance?

I don't think we needed to add both of these SVG sources.

src/assets/img/icons/star-icon.svg
emizzle: That's correct.

src/assets/img/icons/star-icon-outline.svg
emizzle: Ok will swap it out for `favourite`, but are you saying the icons have padding including *in* the svg? Or as part of the component?

I've added padding around the icon in the desktop PR such that there is no background highlight yet there is a larger click surface.

PascalPrecht: That's padding **in** the SVG. We do that by exporting the wrapping layer as well. This is what it looks like for all icons living in statusQ:
Screenshot 2021-09-24 at 12 49 56

The `StatusIcon` component doesn't add any padding on top of that. So you get full control via `width` and `height` proeprties.

PR #424, opened 2021-09-23
WIP: implement cross category drag and drop

0x-r4bbit:
I think UI wise, this seems to be pretty much done. 

To allow drag and drop into and out of categories we need to switch to a different approach:

1. No more use of `DelegateModel`. The reason `DelegateModel` was used in the first place, was so that the chat list can be easily **visually** updated as the user is dragging an item over the list (using its `.move()` API). Because dragged items could origin from a different category (/chat list), there's no benefit to using `DelegateModel` anymore
2. Removing `DelegateModel` also eliminates the warnings reported in status-im/status-desktop#7572 which was significantly slowing down the rendering performance.
3. Instead of having items slide up and down, we now just add a "drop indicator" between items. This makes the whole implementation easier and more predictable.
4. There's another `DropArea` added to `StatusChatList` which is activated when a chat list is empty. This is needed so users can drag chat items into empty categories or even just out of categories into "uncategorized" lists
5. Dragged elements now keep track of not only their new position but also their original position (previous `orignalOrder` was just assigned the new position. We now have `newOrder` for that. Should probably rename both to `originalPosition` and `newPosition` respectively`).
6. `chatItemReordered` signal no longer just emits the same `categoryId` in which the dragged item resides in, but rather its `newCategoryId`, in case it was dragged into a different category. Status-go is smart enough to recognize that and moves the chat item into the new category accordingly

**What's missing**:

Proper visual state update **after** the drop has happened but **before** the signal was sent to the backend.

### DIFF
--- a/src/StatusQ/Components/StatusChatList.qml
+++ b/src/StatusQ/Components/StatusChatList.qml
@@ -18,7 +18,7 @@ Column {
 
     property string categoryId: ""
     property string selectedChatId: ""
-    property alias chatListItems: delegateModel
+    property alias chatListItems: statusChatListItems
     property bool draggableItems: false
 
     property alias statusChatListItems: statusChatListItems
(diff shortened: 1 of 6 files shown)


src/StatusQ/Components/StatusChatList.qml
```diff
@@ -11,12 +11,14 @@ import StatusQ.Controls 0.1
 Column {
     id: statusChatList
 
-    spacing: 4
+    spacing: 0
     width: 288
 
+    property string uuid: Utils.uuid()
```
PascalPrecht, line 17: Can get rid off this one as it's no longer used

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -29,199 +31,262 @@ Column {
 
     signal chatItemSelected(string id)
     signal chatItemUnmuted(string id)
-    signal chatItemReordered(string id, int from, int to)
+    signal chatItemReordered(string categoryId, string id, int from, int to)
```
PascalPrecht, line 34: This is a breaking change and needs to be noted.

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -29,199 +31,262 @@ Column {
...
             popupMenuSlot.sourceComponent = popupMenu
         }
     }
 
-    DelegateModel {
-        id: delegateModel
+    Column {
+        id: statusChatListItemsWrapper
+        width: parent.width
+        spacing: 4
```
PascalPrecht, line 45: This column is needed to get a sense of how many items live in this chatlist.
At the moment, we always pass every chat item to every chat list and hide them if they don't belong to the corresponding category. This means, the model has never a `count: 0` so we have to rely on the `implicitHeight` of the `Repeater` that renders them.

We can't use the root `Column` for that because we have other elements living inside of it.

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -29,199 +31,262 @@ Column {
...
-                            let originalOpenHandler = popupMenuSlot.item.openHandler
-                            let originalCloseHandler = popupMenuSlot.item.closeHandler
-
-                            popupMenuSlot.item.openHandler = function () {
-                                if (!!originalOpenHandler) {
-                                    originalOpenHandler((model.chatId || model.id))
+                        originalOrder: model.position
+                        chatId: model.chatId || model.id
+                        categoryId: model.categoryId || ""
+                        /* name: (!!statusChatList.chatNameFn ? statusChatList.chatNameFn(model) : model.name) + " POSITION: " + statusChatListItem.newOrder */
```
PascalPrecht, line 133: This needs to be removed

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -29,199 +31,262 @@ Column {
...
+    Loader {
+        id: popupMenuSlot
+        active: !!statusChatList.popupMenu
+    }
 
-                onEntered: reorderDelay.start()
-                onDropped: statusChatList.chatItemReordered(statusChatListItem.chatId, drag.source.originalOrder, statusChatListItem.DelegateModel.itemsIndex)
+    Loader {
+        id: emptyListDropAreaLoader
+        active: statusChatListItemsWrapper.implicitHeight == 0
```
PascalPrecht, line 258: This doesn't seem to be enough. When the wrapper gets an `implicitHeight` at runtime, the loader is still active. Hence, we're updating `active` explicityly in `onImplicitHeightChanged`

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -29,199 +31,262 @@ Column {
...
+                        }
+                        onUnmute: statusChatList.chatItemUnmuted(model.chatId || model.id)
 
-                            popupMenuSlot.item.popup(mouse.x + 4, statusChatListItem.y + mouse.y + 6)
-                            popupMenuSlot.item.openHandler = originalOpenHandler
-                            return
+                        StatusChatListDropIndicator {
+                            anchors.bottom: parent.bottom
+                            anchors.bottomMargin: -2
+                            visible: dropArea.containsDrag
```
PascalPrecht, line 184: Misses `opacity: 0.5`

src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -142,7 +142,14 @@ Item {
                         chatList.selectedChatId: statusChatListAndCategories.selectedChatId
                         chatList.onChatItemSelected: statusChatListAndCategories.chatItemSelected(id)
                         chatList.onChatItemUnmuted: statusChatListAndCategories.chatItemUnmuted(id)
-                        chatList.onChatItemReordered: statusChatListAndCategories.chatItemReordered(model.categoryId, id, from, to)
+                        chatList.onChatItemReordered: {
+                            console.log("REORDERING CHAT ITEM:")
+                            console.log("FROM CATEGORY: ", model.categoryId)
+                            console.log("TO: ", categoryId)
+                            console.log("FROM POSITION: ", from)
+                            console.log("TO POSITION: ", to)
```
PascalPrecht, line 150: Obviously these need to be removed. Leaving them for now for debugging purpose. Will go away before merge

src/StatusQ/Components/StatusChatListCategory.qml
```diff
@@ -8,6 +9,7 @@ Column {
 
     spacing: 0
     opacity: dragged ? 0.5 : 1
+    property string uuid: Utils.uuid()
```
PascalPrecht, line 12: Can be removed as it's no longer used. Left over from building/testing

PR #440, opened 2021-10-13
feat(StatusModal): Add edit avatar button

borismelnik:
Part of https://github.com/status-im/status-desktop/issues/3734

### DIFF
--- a/src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
+++ b/src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
@@ -75,6 +75,8 @@ Row {
 
             Rectangle {
                 id: editAvatarIcon
+
+                objectName: "editAvatarImage"
                 anchors.right: parent.right
                 anchors.bottom: parent.bottom
                 anchors.bottomMargin: -3
--- a/src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
+++ b/src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
@@ -12,8 +12,10 @@ Row {
     property int titleElide: Text.ElideRight
     property int subTitleElide: Text.ElideRight
     property bool editable: false
+    property bool headerImageEditable: false
 
     signal editButtonClicked
+    signal headerImageClicked
 
     property StatusImageSettings image: StatusImageSettings {
         width: 40
@@ -39,7 +41,7 @@ Row {
             }
             return statusRoundedImageCmp
         }
-        active: statusImageWithTitle.icon.isLetterIdenticon || 
+        active: statusImageWithTitle.icon.isLetterIdenticon ||
                 !!statusImageWithTitle.image.source.toString()
     }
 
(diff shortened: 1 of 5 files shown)


src/StatusQ/Popups/statusModal/StatusModalHeader.qml
```diff
@@ -36,6 +38,18 @@ Rectangle {
...
+        anchors.bottom: parent.bottom
+        width: parent.width
+        height: parent.radius
+        color: parent.color
+
+        StatusModalDivider {
+            anchors.bottom: parent.bottom
+            width: parent.width
+        }
+    }
```
PascalPrecht, line 51: Any reason this was moved?

src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
```diff
@@ -92,11 +117,12 @@ Row {
...
                 elide: statusImageWithTitle.titleElide
                 color: Theme.palette.directColor1
-                width: !editButton.visible ? parent.width : 
-                      parent.width - editButton.width - parent.spacing
+                width: !editButton.visible ? parent.width :
+                                             parent.width - editButton.width - parent.spacing
             }
             StatusFlatRoundButton {
                 id: editButton
+                objectName: "editAvatarbButton"
```
PascalPrecht, line 125: Is this needed?

PascalPrecht: If yes, what is it for?

src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
```diff
@@ -64,12 +66,35 @@ Row {
...
+
+            StatusRoundButton {
+                type: StatusRoundButton.Type.Secondary
+
+                anchors.right: parent.right
+                anchors.bottom: parent.bottom
+                anchors.bottomMargin: -3
+
+                width: 15
+                height: 15
```
PascalPrecht, line 84: I believe this has to be bigger though. Looking at the figma, the button is 16x16 and also, you can see in the UI that it's super small:
Screenshot 2021-10-13 at 14 03 52

PascalPrecht: Keep in mind that the `border.width` wont affect the overall size

PascalPrecht: Also, it looks like it isn't positioned correctly. It's hanging "off" a but too much

src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
```diff
@@ -64,12 +66,35 @@ Row {
...
+
+                border.color: Theme.palette.indirectColor1
```
PascalPrecht, line 86: I don't think we can do that and have to do no border here. It's hard to provide a border color with alpha channel that works across all backgrounds (avatar images):
Screenshot 2021-10-13 at 14 04 00

So we either have to do a hard check against `Theme.palette.name` and give it a solid border color based on theme, or we remove it entirely

src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
```diff
@@ -64,12 +66,35 @@ Row {
...
+                border.width: 1
+
+                icon.name: "tiny/edit"
+                icon.width: 8
+                icon.height: 8
```
PascalPrecht, line 91: Visually, this seems to be too tiny. It's possible that there's a padding white space in the icon source, so wie might need to bump up the width and height here a little.

src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
```diff
@@ -12,8 +12,10 @@ Row {
     property int titleElide: Text.ElideRight
     property int subTitleElide: Text.ElideRight
     property bool editable: false
+    property bool editAvatar: false
```
PascalPrecht, line 15: Another nitpick: how about naming this: `headerImageEditable` instead across all boards?

borismelnik: no problem :)

src/StatusQ/Popups/statusModal/StatusImageWithTitle.qml
borismelnik: okay

src/StatusQ/Popups/statusModal/StatusModalHeader.qml
borismelnik: bc of `z` order this line hide part of header.

PR #460, opened 2021-10-20
feat(StatusSmartIdenticon): Created a new StatusQ component

Khushboo-dev-cpp:
feat(StatusSmartIdenticon): Created a new StatusQ component to accommodate -

1. A StatusRoundedImage
2. StatusRoundIcon
3. LetterIdenticon

Fallback in case of an error in loading Image to the LetterIdenticon


src/StatusQ/Components/StatusSmartIdenticon.qml
```diff
@@ -0,0 +1,74 @@
+import QtQuick 2.13
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
```
PascalPrecht, line 4: I think this needs `StatusQ.Components 0.1` import as well

PascalPrecht (commented):
Super clean PR @Khushboo-dev-cpp !
This looks good to me. Only thing I've noticed, it seems this breaks some list item variations. See screenshot below:

Screenshot 2021-10-21 at 12 23 53


The two items with super small icons seem too small.

Also can you add a page for `StatusSmartIdenticon` to sandbox?

src/StatusQ/Components/StatusSmartIdenticon.qml
PascalPrecht: > Sure I will add a page for it. Its being used at so many places though and thought it would be okay not too

Actually that's fair enough

> Its works without it or
what do you mean?

Surprisingly it seems to work

PR #477, opened 2021-11-02
fix(StatusSlider) made slider flexible

alexandraB99:
Made slider background and handle to not depend
to root's height so that it covers cases where a
legend is needed thus all should be clickable for
better user experience.

Relates to status-im/status-desktop#3984


PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/31cfc8833ab3a7ca231826033e98764a7a160ee9

PR #483, opened 2021-11-04
fix(StatusChatList) group chat right click

alexandraB99:
As per design when right clicking on a chat list item it
should get selected and activated. Ref pictures on git issue

Relates to https://github.com/status-im/status-desktop/issues/3975


PascalPrecht:
Hey @alexandraB99 I believe we don't want to merge this. See: https://github.com/status-im/status-desktop/issues/3975#issuecomment-962924230

There's probably a misunderstanding.

PR #486, opened 2021-11-15
refactor(StatusChatList): updates due to chat & communities models refactored

alexandraB99:
### What does the PR do
Updated Models.qml for chats and communities models to reflect
changes due to refactor in the actual backend. Updated as well the relevant UI
components as properties like `model.chatId` is now `model.itemId` as well as in order to retrieve and display categories from the communities model.

### Affected areas
`StatusChatList` and `StatusChatListAndCategories`

Relates to https://github.com/status-im/status-desktop/issues/4062

### DIFF
--- a/src/StatusQ/Components/StatusChatList.qml
+++ b/src/StatusQ/Components/StatusChatList.qml
@@ -15,8 +15,7 @@ Column {
     width: 288
 
     property string categoryId: ""
-    property string selectedChatId: ""
-    property alias chatListItems: delegateModel
+    property var model: []
     property bool draggableItems: false
 
     property alias statusChatListItems: statusChatListItems
@@ -24,8 +23,6 @@ Column {
     property Component popupMenu
 
     property var filterFn
-    property var profileImageFn
-    property var chatNameFn
 
     signal chatItemSelected(string id)
     signal chatItemUnmuted(string id)
@@ -39,17 +36,16 @@ Column {
 
     DelegateModel {
         id: delegateModel
-
+        model: statusChatList.model
         delegate: Item {
             id: draggable
             width: statusChatListItem.width
             height: statusChatListItem.height
-
             property alias chatListItem: statusChatListItem
 
             visible: {
                 if (!!statusChatList.filterFn) {
-                    return statusChatList.filterFn(model, statusChatList.categoryId)
+                    return statusChatList.filterFn(model)
                 }
                 return true
             }
@@ -96,31 +92,19 @@ Column {
                 StatusChatListItem {
 
                     id: statusChatListItem
-
-                    property string profileImage: ""
-
                     opacity: dragSensor.active ? 0.0 : 1.0
-                    Component.onCompleted: {
-                        if (typeof statusChatList.profileImageFn === "function") {
-                            profileImage = statusChatList.profileImageFn(model.chatId || model.id) || ""
-                        }
-                    }
                     originalOrder: model.position
-                    chatId: model.chatId || model.id
-                    categoryId: model.categoryId || ""
-                    name: !!statusChatList.chatNameFn ? statusChatList.chatNameFn(model) : model.name
-                    type: model.chatType
-                    muted: !!model.muted
-                    hasUnreadMessages: !!model.hasUnreadMessages || model.unviewedMessagesCount > 0
-                    hasMention: model.mentionsCount > 0
-                    badge.value: model.chatType === StatusChatListItem.Type.OneToOneChat ?
-                        model.unviewedMessagesCount || 0 :
-                        model.mentionsCount || 0
-                    selected: (model.chatId || model.id) === statusChatList.selectedChatId
-
-                    icon.color: model.color || ""
-                    image.isIdenticon: !!!profileImage && !!!model.identityImage && !!model.identicon
-                    image.source: profileImage || model.identityImage || model.identicon || ""
+                    chatId: model.itemId
+                    categoryId: model.parentItemId
+                    name: model.name
+                    type: !!model.type ? model.type : StatusChatListItem.Type.CommunityChat
+                    muted: model.muted
+                    hasUnreadMessages: model.hasUnreadMessages
+                    notificationsCount: model.notificationsCount
+                    selected: model.active
+                    icon.color: model.color
+                    image.isIdenticon: model.isIdenticon
+                    image.source: model.icon
 
                     sensor.cursorShape: dragSensor.cursorShape
                     onClicked: {
@@ -132,7 +116,7 @@ Column {
 
                             popupMenuSlot.item.openHandler = function () {
                                 if (!!originalOpenHandler) {
-                                    originalOpenHandler((model.chatId || model.id))
+                                    originalOpenHandler(model.itemId)
                                 }
                             }
 
@@ -152,10 +136,10 @@ Column {
                             return
                         }
                         if (!statusChatListItem.selected) {
-                            statusChatList.chatItemSelected(model.chatId || model.id)
+                            statusChatList.chatItemSelected(model.itemId)
                         }
                     }
-                    onUnmute: statusChatList.chatItemUnmuted(model.chatId || model.id)
+                    onUnmute: statusChatList.chatItemUnmuted(model.itemId)
                 }
             }
 
@@ -204,9 +188,8 @@ Column {
                     type: draggable.chatListItem.type
                     muted: draggable.chatListItem.muted
                     dragged: true
-                    hasUnreadMessages: draggable.chatListItem.hasUnreadMessages
-                    hasMention: draggable.chatListItem.hasMention
-                    badge.value: draggable.chatListItem.badge.value
+                    hasUnreadMessages: model.hasUnreadMessages
+                    notificationsCount: model.notificationsCount
                     selected: draggable.chatListItem.selected
 
                     icon.color: draggable.chatListItem.icon.color
--- a/src/StatusQ/Components/StatusChatListAndCategories.qml
+++ b/src/StatusQ/Components/StatusChatListAndCategories.qml
@@ -20,11 +20,9 @@ Item {
         text: "More"
     }
 
-    property string selectedChatId: ""
+    property var model: []
     property bool showCategoryActionButtons: false
     property bool showPopupMenu: true
-    property alias chatList: statusChatList.chatListItems
-    property alias categoryList: delegateModel
     property alias sensor: sensor
     property bool draggableItems: false
     property bool draggableCategories: false
@@ -71,26 +69,25 @@ Item {
             StatusChatList {
                 id: statusChatList
                 anchors.horizontalCenter: parent.horizontalCenter
-                visible: chatListItems.count > 0
-                selectedChatId: statusChatListAndCategories.selectedChatId
+                visible: statusChatList.model.count > 0
                 onChatItemSelected: statusChatListAndCategories.chatItemSelected(id)
                 onChatItemUnmuted: statusChatListAndCategories.chatItemUnmuted(id)
                 onChatItemReordered: statusChatListAndCategories.chatItemReordered(categoryId, id, from, to)
                 draggableItems: statusChatListAndCategories.draggableItems
+                model: statusChatListAndCategories.model
                 filterFn: function (model) {
-                    return !!!model.categoryId
+                    return (model.subItems.count === 0)
                 }
                 popupMenu: statusChatListAndCategories.chatListPopupMenu
             }
 
             DelegateModel {
                 id: delegateModel
-
+                model: statusChatListAndCategories.model
                 delegate: Item {
                     id: draggable
                     width: statusChatListCategory.width
                     height: statusChatListCategory.height
-
                     property alias chatListCategory: statusChatListCategory
 
                     StatusChatListCategory {
@@ -99,7 +96,7 @@ Item {
                         property bool dragActive: false
                         property real startY: 0
                         property real startX: 0
-                      
+
                         opacity: dragActive ? 0.0 : 1.0
 
                         dragSensor.drag.target: draggedListCategoryLoader.item
@@ -136,16 +133,15 @@ Item {
                         menuButton.tooltip: statusChatListAndCategories.categoryMenuButtonToolTip
 
                         originalOrder: model.position
-                        categoryId: model.categoryId
+                        categoryId: model.itemId
                         name: model.name
                         showActionButtons: statusChatListAndCategories.showCategoryActionButtons
-                        addButton.onClicked: statusChatListAndCategories.categoryAddButtonClicked(model.categoryId)
+                        addButton.onClicked: statusChatListAndCategories.categoryAddButtonClicked(model.itemId)
 
-                        chatList.chatListItems.model: statusChatListAndCategories.chatList.model
-                        chatList.selectedChatId: statusChatListAndCategories.selectedChatId
+                        chatList.model: model.subItems
                         chatList.onChatItemSelected: statusChatListAndCategories.chatItemSelected(id)
                         chatList.onChatItemUnmuted: statusChatListAndCategories.chatItemUnmuted(id)
-                        chatList.onChatItemReordered: statusChatListAndCategories.chatItemReordered(model.categoryId, id, from, to)
+                        chatList.onChatItemReordered: statusChatListAndCategories.chatItemReordered(model.itemId, id, from, to)
                         chatList.draggableItems: statusChatListAndCategories.draggableItems
 
                         popupMenu: statusChatListAndCategories.categoryPopupMenu
@@ -154,7 +150,7 @@ Item {
                         // Used to set the initial value of "opened" when the
                         // model is bound/changed.
                         opened: {
-                            let openedState = statusChatListAndCategories.openedCategoryState[model.categoryId]
+                            let openedState = statusChatListAndCategories.openedCategoryState[model.itemId]
                             return openedState !== undefined ? openedState : true // defaults to open
                         }
 
@@ -163,7 +159,7 @@ Item {
                         // as the state would be lost each time the model is
                         // changed.
                         onOpenedChanged: {
-                            statusChatListAndCategories.openedCategoryState[model.categoryId] = statusChatListCategory.opened
+                            statusChatListAndCategories.openedCategoryState[model.itemId] = statusChatListCategory.opened
                         }
                     }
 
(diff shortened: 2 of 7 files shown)


src/StatusQ/Components/StatusChatList.qml
```diff
@@ -96,31 +94,27 @@ Column {
...
-                    hasUnreadMessages: !!model.hasUnreadMessages || model.unviewedMessagesCount > 0
-                    hasMention: model.mentionsCount > 0
-                    badge.value: model.chatType === StatusChatListItem.Type.OneToOneChat ?
-                        model.unviewedMessagesCount || 0 :
-                        model.mentionsCount || 0
-                    selected: (model.chatId || model.id) === statusChatList.selectedChatId
+                    hasUnreadMessages: !!model.hasNotification || model.unviewedMessagesCount > 0
+                    hasMention: model.notificationsCount > 0
+                    badge.value: model.notificationsCount
+                    selected: (model.itemId || model.id) === statusChatList.selectedChatId
```
saledjenic, line 113: @alexandraB99 there is `active` field maintained from the backend so this should be just `selected: model.active`

src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -23,7 +23,6 @@ Item {
     property string selectedChatId: ""
```
saledjenic, line 23: @alexandraB99 feel free to remove `selectedChatId` from both `StatusChatListAndCategories` and `StatusChatList` components.

saledjenic (changes requested):
@alexandraB99 changes look good to me, I added few comments like improvements if we want to be completely correct.

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -96,31 +94,27 @@ Column {
...
 
                     icon.color: model.color || ""
                     image.isIdenticon: !!!profileImage && !!!model.identityImage && !!model.identicon
-                    image.source: profileImage || model.identityImage || model.identicon || ""
+                    image.source: profileImage || model.icon || model.identicon || ""
```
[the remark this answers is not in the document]
alexandraB99: ok. how do we know if it's identicon?

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -96,31 +94,27 @@ Column {
                 StatusChatListItem {
 
                     id: statusChatListItem
-
                     property string profileImage: ""
-
                     opacity: dragSensor.active ? 0.0 : 1.0
                     Component.onCompleted: {
                         if (typeof statusChatList.profileImageFn === "function") {
```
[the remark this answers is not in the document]
alexandraB99: done

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -96,31 +94,27 @@ Column {
...
                     muted: !!model.muted
-                    hasUnreadMessages: !!model.hasUnreadMessages || model.unviewedMessagesCount > 0
-                    hasMention: model.mentionsCount > 0
-                    badge.value: model.chatType === StatusChatListItem.Type.OneToOneChat ?
-                        model.unviewedMessagesCount || 0 :
-                        model.mentionsCount || 0
-                    selected: (model.chatId || model.id) === statusChatList.selectedChatId
+                    hasUnreadMessages: !!model.hasNotification || model.unviewedMessagesCount > 0
+                    hasMention: model.notificationsCount > 0
+                    badge.value: model.notificationsCount
```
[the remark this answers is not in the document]
alexandraB99: how do we know whether hasUnreadMessages or hasMention?

src/StatusQ/Components/StatusChatList.qml
alexandraB99: done

src/StatusQ/Components/StatusChatListAndCategories.qml
alexandraB99: done

src/StatusQ/Components/StatusChatList.qml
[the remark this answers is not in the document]
saledjenic: @alexandraB99 qml actually doesn't need to know what is it, whether it's an image or identicon, whether is a profile image or identity image or whatever, qml just need to display what backend says to display (with minimal component related logic), so you just need to set `icon` as a source and don't worry, if displayed image is not the correct one then we know it's up to backend and will fix it there.

See [here](https://github.com/status-im/StatusQ/blob/93563fc22854d7375bf2290e2f73c96707c33921/sandbox/Models.qml#L10) what we have for a model, `name`, `icon`, `color`, the same we have in our real backend model, so your component logic should be:

- if `model.icon` is not an empty string display icon
- if `model.icon` is an empty string display icon letter icon, using the first letter from `model.name` and color defined by `model.color`

src/StatusQ/Components/StatusChatList.qml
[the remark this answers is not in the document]
saledjenic: @alexandraB99 as display says, 
- if `model.hasNotification` is **false** then **don't display** neither `dot badge` nor `count badge` **regardless to the `model. notificationsCount`** (but `model.notificationsCount` in such cases will be surely 0)
- if `model.hasNotification` is **true and `model.notificationsCount` is 0**, that means we have some notification but not direct to the logged in user (so far that means user has some unread messages in that channel) so you should **display `dot badge`**
- if `model.hasNotification` is **true and `model.notificationsCount` is greater 0** then **display `count badge`**

so as you may see `model.hasNotification` informs qml that we should display either dot or count badge, but `model.countNotification` determines if it will be dot badge (`model.countNotification` == 0) or count badge (`model.countNotification` > 0), *respecting the advantage that count badge has a higher priority*.

saledjenic:
@alexandraB99 tried, but I cannot add you as a reviewer here, most likely cause you opened this PR. :)
All in all this is ready to be reviewed and merged into `base_bc`.

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -96,31 +92,19 @@ Column {
...
-                    badge.value: model.chatType === StatusChatListItem.Type.OneToOneChat ?
-                        model.unviewedMessagesCount || 0 :
-                        model.mentionsCount || 0
-                    selected: (model.chatId || model.id) === statusChatList.selectedChatId
-
-                    icon.color: model.color || ""
-                    image.isIdenticon: !!!profileImage && !!!model.identityImage && !!model.identicon
-                    image.source: profileImage || model.identityImage || model.identicon || ""
+                    chatId: model.itemId
+                    categoryId: model.itemId
```
jrainville, line 98: I think this one is wrong. Probably `categoryId`?

src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -136,16 +133,15 @@ Item {
                         menuButton.tooltip: statusChatListAndCategories.categoryMenuButtonToolTip
 
                         originalOrder: model.position
-                        categoryId: model.categoryId
+                        categoryId: model.itemId
```
jrainville, line 136: I see you use `itemId` here as well. Maybe then something else needs to be changed so it's not as confusing?

jrainville (commented):
Looks good, I just wonder if there is a mistake with the `categoryId`

src/StatusQ/Components/StatusChatList.qml
saledjenic: @jrainville even it looks like it's wrong, actually it's not :) The reason is cause components used on the qml side doesn't follow model's struct from the backend side. And what's the point here... on the backend we have a single model which represents list of chats in case of chat section and list of channels/categories in case of community section. That further means `itemId` in case of chat section will be `chatId`, but in case of community section it will be either `channelId` or `categoryId` (only in case if an item is a category, then it has `subItems` which is a model representing channels belonging to that category). But qml side doesn't follow that struct, old qml components got minimal updates in order to handle new backend model correctly, and qml part recognizes categoryId as a special property, different than chatId/channelId, and that's why the same model item is used on both places.

src/StatusQ/Components/StatusChatListAndCategories.qml
saledjenic: @jrainville I explained above what's like that. The other way is to have totally different component in the `statusq` which will know how to display chats/channels/category channels based on the single model from the backend. Actually backend model is very simple, we can look into it as a model whose items are either final items or group of items(in case of channel's category). 

It was a bit harder to apply classic `composite design pattern` in Nim, although it reminds of it, but once we switch to c++ it will be it (a common and easy technique for such cases).

src/StatusQ/Components/StatusChatList.qml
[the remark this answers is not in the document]
PascalPrecht: We do need to know though, if it's an identicon or not because the visuals change (there's a border rendered around the image, which is not the case when it's a normal profile picture). So we need some flag to know.

Various ways to go about this:

- Either introduce an `isIdenticon` flag,
- or introduce a more complex object like `model.image` which has all the information attached to it , similar to how it's set in QML

src/StatusQ/Components/StatusChatList.qml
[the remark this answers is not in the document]
PascalPrecht: Almost correct (two fixes):

1. When `hasNotification && notificationCount == 0` (case: has unread messages) we **don't** render a dot badge. The badge is **only** shown when `notifiactionCount > 0`. Maybe you confuse it with the badge rendered on the navigation tab bar icons, but this is individual chat items.
2.  When `hasNotification && notificationCount > 0` (case: unread mentions) then the chat item is **bold** as well.

So all in all there's just 2 cases (ignoring muted, inactive, normal and selected chat list items:

1. Text is bold (unread messages)
2. Text is bold + badge (mentions)

src/StatusQ/Components/StatusChatListAndCategories.qml
PascalPrecht: Do we lose functionality in the sandbox app when this is removed? Cause I think it relies on this property, but would have to double check

saledjenic: @PascalPrecht an item as well as subitem now have a property named `active`, so the old way `selected: model.chatId === statusChatList.selectedChatId` is now `selected: model.active`, and the value of `active` is maintained by the new refactored backend.

src/StatusQ/Components/StatusChatList.qml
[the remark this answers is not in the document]
saledjenic: > Text is bold (unread messages)
Text is bold + badge (mentions)

Exactly. :)

src/StatusQ/Components/StatusChatList.qml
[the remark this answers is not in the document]
saledjenic: @PascalPrecht thanks for this. 

> Either introduce an `isIdenticon` flag

This sounds ok, and easier do to than exposing an object in Nim. Will add it soon.

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -96,31 +92,19 @@ Column {
...
+                    name: model.name
+                    type: !!model.type ? model.type : StatusChatListItem.Type.CommunityChat
+                    muted: model.muted
+                    hasNotifications: model.hasNotification
```
PascalPrecht, line 102: Actually one more question here: shouldn't this be `hasNotifiactions` ?
And also... isn't `hasUnreadMessages` more correct? Not wanna bikeshed here, but... having notifications is one thing (that's why we have the `notificationCount`, having unread messages another)

src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -16,9 +16,8 @@ Rectangle {
     property string chatId: ""
     property string categoryId: ""
     property string name: ""
-    property alias badge: statusBadge
```
PascalPrecht, line 19: Please do not remove the `badge` API, this is needed in other places where we need more control over the component itself

PascalPrecht (changes requested):

src/StatusQ/Components/StatusChatList.qml
saledjenic: > Actually one more question here: shouldn't this be hasNotifiactions ?

@PascalPrecht on the BE side it's exposed as `hasNotification`, not a problem at all to update it.

> isn't hasUnreadMessages more correct?

@PascalPrecht depends on how you're looking into it :) `hasNotifications` is somehow general I would say, so if we're talking about unread messages, then `hasNotifications` signifies a notification about unread messages, if we're talking about mentions then it signifies that there are mention notifications, if we're talking about replies then it signifies that there are replies notifications.

> having notifications is one thing (that's why we have the notificationCount, having unread messages another)

A message which has mention or is a reply to any of my previous messages is in both case unread message. :)

All in all, they are all in real just unread messages but some contains only text, some mentions, some are replies, but all them are still unviewed/unseen/unread messages. So I agree with whatever you say regarding this. Just say how would you like to see this and I will update it?

src/StatusQ/Components/StatusChatListItem.qml
saledjenic: @PascalPrecht ok, no problem.

src/StatusQ/Components/StatusChatList.qml
PascalPrecht: I personally prefer `hasUnreadMessages`. We already know if we have notifications (`noticiationsCount > 0`). That should make it very explicit.

saledjenic (approved):

src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -95,8 +95,8 @@ Rectangle {
                 if (statusChatListItem.muted && !sensor.containsMouse && !statusChatListItem.highlighted) {
                     return 0.4
                 }
-                return statusChatListItem.hasMention ||
-                        statusChatListItem.hasUnreadMessages ||
+                return statusChatListItem.hasUnreadMessages ||
+                        statusChatListItem.notificationsCount > 0 ||
                         statusChatListItem.selected ||
                         statusChatListItem.highlighted ||
                         statusBadge.visible ||
```
PascalPrecht, line 102: Nit: Can remove `statusBadge.visible` from here, since we relying on `notificationsCount` already.

PascalPrecht (commented):
The changes LGTM!

However, I've just tested this for the first time (sorry I haven't done this earlier), and it seems this model change breaks the drag and drop functionality of channels.

This can be tested and verified in latest `master` in the sandbox app, where chat items can be dragged inside of community categories (can also be done in the normal chat list, but needs to be enabled there).

This does no longer work in this branch.

While I understand that this commit introduces breaking changes in the StatusChatList APIs, I don't think we want to introduce regressions in the features it supports, right?

src/StatusQ/Components/StatusChatListItem.qml
saledjenic: @PascalPrecht I am thinking now, since your previous request was that you want to have `statusBadge` exposed to the outside world from this component, then its visibility and value should not be determined inside the component at all. In that case a full and correct flexibility will be achieved by removing lines 21, 99, 142, 150, 189 and 190 from this file and set those 2 props from the outside of this component via exposed `statusBadge`. Any other case where we have `notificationsCount` and `statusBadge` as exposed parameters is kinda mixing contexts cause `notificationsCount` influences to the `statusBadge` value and visibility.

saledjenic:
> This does no longer work in this branch.

@PascalPrecht  yes, it doesn't work since we don't have BE methods which will accept and handle order changes made by dragging and dropping channels. That's left for the second step there, firstly wanted to make sure that we will have chat/channel list displayed well since entire logic on the BE is completely different.

> While I understand that this commit introduces breaking changes in the StatusChatList APIs, I don't think we want to introduce regressions in the features it supports, right?

@PascalPrecht  yes this is a breaking change, and in this moment it's impossible to do a regression checks since we don't have all features moved to the new BE.

PascalPrecht:
> , firstly wanted to make sure that we will have chat/channel list displayed well since entire logic on the BE is completely different.

I understand. Thing is though, it works **without** a backend. The component is capable of doing D'n'D and emitting signals so the backend part can **persist** those changes. But D'n'd in itself is a purely UI related thing. 

So IMO, the component should still be capabable of doing that and emitting signals that items need to be reordered in the backend.

At the end of the day, all we did in this PR is changing the shape of the model. We're not even touching its visuals.

I'd appreciate if we could explore how much effort is needed to preserve this in this branch.

saledjenic:
> Thing is though, it works without a backend.

@PascalPrecht in the desktop app there is no chance to work without backend. I've already seen everything there and  signals/slots you're talking about are in place, and that's ok, but they are not calling/rely on any BE methods (since we don't have them on the BE side) and without that there is no chance to work, the way how is that working in the DemoApp is other story.

> I'd appreciate if we could explore how much effort is needed to preserve this in this branch.

@PascalPrecht I guess not too much, we just need to add some BE methods and correctly accept and update model based on drag and drop actions taken on the UI side, that's all. I thought this may be done as a new ticket, what do you think?

src/StatusQ/Components/StatusChatList.qml
```diff
@@ -15,17 +15,14 @@ Column {
     width: 288
 
     property string categoryId: ""
-    property string selectedChatId: ""
-    property alias chatListItems: delegateModel
+    property var model: []
```
borismelnik, line 18: this can be alias

src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -20,11 +20,9 @@ Item {
         text: "More"
     }
 
-    property string selectedChatId: ""
+    property var model: []
```
borismelnik, line 23: can be alias

src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -71,26 +69,25 @@ Item {
             StatusChatList {
                 id: statusChatList
                 anchors.horizontalCenter: parent.horizontalCenter
-                visible: chatListItems.count > 0
-                selectedChatId: statusChatListAndCategories.selectedChatId
+                visible: statusChatList.model.count > 0
```
borismelnik, line 72: `property var model: []` this type doesn't have `count`

PascalPrecht:
> @PascalPrecht I guess not too much, we just need to add some BE methods and correctly accept and update model based on drag and drop actions taken on the UI side, that's all. I thought this may be done as a new ticket, what do you think?

@saledjenic I was primarily referring to, what's the effort to keep the d'n'd (which is purely frontend) in tact.
The concerns you're talking about don't apply to what I'm raising here because the d'n'd (actually being able to drag and drop items) is not related to any backend. Again, we've changed the model. Which I understand can affect this. However, just because there aren't any backend methods to persist some order of chat items, doesn't mean we can't order them in the front-end. In other words: d'n'd in the sandbox app should just work as before.

I don't want this to be a blocker now though because we have to make progress on the refactor front as well. I'll approve this now, but we need to revisit this.

PascalPrecht (approved):

saledjenic:
@PascalPrecht drag and drop feature works fine now.

PascalPrecht:
Awesome!!

PR #488, opened 2021-11-22
fix(StatusAppNavBar): Fixed position of profile button on the NavBar

Khushboo-dev-cpp:
fix(StatusAppNavBar): Fixed position of profile button on the NavBar

fixes #3521


PascalPrecht (approved):
Despite `23` being a very odd margin (should probably be `24`?) this LGTM

Khushboo-dev-cpp:
> Despite `23` being a very odd margin (should probably be `24`?) this LGTM

Hahah yes, but that's what it is in Figma. Do you want me to double check or something?

PascalPrecht:
Nah, I think it's fine. I'd expect it has to be 24 though

PR #491, opened 2021-11-29
feat(StatusListItem): Expose the Aside text so that it can be modified from the outside

Khushboo-dev-cpp:
feat(StatusListItem): Expose the Aside text so that it can be modified from the outside


PascalPrecht (approved):
Awesome changes!

PR #496, opened 2021-12-02
fix(StatusChatListItem): Fixes the issue of the chat list getting a highlight even when they are not really hovered

Khushboo-dev-cpp:
fix(StatusChatListItem): Fixes the issue of the chat list getting a highlight even when they are not really hovered

The issue is caused by the fact that the MouseArea returns containsMouse as true even when the item is not hovered.
The issue is reproduced when the model changes dynamically in the StatusChatList, thus causing the StatusChatListItem to be recreated and
this is when the containsMouse remains true even though the item is not hovered.
I see a bunch of issues in the QT bug list with regards to the containsMouse property but are still open.
A solution to this is to use HoverHandler which is a more reliable way getting hovered events for an item.

### DIFF
--- a/src/StatusQ/Components/StatusChatListItem.qml
+++ b/src/StatusQ/Components/StatusChatListItem.qml
@@ -58,7 +58,7 @@ Rectangle {
         if (selected) {
             return Theme.palette.statusChatListItem.selectedBackgroundColor
         }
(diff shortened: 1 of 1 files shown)


PascalPrecht (approved):
**NICE!**

Awesome you've figured this out @Khushboo-dev-cpp !

PR #503, opened 2021-12-21
refactor(StatusChatToolBar): open/close handlers get called if they a…

saledjenic:
…re set for `popupMenu` of `StatusChatToolBar` component

### DIFF
--- a/src/StatusQ/Components/StatusChatToolBar.qml
+++ b/src/StatusQ/Components/StatusChatToolBar.qml
@@ -93,6 +93,25 @@ Rectangle {
 
             property bool showMoreMenu: false
             onClicked: {
+                menuButton.highlighted = true
+
+                let originalOpenHandler = popupMenuSlot.item.openHandler
+                let originalCloseHandler = popupMenuSlot.item.closeHandler
+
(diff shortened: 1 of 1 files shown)


borismelnik (approved):

PascalPrecht:
I see this was being merged into `refactoring/base_bc`
It looks like however, this should be done in `master` (as well). It also seems like this is either a fix or a feature, depending on how you look at it.

@saledjenic Would you mind sending this against `master` as well? `refactoring/base_bc` then needs to be rebased on top of `master`

saledjenic:
@PascalPrecht sorry, if I had to wait for 2 approvals here as well, the point why I merged it after one approval is that I wanted to merge corresponding part in the `status-desktop` cause I need those changes for my further work. 

Since at the end (once we're done with refactor) this will go to the `master` branch definitely, does it necessary to do it now if there is no explicit need at the moment to have this in `master`?

PascalPrecht:
> does it necessary to do it now if there is no explicit need at the moment to have this in master?

Yea that's a good point. I thought, since it's a small change and improves the current state, it'd be nice to see that improvement in `master` as well. Probably not a big deal if it isn't right now.

PR #513, opened 2022-01-10
Added new Status Bar UI component for modal popups

noeliaSD:
Necessary to close [#3547](https://github.com/status-im/status-desktop/issues/3547)

**What does the PR do**
It has been created a new Status Bar UI component to be reused for modal popups whenever the status of a user wants to be highlighted.

### DIFF
--- a/src/StatusQ/Popups/qmldir
+++ b/src/StatusQ/Popups/qmldir
@@ -8,6 +8,7 @@ StatusMenuHeadline 0.1 StatusMenuHeadline.qml
 StatusModal 0.1 StatusModal.qml
 StatusSearchPopup 0.1 StatusSearchPopup.qml
 StatusModalDivider 0.1 StatusModalDivider.qml
+StatusModalStatusBar 0.1 StatusModalStatusBar.qml
 StatusSearchPopupMenuItem 0.1 StatusSearchPopupMenuItem.qml
 StatusSearchLocationMenu 0.1 StatusSearchLocationMenu.qml
 StatusSpellcheckingMenuItems 0.1 StatusSpellcheckingMenuItems.qml
--- a/src/StatusQ/Popups/qmldir
+++ b/src/StatusQ/Popups/qmldir
@@ -8,7 +8,6 @@ StatusMenuHeadline 0.1 StatusMenuHeadline.qml
 StatusModal 0.1 StatusModal.qml
 StatusSearchPopup 0.1 StatusSearchPopup.qml
 StatusModalDivider 0.1 StatusModalDivider.qml
-StatusModalStatusBar 0.1 StatusModalStatusBar.qml
 StatusSearchPopupMenuItem 0.1 StatusSearchPopupMenuItem.qml
 StatusSearchLocationMenu 0.1 StatusSearchLocationMenu.qml
 StatusSpellcheckingMenuItems 0.1 StatusSpellcheckingMenuItems.qml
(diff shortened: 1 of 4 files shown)


src/StatusQ/Popups/qmldir
```diff
@@ -8,6 +8,7 @@ StatusMenuHeadline 0.1 StatusMenuHeadline.qml
 StatusModal 0.1 StatusModal.qml
 StatusSearchPopup 0.1 StatusSearchPopup.qml
 StatusModalDivider 0.1 StatusModalDivider.qml
+StatusModalStatusBar 0.1 StatusModalStatusBar.qml
```
alaibe, line 11: is it dedicated to modal? Should it be StatusBanner?

alexandraB99 (commented):
looks good to me, however I am not sure whether it makes sense for this to be a component in StatusQ, is it used only in the blocked user case? If so, it could just be part of that specific popup instead?

src/StatusQ/Popups/qmldir
noeliaSD: Certainly, it is not strictly necessary to be part of a modal. It fits with parent width so we could move it maybe to **StatusQ.Controls**. Is it a better location? 

It is a general approximation as to be more generalist it should incorporate some text length checks but it is something we can improve in the future.

It sounds good **StatusBanner**. I’ll change the component name to that.

noeliaSD:
> 
It is currently not used in other contexts but as it is a general approximation it could be used in the future to display other status banner, not only the user Blocked one, as it can be customized with text and colours.

Nowadays it can also be used in another view, _ChatColumnView.qml_ where the blocked status bar is already displayed and is currently defined in the own component. We can replace it by the new ui control. What do you think?

noeliaSD:
I've aready pushed an update. Please review it again @alaibe and @alexandraB99!

alaibe (approved):
LGTM

alexandraB99:
@noeliaSD thanks for that! StatusBanner and putting in controls sounds great! could you please also update the commit message and description accordingly?

noeliaSD:
> @noeliaSD thanks for that! StatusBanner and putting in controls sounds great! could you please also update the commit message and description accordingly?

I’ve referenced the StatusBanner.qml control name and location into the squashed commit. Thank you!

PascalPrecht:
Hey friends, sorry I've missed this PR. I just saw it in `master` now and left some comments on the commit. Gonna echo them here:

@noeliaSD Very cool to see your first contribution here, just a couple of thoughts:

When we introduce a new component (or variant/feature) we usually add a page for that in the sandbox app so we can actually test that the component works as intended
Commits here follow a certain convention. This is needed so they'll be picked up by the changelog generator: https://github.com/status-im/StatusQ/blob/master/CHANGELOG.md (the conventions you can find here: https://www.conventionalcommits.org/en/v1.0.0/)
Generally, changes should be sent as a PR so we get a chance to review. There seem to be a few things in here that need update. Would've been cool if we caught those before they land in master

noeliaSD:
Regarding the commits conventions... I now better understand what @alexandraB99 refer by telling "update commit message and descrption". I reviewed that and wrote a clear one when squaching but didn't follow the convention, sorry!

PascalPrecht:
>  I reviewed that and wrote a clear one when squaching but didn't follow the convention, sorry!

No problem at all. Nobody got it right the first time so far lol

PR #514, opened 2022-01-12
chore(sandbox/controls): Added `StatusBanner` component in controls view

noeliaSD:
### What does the PR do

In order to test StatusBanner.qml ui component it has been added into the Sandbox app as another Control

### Affected areas

Controls/StatusBanner

PRStatusBanner

### DIFF
--- a/sandbox/Controls.qml
+++ b/sandbox/Controls.qml
@@ -188,4 +188,14 @@ GridLayout {
         to: 100
         value: 40
     }
+
+    StatusBanner {
+        id: banner
+        width: 360
+        topPadding: 20
+        backgroundColor: Theme.palette.primaryColor3
+        bordersColor: Theme.palette.primaryColor2
+        fontColor: Theme.palette.primaryColor1
+        statusText: "Banner"
+    }
 }
--- a/sandbox/Controls.qml
+++ b/sandbox/Controls.qml
@@ -193,9 +193,7 @@ GridLayout {
         id: banner
         width: 360
         topPadding: 20
-        backgroundColor: Theme.palette.primaryColor3
-        bordersColor: Theme.palette.primaryColor2
-        fontColor: Theme.palette.primaryColor1
+        type: StatusBanner.Type.Danger
         statusText: "Banner"
     }
 }
--- a/src/StatusQ/Controls/StatusBanner.qml
+++ b/src/StatusQ/Controls/StatusBanner.qml
@@ -62,28 +62,24 @@ Column {
     // Behavior
     states: [
         State {
-            name: "Info"
             when: statusBanner.type === StatusBanner.Type.Info
             PropertyChanges { target: d; backgroundColor: Theme.palette.primaryColor3}
             PropertyChanges { target: d; bordersColor: Theme.palette.primaryColor2}
             PropertyChanges { target: d; fontColor: Theme.palette.primaryColor1}
         },
         State {
-            name: "Danger"
             when: statusBanner.type === StatusBanner.Type.Danger
             PropertyChanges { target: d; backgroundColor: Theme.palette.dangerColor3}
             PropertyChanges { target: d; bordersColor: Theme.palette.dangerColor2}
             PropertyChanges { target: d; fontColor: Theme.palette.dangerColor1}
         },
         State {
-            name: "Success"
             when: statusBanner.type === StatusBanner.Type.Success
             PropertyChanges { target: d; backgroundColor: Theme.palette.successColor2}
-            PropertyChanges { target: d; bordersColor: Theme.palette.successColor3}
+            PropertyChanges { target: d; bordersColor: Theme.palette.successColor2}
             PropertyChanges { target: d; fontColor: Theme.palette.successColor1}
         },
         State {
-            name: "Warning"
             when: statusBanner.type === StatusBanner.Type.Warning
             PropertyChanges { target: d; backgroundColor: Theme.palette.pinColor3}
             PropertyChanges { target: d; bordersColor: Theme.palette.pinColor2}
--- a/src/StatusQ/Controls/StatusBanner.qml
+++ b/src/StatusQ/Controls/StatusBanner.qml
@@ -14,49 +14,23 @@ Column {
 
     // "private" properties
     QtObject {
-           id: pProp
+           id: d
            property color backgroundColor
            property color bordersColor
            property color fontColor
     }
 
     enum Type {
-        None, // 0
-        Info, // 1
-        Danger, // 2
-        Success, // 3
-        Warning // 4
-    }
-
-    function setBannerBackgroundColor() {
-        if(statusBanner.type === StatusBanner.Type.Info) {
-            pProp.backgroundColor = Theme.palette.primaryColor3
-            pProp.bordersColor = Theme.palette.primaryColor2
-            pProp.fontColor = Theme.palette.primaryColor1
-        }
-        else if(statusBanner.type === StatusBanner.Type.Danger) {
-            pProp.backgroundColor = Theme.palette.dangerColor3
-            pProp.bordersColor = Theme.palette.dangerColor2
-            pProp.fontColor = Theme.palette.dangerColor1
-        }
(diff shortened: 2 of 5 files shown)


noeliaSD:
Collecting @PascalPrecht comments from [commit c56dcd9](https://github.com/status-im/StatusQ/commit/c56dcd9eb47ad8e22d24257b53bbc12ba407357f#commitcomment-63733961):

1. Instead of assigning concrete colors here, `StatusBanner` should rather have variants for `Success` and `Danger` (and possibly others if needed). 

2. I wouldn't bother with adding new colors to `ui/imports/Themes` as those are legacy themes.
`StatusQ.Core.Theme` is where the party's at. And we already have `dangerColor` variants for those.

sandbox/Controls.qml
```diff
@@ -188,4 +188,14 @@ GridLayout {
...
         value: 40
     }
+
+    StatusBanner {
+        id: banner
+        width: 360
+        topPadding: 20
+        backgroundColor: Theme.palette.primaryColor3
+        bordersColor: Theme.palette.primaryColor2
+        fontColor: Theme.palette.primaryColor1
```
PascalPrecht, line 198: So you see here how you're forced to set `backgroundColor`, `bordersColor` and `fontColor`?

I don't want to worry about that as a user. I just wanna say: "Render a warning banner" or "render a success banner".

So accomplish this it'd be better of `StatusBanner` had a `type` which is an enum to allow consumers of this API to choose what kind of banner they want:

```qml
StatusBanner {
  text: "This is a warning banner"
  type: StatusBanner.Type.Danger
}

StatusBanner {
  text: "This is a success banner"
  type: StatusBanner.Type.Success
}

StatusBanner {
  text: "This is a default banner"
}
```

The default could use those `primaryColor`s you've used here.

PascalPrecht (changes requested):

PascalPrecht:
One note about the commit message:

What you've done is technically correct, but we only want to mark things as feature when they are actual features of the StatusQ API. You adding the component to the sandbox app doesn't introduce a new feature to StatusQ, so you could probably just do someting like:

```
chore: add `StatusBanner` variants to sandbox app
```

^ Normally those commits wouldn't exist because these changes would have been part of a commit that introduces `StatusBanner`, which indeed would a feature.

But as you know, your banner has already landed in `master` so we can't undo this.

sandbox/Controls.qml
PascalPrecht: There's a bunch of components in StatusQ that do this kind of thing. Here's one example to give you some inspiration: https://github.com/status-im/StatusQ/blob/master/src/StatusQ/Components/StatusExpandableItem.qml#L35-L39

noeliaSD: Yes, got it! Great idea! I'm working on that

noeliaSD:
Please @PascalPrecht and @alexandraB99 review again the pushed changes.

src/StatusQ/Controls/StatusBanner.qml
```diff
@@ -1,31 +1,71 @@
...
+
+    enum Type {
+        None, // 0
+        Info, // 1
+        Danger, // 2
+        Success, // 3
+        Warning // 4
+    }
+
+    function setBannerBackgroundColor() {
```
alexandraB99, line 31: @noeliaSD could you please implement this using states? it makes more sense and will look cleaner having corresponding states to type changes. also I see all property changes are happening in private component's properties.

src/StatusQ/Controls/StatusBanner.qml
```diff
@@ -1,31 +1,71 @@
...
-    property color bordersColor
-    property color fontColor
     property string statusText
+    property int type: StatusBanner.Type.Info
     property int textPixels: 15
     property int statusBannerHeight: 38
 
+    // "private" properties
+    QtObject {
+           id: pProp
```
alexandraB99, line 17: would it make sense to id this `private` or `d`?

alexandraB99 (changes requested):

noeliaSD:
@alexandraB99 and @PascalPrecht I've pushed an update, please check again!

src/StatusQ/Core/Theme/StatusDarkTheme.qml
```diff
@@ -110,6 +110,7 @@ ThemePalette {
 
     successColor1: getColor('green3')
     successColor2: getColor('green3', 0.2)
+    successColor3: getColor('green3', 0.3)
```
PascalPrecht, line 113: This doesn't seem to exist in the design system: https://www.figma.com/file/MtAO3a7HnEH5xjCDVNilS7/%F0%9F%8E%A8-Design-System-%E2%8E%9C-Desktop?node-id=1%3A10

Has this color been discussed with design?

src/StatusQ/Controls/StatusBanner.qml
```diff
@@ -1,31 +1,45 @@
...
-    property color bordersColor
-    property color fontColor
     property string statusText
+    property int type: StatusBanner.Type.Info
     property int textPixels: 15
     property int statusBannerHeight: 38
 
+    // "private" properties
+    QtObject {
+           id: d
```
PascalPrecht, line 17: Any reason this is called `d`? has this some sort of semantic meaning?

PascalPrecht (commented):
All in all LGTM, we just need to double check that new green color you've added to the color system.

src/StatusQ/Controls/StatusBanner.qml
noeliaSD: Here in [Qt encoding conventions](https://doc.qt.io/archives/qt-4.8/qml-coding-conventions.html#private-properties) they propose it like this, but if you have another specific convention, I can change without problem. Share with me which ones you follow, please!

src/StatusQ/Controls/StatusBanner.qml
```diff
@@ -34,14 +48,46 @@ Column {
...
-        color: statusBanner.bordersColor
+        color: d.bordersColor
         height: 1
         width: parent.width
     }
+
+    // Behavior
+    states: [
+        State {
+            name: "Info"
```
alexandraB99, line 65: Great @noeliaSD, thanks! since we set the `when` property then `name` is not needed, or is it used elsewhere I missed?

src/StatusQ/Core/Theme/StatusDarkTheme.qml
PascalPrecht: So generally I'd say, let's just leave it as is. But as soon as we rely to much on `successColor3` in the future, this can become a hassle if there was indeed an officially introduced `successColor3`.

I guess it's fine to take this risk, as the color system hasn't been touched since we set it up originally, so it should be quite timeliness. Nonetheless, I wouldn't mind, if you double check with the design team if there's plans for such a color in our color system, and at least bring this to their attention.

src/StatusQ/Controls/StatusBanner.qml
PascalPrecht: No need to change this! Was just curious as `d` doesn't seem to have any particular meaning.

PascalPrecht:
@noeliaSD I guess we can merge this, once @alexandraB99 last comment was addressed!

src/StatusQ/Core/Theme/StatusDarkTheme.qml
PascalPrecht: > Do you think in this specific Success case, border colours will be like background (so using successColor2)?

Actually I think we should do this instead. Turns out there is indeed designs of a success variant that do exactly that.

noeliaSD:
Pushed requested modifications. Please review it again @alexandraB99 and @PascalPrecht.

PascalPrecht (approved):
LGTM!

Can you squash your commits into one and make it a `feat` commit (since you're introducing the `type` variants?

Something like: 

```
feat(StatusBanner): introduce type variants for different banner styles
```

alexandraB99 (approved):
Thanks @noeliaSD looks great! Just please squash the commits as Pascal already suggested!

PR #533, opened 2022-01-18
fix(build): broken MacOS build

alexandraB99:
moved DESTDIR in sandbox.pro above copydata command as it was breaking the build process on MacOS

### DIFF
--- a/sandbox/sandbox.pro
+++ b/sandbox/sandbox.pro
@@ -22,6 +22,7 @@ SOURCES += \
 }
 
 TARGET = sandboxapp
+DESTDIR = $$PWD/bin
 
 macx {
     CONFIG -= app_bundle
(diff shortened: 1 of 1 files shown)


PascalPrecht (approved):
This LGTM!

Bonus points if you make the subject of the commit message more meaningful

alexandraB99:
> This LGTM!
> 
> Bonus points if you make the subject of the commit message more meaningful

hmm I took [this](https://github.com/status-im/StatusQ/commit/14ab4ce52d5cd4899b35e71c99f1dfb3eaa40305) as an example :D ok I'll try to explain a bit more..

borismelnik (approved):
Great! Don't understand how it works before :D

PascalPrecht:
> hmm I took this as an example :D ok I'll try to explain a bit more..

That is very fair. That other build commit should've had more meaning as well.
Generally, it's good ask yourself: 

"What will this look like in the changelog, and how can it be written so it has useful meaning?"

With a commit like `fix(build): macos build`, what we get is a changelog entry that says:

```
## Fixes
  - **build**: macos build
```

Anyways, not a huge biggie. I'd just like to avoid commit messages that aren't useful (also when working with the commit history later on)

PR #534, opened 2022-01-18
refactor: refactored/cleaned up sandbox app

alexandraB99:
Grouped QML files by pages, controls and demo app
related stuff

PR #535, opened 2022-01-19
feat(StatusPasswordStrengthIndicator): Add password strength indicator component

### LINKED ISSUE #528: Add `StatusPasswordStrengthIndicator`
figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=11%3A4921

Screen Shot 2022-01-12 at 4 17 05 PM

noeliaSD:
Closes #528

### What does the PR do

Create a password strength indicator component with the corresponding variations.

Create a base `StatusProgressBar` component.

Add new `miscColor12` agreed with designer.

### Affected areas

Controls/StatusProgressBar
Controls/StatusPasswordStrengthIndicator

Light mode:
image

Dark mode:
image

### DIFF
--- a/src/StatusQ/Controls/StatusPasswordStrengthIndicator.qml
+++ b/src/StatusQ/Controls/StatusPasswordStrengthIndicator.qml
@@ -1,0 +1,54 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.14
+
+import StatusQ.Core.Theme 0.1
+import StatusQ.Core 0.1
+
+StatusProgressBar {
+    id: control
+
+    property var strength
+    property string labelVeryWeak: "Very weak"
+    property string labelWeak: "Weak"
+    property string labelSoso: "So-so"
+    property string labelGood: "Good"
+    property string labelGreat: "Great"
+
+    enum Strength {
+        VeryWeak, // 0
+        Weak, // 1
+        SoSo, // 2
+        Good, // 3
+        Great // 4
+    }
+
+    // Behavior:
+    states: [
+        // Strength states definition:
+        State {
+            when: control.strength === StatusPasswordStrengthIndicator.Strength.VeryWeak
+            PropertyChanges { target: control; fillColor : Theme.palette.dangerColor1}
+            PropertyChanges { target: control; text: labelVeryWeak}
+        },
+        State {
+            when: control.strength === StatusPasswordStrengthIndicator.Strength.Weak
+            PropertyChanges { target: control; fillColor : Theme.palette.pinColor1}
+            PropertyChanges { target: control; text: labelWeak}
+        },
+        State {
+            when: control.strength === StatusPasswordStrengthIndicator.Strength.SoSo
+            PropertyChanges { target: control; fillColor : Theme.palette.miscColor7}
+            PropertyChanges { target: control; text: labelSoso}
+        },
+        State {
+            when: control.strength === StatusPasswordStrengthIndicator.Strength.Good
+            PropertyChanges { target: control; fillColor : Theme.palette.miscColor12}
+            PropertyChanges { target: control; text: labelGood}
+        },
+        State {
+            when: control.strength === StatusPasswordStrengthIndicator.Strength.Great
+            PropertyChanges { target: control; fillColor : Theme.palette.successColor1}
+            PropertyChanges { target: control; text: labelGreat}
+        }
+    ]
+}
--- a/statusq.qrc
+++ b/statusq.qrc
@@ -316,5 +316,7 @@
         <file>src/assets/img/icons/tiny/tribute-to-talk.svg</file>
         <file>src/assets/img/icons/tiny/warning.svg</file>
         <file>src/StatusQ/Controls/StatusBanner.qml</file>
+        <file>src/StatusQ/Controls/StatusProgressBar.qml</file>
+        <file>src/StatusQ/Controls/StatusPasswordStrengthIndicator.qml</file>
     </qresource>
 </RCC>
--- a/sandbox/controls/Controls.qml
+++ b/sandbox/controls/Controls.qml
@@ -196,4 +196,11 @@ GridLayout {
         type: StatusBanner.Type.Danger
         statusText: "Banner"
     }
+
+    StatusProgressBar {
+        id: progressBar        
+        text: "Weak"
+        value: 0.5
+        fillColor : Theme.palette.pinColor1
+    }
 }
--- a/sandbox/main.qml
+++ b/sandbox/main.qml
@@ -227,6 +227,11 @@ StatusWindow {
                             selected: viewLoader.source.toString().includes(title)
                             onClicked: mainPageView.page("StatusExpandableSettingsItem");
                         }
+                        StatusNavigationListItem {
+                            title: "StatusPasswordStrengthIndicator"
+                            selected: viewLoader.source.toString().includes(title)
+                            onClicked: mainPageView.page(title);
+                        }
                         StatusListSectionHeadline { text: "StatusQ.Popup" }
                         StatusNavigationListItem { 
                             title: "StatusPopupMenu"
--- a/sandbox/pages/StatusPasswordStrengthIndicatorPage.qml
+++ b/sandbox/pages/StatusPasswordStrengthIndicatorPage.qml
@@ -1,0 +1,38 @@
+import QtQuick.Layouts 1.14
+import StatusQ.Controls 0.1
+
+GridLayout {
+    columns: 1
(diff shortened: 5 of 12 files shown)


sandbox/Controls.qml
```diff
@@ -196,4 +196,41 @@ GridLayout {
...
+        id: goodPw
+        strength: StatusPasswordStrengthIndicator.Strength.Good
+        value: 1
+    }
+
+    StatusPasswordStrengthIndicator {
+        id: greatPw
+        strength: StatusPasswordStrengthIndicator.Strength.Great
+        value: 0.3
+    }
```
PascalPrecht, line 235: Can you please put these into their own "page" component? Eventually `Controls` will go away. Every component should have its own with all its variations.

src/StatusQ/Controls/StatusPasswordStrengthIndicator.qml
```diff
@@ -0,0 +1,49 @@
...
+        },
+        State {
+            when: control.strength === StatusPasswordStrengthIndicator.Strength.Good
+            PropertyChanges { target: control; fillColor : Theme.palette.miscColor12}
+            PropertyChanges { target: control; text: qsTr("Good")}
+        },
+        State {
+            when: control.strength === StatusPasswordStrengthIndicator.Strength.Great
+            PropertyChanges { target: control; fillColor : Theme.palette.successColor1}
+            PropertyChanges { target: control; text: qsTr("Great")}
```
PascalPrecht, line 46: So far we have avoided putting `qsTr()` calls in StatusQ. The reason being that status desktop will replace all occurances of `qsTr()` with `qsTrId()`.

So essentially, it will change the source of StatusQ when it performs i18n.
Hence, we rather **expose** text properties (with a decent default) so apps like Status Desktop can then assign a `qsTr` from the outside.

Can you please change that accordingly?

statusq.qrc
```diff
@@ -315,5 +315,6 @@
         <file>src/assets/img/icons/tiny/time.svg</file>
         <file>src/assets/img/icons/tiny/tribute-to-talk.svg</file>
         <file>src/assets/img/icons/tiny/warning.svg</file>
+        <file>src/StatusQ/Controls/StatusProgressBar.qml</file>
```
PascalPrecht, line 318: This misses the strength indicator component.

PascalPrecht (changes requested):
Very well done @noeliaSD !
There are a few minor things that need change.

Also, can you put the `Closes #...` part of your commit message at the **end** of your commit message?

Not 100% GitHub will recognize it otherwise.
Nonetheless, to follow conventional commits, this information has to be in the footer of the message.

Great work!

src/StatusQ/Controls/StatusPasswordStrengthIndicator.qml
noeliaSD: But by doing that (exposing the text property) you are ‘losing’ some automation in the component itself so you cannot link the variant with the text, only with the colour.  Or do you mean exposing one text property per variant so in this case having 5 different text properties?

And I’m wondering, is there another way of doing that as for example establishing here in StatusQ a unique text ID and then, status-desktop will resolve it accordingly (maybe having a common IDs file or similar)?  For example establishing a hierarchical ID structure like this:
`StatusQ.{ComponentName}.{textId}`

PascalPrecht: > Or do you mean exposing one text property per variant so in this case having 5 different text properties?

Exactly. For now that would be something like:

```
StatusPaswordStrengthIndicator {
  labelVeryWeak: ...
  labelWeak: ...
  labelSoso: ...
  ...
}
```

Again, these properties would have some decent defaults (`"Very weak"`, `"Weak"`, ... etc), but we would give consuming applications full control over how they want to solve i18n.

> And I’m wondering, is there another way of doing that as for example establishing here in StatusQ a unique text ID and then, status-desktop will resolve it accordingly (maybe having a common IDs file or similar)? For example establishing a hierarchical ID structure like this:
StatusQ.{ComponentName}.{textId}

Sounds like a good idea. Feel free to open a separate issue for that where this can be discussed!

PascalPrecht (approved):
This looks very good to me now.
However, I've just merged #534, so you'll have to rebase your's on top of latest `master`

PascalPrecht:
@noeliaSD if you rebase this today, we can merge it and release it next week

noeliaSD:
> @noeliaSD if you rebase this today, we can merge it and release it next week

I'm waiting review from @alexandraB99 and @abhi-status, as well!

sandbox/Controls.qml
```diff
@@ -196,4 +196,11 @@ GridLayout {
         type: StatusBanner.Type.Danger
         statusText: "Banner"
     }
+
+    StatusProgressBar {
+        id: progressBar        
+        text: "Weak"
+        value: 0.5
+        fillColor : Theme.palette.pinColor1
+    }
```
alexandraB99, line 205: Shouldn't we have here `StatusPasswordStrengthIndicator` as well?

alexandraB99 (changes requested):
@noeliaSD please rebase this

alexandraB99:
@noeliaSD feel free to merge as soon as you rebase, I am good with the changes. thanks!

sandbox/Controls.qml
noeliaSD: I've created a specific page in sandbox that includes all `StatusPasswordStrengthIndicator` variants and only include the `StatusProgressBar` in Controls, as it is a base one.

noeliaSD:
> @noeliaSD if you rebase this today, we can merge it and release it next week

@PascalPrecht please check and if ok, you can merge it!

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/626695da25d86878336d4056d568c2bbb1b4bd8c

Great work @noeliaSD

src/StatusQ/Controls/StatusPasswordStrengthIndicator.qml
osmaczko: > Hence, we rather expose text properties (with a decent default) so apps like Status Desktop can then assign a qsTr from the outside.

@PascalPrecht does it mean that if we don't provide values for these exposed properties (by relying on default ones that come with component) the translations will simply not work?

I am almost sure it will result in many bugs related to missing translations. Shouldn't we avoid providing these default strings?

With Qt 5.15 comes the 'required' keyword which ensures each property is assigned by the client who uses the Component.
```
    required property string labelVeryWeak
    required property string labelWeak
    required property string labelSoso
    required property string labelGood
    required property string labelGreat
```

PascalPrecht: > @PascalPrecht does it mean that if we don't provide values for these exposed properties (by relying on default ones that come with component) the translations will simply not work?
 
 If they aren't marked with `qsTr()` the i18n script of Status Desktop won't touch it (otherwise it will replace it with `qsTrId()`.
 
 So yea, it will result in you seeing the default label.
 
 > Shouldn't we avoid providing these default strings?
 
 One could argue that, if no default strings are provided, you'll most likely see many more bugs (e.g. components having no labels at all). So I guess given those two (non-optimal options): which one is worse? No labels and broken UI, or at least english labels in any case?
 
 > With Qt 5.15 comes the 'required' keyword which ensures each property is assigned by the client who uses the Component.
 
 Sounds like a reasonable thing to do then, if we don't come up with a better i18n solution!
 Should be noted that I'm not super familiar with how qsTr/Id() are **supposed** to work. Maybe Desktop isn't doing the right thing in the first place. As in, are apps really required to rewrite their source (from qsTr() -> qsTrId()) to make i18n work? Or is there actually a different, recommended way to use this?

osmaczko: > Should be noted that I'm not super familiar with how qsTr/Id() are supposed to work. Maybe Desktop isn't doing the right thing in the first place. As in, are apps really required to rewrite their source (from qsTr() -> qsTrId()) to make i18n work? Or is there actually a different, recommended way to use this?

That's a good point, I've wondered about that before too. In general `qsTr()` is enough for translations, it does not need to be converted to `qsTrId()`. It seems we are doing that to re-use translations from status-react which are id-based: https://hackmd.io/@status-desktop/r1ytYSzn_

I think we could avoid running `qstrConverter.js` (and so the conversion from `qsTr()` to `qsTrId()`) by enhancing the `xmlTranslator.js` script with appropriate logic.

PR #536, opened 2022-01-19
feat(StatusIconTabButton): enable changing IconTabButton hovered color

jrainville:


PascalPrecht:
Left a comment here: https://github.com/status-im/status-desktop/pull/4501#issuecomment-1016750544
Not sure we really want this because this is by design.

PR #537, opened 2022-01-19
feat(StatusQ.Components): Adding StatusTagSelector component

### LINKED ISSUE #526: Implement `StatusTagSelector` component
figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=391%3A100984

Screen Shot 2022-01-12 at 4 08 52 PM

Screen Shot 2022-01-12 at 4 08 17 PM

alexandraB99:
Added StatusTagSelector component needed for creating new chat channels, either one on one or group based on updated designs on Figma

Also added corresponding page in API Documentation

Closes #526

https://user-images.githubusercontent.com/31625338/150600131-9f9ed21b-b3a1-4efc-9637-131a727dc2cf.mov

https://user-images.githubusercontent.com/31625338/150442686-f24ff26e-03a4-47a3-a0fb-337e6a387768.mov


src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,109 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.12
+import QtQuick.Controls 2.14
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+//TODO check colors, sizes, margins, general clean up
```
PascalPrecht, line 8: So.. is this component a work-in-progress?

src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,109 @@
...
+        RowLayout {
+            anchors.fill: parent
+            anchors.leftMargin: 16
+            anchors.rightMargin: 16
+            spacing: 8
+            StatusBaseText {
+                Layout.preferredWidth: 22
+                Layout.alignment: Qt.AlignVCenter | Qt.AlignLeft
+                color: Theme.palette.baseColor1
+                text: "To: "
```
PascalPrecht, line 34: This needs to be exposed for i18n

src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,109 @@
...
+                        text: name
+                    }
+                    ToolButton {
+                        width: 24
+                        height: 24
+                        anchors.verticalCenter: parent.verticalCenter
+                        anchors.left: nameText.right
+                        background: null
+                        contentItem: StatusIcon {
+                            color: Theme.palette.white
```
PascalPrecht, line 73: Does this work across themes?

src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,109 @@
...
+                }
+            }
+
+            StatusBaseText {
+                visible: (namesModel.count === 5)
+                Layout.preferredWidth: 120
+                Layout.alignment: Qt.AlignVCenter | Qt.AlignRight
+                font.pixelSize: 10
+                color: Theme.palette.dangerColor1
+                text: "5 USER LIMIT REACHED"
```
PascalPrecht, line 105: This is a tag selector component that doesn't necessarily know the semantics of "users".

So few thoughts:

1. This message shouldn't be hardcoded (needs to be i18n'able anyways)
2. Would be cool if this component would leverage `StatusValidator` and then this message would be retrieved from the validator instance

Do you think it's feasible to replace `TextEdit` with `StatusBaseInput` (and tweak it as needed) so we get validation capabilities?

Ultimately, all (custom) controls should have validation built-in following the same APIs

PascalPrecht (changes requested):
That's a cool start @alexandraB99 !

There are some things that need change.
Also:

1. Can you add a page component for this component accordingly, so it can actually be tested? I had no way to give this a spin...
2. Bonus points if the commit message includes the `Status.Controls` scope

Let me know what you think!

PascalPrecht:
Ah, seeing now that you flagged it as "don't merge" so I guess it **is** a WIP.

alexandraB99:
> That's a cool start @alexandraB99 !
> 
> There are some things that need change. Also:
> 
> 1. Can you add a page component for this component accordingly, so it can actually be tested? I had no way to give this a spin...
> 2. Bonus points if the commit message includes the `StatusQ.Controls` scope
> 
> Let me know what you think!

@PascalPrecht yes, that is WIP :) Nevertheless I added the page and updated the commit message. It's good that I receive your comments from now so that I can work on those in one round :)

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: Thought about it (didn't try though), the problem is the internal Rectangle it has that paints the grey background and that's not customizable, unless I hide it, but still the red borders and the whole "invalid state" UI will look weird. Other than this, I don't see any invalid state [here](https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba⎜Desktop?node-id=391%3A100984)

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: I see all status components are using this, is this not supposed to be working?

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: done

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: done

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: more or less. I need to double check colors, fonts, etc but I believe most of functionality is already there

src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+                Layout.alignment: Qt.AlignVCenter
+                model: namesModel
+                orientation: ListView.Horizontal
+                spacing: 8
+                clip: true
+                onContentWidthChanged: {
+                    positionViewAtEnd();
+                }
+
+                delegate: Rectangle {
```
PascalPrecht, line 67: Realising now that this is the actual tag component.
I wonder if we should make some sort of button out of that. At, least, it should have proper hover effects/visuals (even if only on the "X" icon).

src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+                        id: nameText
+                        anchors.left: parent.left
+                        anchors.leftMargin: 8
+                        anchors.verticalCenter: parent.verticalCenter
+                        color: Theme.palette.white
+                        text: name
+                    }
+                    ToolButton {
+                        width: 24
+                        height: 24
```
PascalPrecht, line 83: Not exactly sure what `ToolButton` is, but the size seems to be off. The icon needs to be rendered a bit bigger.

Also, can you ensure this turns into a cursor pointer when hovered?

src/StatusQ/Controls/StatusTagSelector.qml
PascalPrecht: Ah, my point was rather that, what is `white` here should probably not be `white` across all themes. Judging by other designs and components (e.g. StatusBadge), this should most likely become black in dark theme. So it has to be `Theme.palette.indirectColor`.

src/StatusQ/Controls/StatusTagSelector.qml
PascalPrecht: Okay, that's fair for now!

src/StatusQ/Controls/StatusTagSelector.qml
PascalPrecht: Gotcha!

src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+        RowLayout {
+            anchors.fill: parent
+            anchors.leftMargin: 16
+            anchors.rightMargin: 16
+            spacing: 8
+            StatusBaseText {
+                Layout.preferredWidth: 22
+                Layout.alignment: Qt.AlignVCenter | Qt.AlignLeft
+                color: Theme.palette.baseColor1
+                text: qsTr("To: ")
```
PascalPrecht, line 49: This still needs to be exposed. Probably as something like `toLabel`.

Remember that Status Desktop replaces `qsTr()` with `qsTrId()` in the sources (including vendors). That's why we expose all i18n'able labels so consumers of components can decide how they want to do i18n (e.g. using `qsTr()` or any other method)

sandbox/demoapp/CreateChatView.qml
```diff
@@ -0,0 +1,215 @@
...
+                console.log("Confirm Join public chat");
+            }
+        }
+
+        StatusIconTabButton {
+            icon.name: "notification"
+            onClicked: {
+                print("create chat...");
+            }
+        }
```
PascalPrecht, line 67: Cosmetics, but we can probably remove this one. The notification bell in the designs come from the `StatusChatToolBar` so I guess we 

1. Either use that here and draw the `StatusTagSelector` on top of it
2. Teach `StatusChatToolBar` to render components inside of it (like `StatusTagSelector`)
3. Just remove this notification thingy and don't assume that this is supposed to be the actual chat tool bar

Up to you!

PascalPrecht (changes requested):
Thanks for the update @alexandraB99 !
This is getting in good shape. As you said, the base functionality is there and some visuals need to be changed.

I've left some more inline comments to address those.

In addition:

1. I think at the very least, users should be able to backspace items from the list. Being only able to "x" them out with the mouse is a big downside. 
2. The "X" of the tag component needs hover indicators. There aren't any in the design, so I'd check with the designers what that should look like (in both, light and dark mode)
3. Wonder if even clicking the entire tag should remove it (not just the "X"), in which case it needs updated hover visuals as well. But that is something design should decide.
4. There's this thing where the "too many users" message reserves a good amount of spaces that results in funny UI. In the screen shot below there was enough visual space, still the latest item pushed the tag list into the scroll space. 

Screenshot 2022-01-21 at 11 21 12

Maybe we can week the available space for tags as big as the input **until** the limit is reached?

5. Darkmode needs some love, in the component as well as in the demo application
Screenshot 2022-01-21 at 11 23 31

Great work so far!!

sandbox/demoapp/CreateChatView.qml
alexandraB99: you are right, removed!

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: exposed :)

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: this is the delegate, correct. There is no hovered state defined on Figma for this nevertheless, the `StatusIconTabButton` is used for the close icon which already has a hovered effect.

src/StatusQ/Controls/StatusTagSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+                    width: (nameText.contentWidth + 30)
+                    height: 30
+                    color: Theme.palette.primaryColor1
+                    radius: 8
+                    StatusBaseText {
+                        id: nameText
+                        anchors.left: parent.left
+                        anchors.leftMargin: 8
+                        anchors.verticalCenter: parent.verticalCenter
+                        color: Theme.palette.white
```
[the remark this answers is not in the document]
alexandraB99: done

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: part of WIP, changed to `StatusIconTabButton`

src/StatusQ/Controls/StatusTagSelector.qml
alexandraB99: done

alexandraB99:
> Thanks for the update @alexandraB99 ! This is getting in good shape. As you said, the base functionality is there and some visuals need to be changed.
> 
> I've left some more inline comments to address those.
> 
> In addition:
> 
> 1. I think at the very least, users should be able to backspace items from the list. Being only able to "x" them out with the mouse is a big downside.
> 2. The "X" of the tag component needs hover indicators. There aren't any in the design, so I'd check with the designers what that should look like (in both, light and dark mode)
> 3. Wonder if even clicking the entire tag should remove it (not just the "X"), in which case it needs updated hover visuals as well. But that is something design should decide.
> 4. There's this thing where the "too many users" message reserves a good amount of space that results in funny UI. In the screen shot below there was enough visual space, still the latest item pushed the tag list into the scroll space.
> 
> Screenshot 2022-01-21 at 11 21 12
> 
> Maybe we can make the available space for tags as big as the input **until** the limit is reached?
> 
> 5. Darkmode needs some love, in the component as well as in the demo application
> 
> Screenshot 2022-01-21 at 11 23 31
> 
> Great work so far!!

2,4,5 fixed, 1 and 3 I would check with designers since nothing is specified in Figma.

PascalPrecht:
Thanks @alexandraB99 !
It seems the icon button used for the rendering the close icon still needs some tweaking. Below you can see a screenshot of both figma and sandbox. Notice how the "X" seems to be too big, and I believe the `leftMargin` needs to be smaller too.

In addition, notice how the "Tracy" tag doesn't have an "X"? The icon seems to disappear as soon as you hover the tag...

Screenshot 2022-01-24 at 15 56 44

This breaks in both themes.

> 2,4,5 fixed, 1 and 3 I would check with designers since nothing is specified in Figma.

Sounds good!

alexandraB99:
@PascalPrecht, I updated the spaces and sizes, regarding the color is the hover state default color from the `StatusIconTabButton` component, not sure what the hovered state of this component should be as it's not defined in Figma.

@John-44 button fixed. 
no_contact
contact

A few questions:

- Should users be able to delete items from the list by pressing backspace?
- Wonder if even clicking the entire tag should remove it (not just the "X"), in which case it needs updated hover visuals as well. But that is something design should decide.???
- What is the hover state of the name tag delegate?

John-44:
> A few questions:
> 
> * Should users be able to delete items from the list by pressing backspace?
> * Wonder if even clicking the entire tag should remove it (not just the "X"), in which case it needs updated hover visuals as well. But that is something design should decide.???
> * What is the hover state of the name tag delegate?

@alexandraB99 good questions! :-)

- Yes users should be able to delete items by pressing backspace (thanks for bringing this up)
- Re. clicking the entire tag should remove it: good idea, I don't see any reason not to do this because the tag has no other on-click action, and it would make touch interaction easier.  +1, good idea :-)

alexandraB99:
> > A few questions:
> > 
> > * Should users be able to delete items from the list by pressing backspace?
> > * Wonder if even clicking the entire tag should remove it (not just the "X"), in which case it needs updated hover visuals as well. But that is something design should decide.???
> > * What is the hover state of the name tag delegate?
> 
> @alexandraB99 good questions! :-)
> 
> * Yes users should be able to delete items by pressing backspace (thanks for bringing this up)
> * Re. clicking the entire tag should remove it: good idea, I don't see any reason not to do this because the tag has no other on-click action, and it would make touch interaction easier.  +1, good idea :-)

thanks for the quick answer @John-44 , all these was @PascalPrecht's idea btw :) I have added both requirements, the only thing missing is the hovered state of the name tag, what would be that?

John-44:
> the only thing missing is the hovered state of the name tag, what would be that?

@alexandraB99 For now let's go with something like this:
Screenshot_2022-02-01_at_14 57 14

We will want to update this to something different later when the design system is more established (e.g. when David is at the point of defining all the button behaviours), but for now we can use this so development isn't blocked

alexandraB99:
@John-44 thanks a lot, done
hovered


Nevertheless, fyi, we shall wait for this PR https://github.com/status-im/StatusQ/pull/546 to get merged first before merging this one

PascalPrecht:
@John-44 If we can't get an opinion about this from design, I'd rather have us stay consistent and use the same behaviour (color-wise) we have in other buttons in our design system.

This button variant here:
Screenshot 2022-02-01 at 20 50 39

Looks like this in dark mode:
Screenshot 2022-02-01 at 20 51 03

And the hovers look like this:
Screenshot 2022-02-01 at 20 51 13
Screenshot 2022-02-01 at 20 51 29

So instead of introducing this "red" color which, honestly, doesn't go well with the blue, how about we only change the color of the entire button and just in the same way as the one above? That way it's consistent and we don't have this funny new thing :D 

Wdyt?

John-44:
> And the hovers look like this: Screenshot 2022-02-01 at 20 51 13 Screenshot 2022-02-01 at 20 51 29
> 
> So instead of introducing this "red" color which, honestly, doesn't go well with the blue, how about we only change the color of the entire button and just in the same way as the one above? That way it's consistent and we don't have this funny new thing :D
> 
> Wdyt?

@PascalPrecht yes what you are suggesting above is much better, +1 go with that :-)  

cc'ing @alexandraB99

alexandraB99:
Ok @John-44 , @PascalPrecht I have changed the hovered to state as per @PascalPrecht's proposal. I have also moved StatusTagSelector to components as it makes more sense for it to be a component than a control. I also added the corresponding page in the API Documentation section.
normal
hovered
tagselectorpage

PascalPrecht (approved):

PascalPrecht:
Amazing job @alexandraB99 !
This looks very good now.

I've rebased it for you and landed it as https://github.com/status-im/StatusQ/commit/094dee492825520b47d3f4941a6720020dcbe6d3

PR #539, opened 2022-01-24
feat(StatusMemberListItem): Implement `StatusMemberListItem`

### LINKED ISSUE #515: Implement `StatusMemberListItem`
relevant spec: https://github.com/status-im/feature-specs/pull/42/files
relevant figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=932%3A284701

new member component that includes online/offline indicator, trust indictor, optional chat key, optional original username

Screen Shot 2022-01-12 at 2 43 39 PM

noeliaSD:
### What does the PR do

Create new member component that includes online/offline indicator, trust indictor and optional chat key and original username.

To do that, a new component `StatusMemberListItem` has been created and included in `sandbox\controls\ListItems` for testing all its variants.

Extend existing components `StatusListItem` and `StatusSmartIndenticon`:

1. `StatusListItem` component has now the possibility of adding 2 icons after the title row.
2. `StatusSmartIndenticon` has now the possibility of displaying online/offline indicator (or a most generic state indicator).

Closes #515

### Affected areas

Components/StatusMemberListItem [NEW]
Components/StatusListItem [MOD]
Components/StatusSmartIdenticon [MOD]

image

### DIFF
--- a/src/StatusQ/Components/StatusListItem.qml
+++ b/src/StatusQ/Components/StatusListItem.qml
@@ -9,43 +9,21 @@ import StatusQ.Controls 0.1
 Rectangle {
     id: statusListItem
 
-    implicitWidth: 448
-    implicitHeight: Math.max(64, statusListItemTitleArea.height + 16)
-
-    enum Type {
-        Primary,
-        Secondary,
-        Danger
-    }
-
-    color: {
-        if (sensor.containsMouse) {
-            switch(type) {
-                case StatusListItem.Type.Primary:
-                    return Theme.palette.baseColor2
-                case StatusListItem.Type.Secondary:
-                    return Theme.palette.statusListItem.secondaryHoverBackgroundColor
-                case StatusListItem.Type.Danger:
-                    return Theme.palette.dangerColor3
-            }
-        }
-        return Theme.palette.statusListItem.backgroundColor
-    }
-
-    radius: 8
-
     property string itemId: ""
     property string titleId: ""
-
     property string title: ""
     property string titleAsideText: ""
+    property bool titleIcon1Visible
+    property bool titleIcon2Visible
     property string subTitle: ""
-    property string tertiaryTitle: ""
-    property alias badge: statusListItemBadge
-
+    property string tertiaryTitle: ""    
+    property string label: ""
     property real leftPadding: 16
     property real rightPadding: 16
     property bool enabled: true
+    property int type: StatusListItem.Type.Primary
+    property list<Item> components
+
     property StatusIconSettings icon: StatusIconSettings {
         height: isLetterIdenticon ? 40 : 20
         width: isLetterIdenticon ? 40 : 20
@@ -73,23 +51,68 @@ Rectangle {
         height: 40
         isIdenticon: false
     }
-    property string label: ""
-
-    property int type: StatusListItem.Type.Primary
+    property StatusIconSettings titleIcon1: StatusIconSettings {
+        width: dummyImage.width
+        height: dummyImage.height
+        background: StatusIconBackgroundSettings {
+            width: 10
+            height: 10
+        }
+        // Only used to get implicit width and height from the actual image
+        property Image dummyImage: Image {
+            source: titleIcon1.name ? "../../assets/img/icons/" + titleIcon1.name + ".svg": ""
+            visible: false
+        }        
+    }
+    property StatusIconSettings titleIcon2: StatusIconSettings {
+        width: dummyImage.width
+        height: dummyImage.height
+        background: StatusIconBackgroundSettings {
+            width: 10
+            height: 10
+        }
+        // Only used to get implicit width and height from the actual image
+        property Image dummyImage: Image {
+            source: titleIcon2.name ? "../../assets/img/icons/" + titleIcon2.name + ".svg": ""
+            visible: false
+        }
+    }
 
     property alias sensor: sensor
-
+    property alias badge: statusListItemBadge
     property alias statusListItemIcon: iconOrImage
     property alias statusListItemTitle: statusListItemTitle
     property alias statusListItemTitleAside: statusListItemTitleAsideText
+    property alias statusListItemTitleArea: statusListItemTitleArea
     property alias statusListItemSubTitle: statusListItemSubTitle
     property alias statusListItemTertiaryTitle: statusListItemTertiaryTitle
     property alias statusListItemComponentsSlot: statusListItemComponentsSlot
 
-    property list<Item> components
-
     signal clicked(string itemId)
     signal titleClicked(string titleId)
+
+    enum Type {
+        Primary,
+        Secondary,
+        Danger
+    }
+
+    implicitWidth: 448
+    implicitHeight: Math.max(64, statusListItemTitleArea.height + 16)
+    color: {
+        if (sensor.containsMouse) {
+            switch(type) {
+                case StatusListItem.Type.Primary:
+                    return Theme.palette.baseColor2
+                case StatusListItem.Type.Secondary:
+                    return Theme.palette.statusListItem.secondaryHoverBackgroundColor
+                case StatusListItem.Type.Danger:
+                    return Theme.palette.dangerColor3
+            }
+        }
+        return Theme.palette.statusListItem.backgroundColor
+    }
+    radius: 8
 
     onComponentsChanged: {
         if (components.length) {
@@ -125,6 +148,7 @@ Rectangle {
             active: statusListItem.icon.isLetterIdenticon ||
                     !!statusListItem.icon.name ||
                     !!statusListItem.image.source.toString()
+            badge.border.color: statusListItem.color
         }
 
         Item {
@@ -144,7 +168,8 @@ Rectangle {
                 height: visible ? contentHeight : 0
                 wrapMode: Text.WrapAtWordBoundaryOrAnywhere
                 anchors.left: parent.left
-                anchors.right: !statusListItem.titleAsideText ? parent.right : undefined
+                anchors.right: !statusListItem.titleAsideText && !statusListItem.titleIcon1Visible && !statusListItem.titleIcon2Visible
+                               ? parent.right : undefined
                 color: {
                   if (!statusListItem.enabled) {
                     return Theme.palette.baseColor1
@@ -181,6 +206,38 @@ Rectangle {
                 color: Theme.palette.baseColor1
                 visible: !!statusListItem.titleAsideText
             }
+
+            Row {
+                id: titleIconsRow
+                spacing: 4
+                anchors.left: !statusListItem.titleAsideText ? statusListItemTitle.right : statusListItemTitleAsideText.right
+                anchors.verticalCenter: statusListItemTitle.verticalCenter
+                anchors.leftMargin: titleIconsRow.spacing
+
+                StatusRoundIcon {
+                    visible: statusListItem.titleIcon1Visible
+                    icon.name: statusListItem.titleIcon1.name
+                    icon.width: statusListItem.titleIcon1.width
+                    icon.height: statusListItem.titleIcon1.height
+                    icon.rotation: statusListItem.titleIcon1.rotation
+                    icon.color: statusListItem.titleIcon1.color
+                    icon.background.color: statusListItem.titleIcon1.background.color
+                    icon.background.width: statusListItem.titleIcon1.background.width
+                    icon.background.height: statusListItem.titleIcon1.background.height
+                }
+
+                StatusRoundIcon {
+                    visible: statusListItem.titleIcon2Visible
+                    icon.name: statusListItem.titleIcon2.name
+                    icon.width: statusListItem.titleIcon2.width
+                    icon.height: statusListItem.titleIcon2.height
+                    icon.rotation: statusListItem.titleIcon2.rotation
+                    icon.color: statusListItem.titleIcon2.color
+                    icon.background.color: statusListItem.titleIcon2.background.color
+                    icon.background.width: statusListItem.titleIcon2.background.width
+                    icon.background.height: statusListItem.titleIcon2.background.height
+                }
+             }
 
             StatusBaseText {
                 id: statusListItemSubTitle
--- a/src/StatusQ/Components/StatusMemberListItem.qml
+++ b/src/StatusQ/Components/StatusMemberListItem.qml
@@ -1,0 +1,76 @@
+import QtQuick 2.0
+import StatusQ.Core.Theme 0.1
+import StatusQ.Core 0.1
+
+
+StatusListItem {
+    id: root
+
+    property string nickName: ""
+    property string userName: ""
+    property string chatKey: ""
+    property bool isMutualContact: false
+    property var trustIndicator: StatusMemberListItem.TrustedType.None
+    property bool isOnline: false
+
+    enum TrustedType {
+        None, //0
+        Verified, //1
+        Untrustworthy //2
+    }
+
+    // Subtitle composition:
+    function composeSubtitile() {
+        var compose = ""
+        if(root.userName !== "")
+            compose = "(" + root.userName + ")"
+
+        if(compose !== "" && root.chatKey !== "")
+            // Composition
+            compose += " • " + composeShortKeyChat(root.chatKey)
+
+        else if(root.chatKey !== "")
+            compose = composeShortKeyChat(root.chatKey)
+
+        return compose
+    }
+
+    // Short keychat composition:
+    function composeShortKeyChat(chatKey) {
+        return chatKey.substring(0, 5) + "..." + chatKey.substring(chatKey.length - 3)
+    }
+
+    // root object settings:
+    title: root.nickName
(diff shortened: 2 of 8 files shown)


src/StatusQ/Components/StatusListItem.qml
```diff
@@ -144,7 +169,7 @@ Rectangle {
                 height: visible ? contentHeight : 0
                 wrapMode: Text.WrapAtWordBoundaryOrAnywhere
                 anchors.left: parent.left
-                anchors.right: !statusListItem.titleAsideText ? parent.right : undefined
+                anchors.right: !statusListItem.titleAsideText && !titleIconsRow.visible ? parent.right : undefined
```
noeliaSD, line 172: Here there is an issue that I propose to address in a separate task as it afects `StatusListItem` component and I've seen the same issue in `StatusMessageHeader.qml` component as when the `anchors.right` property is set as `undefined` neither `titleIconsRow` nor `titleAsideText` are correctly positioned and the `statusListItemTitle` comes out of the expected boundaries.

src/StatusQ/Components/StatusSmartIdenticon.qml
```diff
@@ -6,6 +6,11 @@ Loader {
     id: statusSmartIdenticon
 
     property string name: ""
+    property bool isBadgeVisible: false
+
+    // Badge color properties must be set if isBadgeVisible = true
+    property color badgeColor
```
noeliaSD, line 12: Here we could expose the complete `badge` component to have more flexibility. What do you think is it a better option @PascalPrecht @Khushboo-dev-cpp ?

src/StatusQ/Components/StatusMemberListItem.qml
```diff
@@ -0,0 +1,64 @@
...
+
+    // root object settings:
+    title:  root.nickName
+    titleIcon1Visible: root.isMutualContact
+    titleIcon2Visible: root.trustIndicator !== StatusMemberListItem.TrustedType.None
+    subTitle: printSubtitile()
+    subTitlePixelSize: 10
+    icon.isLetterIdenticon: root.imageSource === ""
+    image.source: root.imageSource
+    image.isIdenticon: root.imageSource !== ""
```
PascalPrecht, line 52: Hm.. so this doesn't look right to me. Why do we assume that we're dealing with an identicon, when `imageSource != ""` ?

Keep in mind that a normal profile picture is also just an image source.

So I guess what needs to happen here instead is that we remove the `imageSource` introduced n `StatusMemberListItem` and simply rely on the already existing `image.[...]` properties which allow consumers to set `isIdenticon` etc respectively.

Am I missing something here otherwise?

src/StatusQ/Components/StatusMemberListItem.qml
```diff
@@ -0,0 +1,64 @@
...
+    statusListItemIcon.isBadgeVisible: true
+    statusListItemIcon.badgeColor: root.isOnline ? Theme.palette.successColor1 : Theme.palette.baseColor1
+
+    // Trusted type icons definition:
+    titleIcon1.name: "tiny/tiny-contact"
+    titleIcon1.color: Theme.palette.indirectColor1
+    titleIcon1.background.color: Theme.palette.primaryColor1
+    // None and Untrustworthy types, same aspect (Icon will not be visible in case of None type):
+    titleIcon2.name: trustIndicator === StatusMemberListItem.TrustedType.Verified ? "tiny/tiny-checkmark" : "tiny/subtract"
+    titleIcon2.color: Theme.palette.indirectColor1//trustIndicator === StatusMemberListItem.TrustedType.Verified ? Theme.palette.indirectColor1 : Theme.palette.directColor1
```
PascalPrecht, line 62: I think this is a leftover?

src/StatusQ/Components/StatusSmartIdenticon.qml
PascalPrecht: Yes, I think that's a good idea. Rather introduce the entire badge so we don't need to reintroduce all the additional properties just to propagate them.

PascalPrecht (changes requested):
Nice work @noeliaSD !

I've left some inline comments that probably need to be addressed.

Also, since we now finally get such a component, would you mind making use of it in the demo application? We have a member list there as well that uses some rather rudimentary implementation at the moment.

Oh, and... should the member list item not be smaller? Based on what I see in sandbox, it seems to have the same default size as the StatusListItem. I don't think they should be that big default no?

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -73,24 +52,69 @@ Rectangle {
...
+                    return Theme.palette.baseColor2
+                case StatusListItem.Type.Secondary:
+                    return Theme.palette.statusListItem.secondaryHoverBackgroundColor
+                case StatusListItem.Type.Danger:
+                    return Theme.palette.dangerColor3
+            }
+        }
+        return Theme.palette.statusListItem.backgroundColor
+    }
+    radius: 8
```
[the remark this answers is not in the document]
noeliaSD: Yeah! You are right. Next time I'll do it separately. I've just only organized it to follow the QML coding convention in order to make the file more readable.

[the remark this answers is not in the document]
noeliaSD: Yeah! You are right. Next time I'll do it separately. I've just only organized it to follow the QML coding convention in order to make the file more readable.

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -73,24 +52,69 @@ Rectangle {
...
+        height: dummyImage.height
+        background: StatusIconBackgroundSettings {
+            width: 10
+            height: 10
+        }
+        // Only used to get implicit width and height from the actual image
+        property Image dummyImage: Image {
+            source: titleIcon1.name ? "../../assets/img/icons/" + titleIcon1.name + ".svg": ""
+            visible: false
+        }
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: I needed to use this in the StatusMessage header as well. that's because the image size is rather small and width and height are not the same, and because they are so tiny, setting a default one really breaks the way to image looks. Not sure if Noelia has a better way :)

src/StatusQ/Components/StatusMemberListItem.qml
```diff
@@ -0,0 +1,64 @@
...
+
+    // Subtitle composition:
+    function printSubtitile() {
+        var print = ""
+        if(root.userName !== "")
+            print = "(" + root.userName + ")"
+
+        if(print !== "" && root.keyChat !== "")
+            // Composition
+            print += " • " + printShortKeyChat(root.keyChat)
```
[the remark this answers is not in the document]
Khushboo-dev-cpp: Can we rename this? Maybe I not used to the print naming, but it could be rather compose SubTitile? anyway, just my feeling.

src/StatusQ/Components/StatusMemberListItem.qml
```diff
@@ -0,0 +1,64 @@
...
+
+    property string nickName: ""
+    property string userName: ""
+    property string keyChat: ""
+    property bool isMutualContact: false
+    property var trustIndicator: StatusMemberListItem.TrustedType.None
+    property string imageSource: ""
+    property bool isOnline: false
+
+    enum TrustedType {
```
Khushboo-dev-cpp, line 17: I think in StatusMessage we use Enum ContactType? Can we use similar enum so at to keep consistency when using this? You can also change it in StatusMessage, whatever you prefer.

Khushboo-dev-cpp (commented):
Hey some minor things from me as well :) Looks good overall !

src/StatusQ/Components/StatusListItem.qml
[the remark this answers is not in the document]
noeliaSD: I agree with you that this is not the cleanest approach and, yes, we could approach it by using for example a fixed size of 6x8 (as in this case is the size of these tiny specific icons) but you lose some flexibility as if you use the component with bigger icon sizes, it will not be rendered correctly (or at least is what I’ve experimented). Maybe this shouldn't be happening, and we should review our base class `StatusRoundIcon`. 

- For example, if I fix it to 10x10:
10x10

- This is the dynamic size approach, in this case, same as 6x8:
6x8

Can you appreciate the difference?

This approach is, as well, used in the `StatusMessageHeader.qml` (as @Khushboo-dev-cpp says) and gives some dynamic opportunities to the component.

src/StatusQ/Components/StatusMemberListItem.qml
noeliaSD: I wanted to open another issue for that as I was wondering that maybe we could share this enum and icons / colors definition however, I've realized that in `StatusMessage` it is not used the same implementation as here and I believe that it is a misunderstood from one or the other side because I understand it should be the same (or it is my interpretation of the implementation in `StatusMessage`, please correct me if I'm wrong).

In `StatusMessage` there is a unique enum and a unique `contentType` property in message details but you can have 2 different icons as for example, the **mutual contact** and the **verified**. So in the current `StatusMessage` implementation only one of them can be displayed at the same time, isn't it?

src/StatusQ/Components/StatusListItem.qml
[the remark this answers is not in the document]
PascalPrecht: Wonderful, thanks for clarifying!

src/StatusQ/Components/StatusMemberListItem.qml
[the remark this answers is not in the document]
noeliaSD: > Just a thought here: we also have `StatusAddress` which can be used to render an address. Although it might make things unnecessarily complex here.
> 
> Have a look if you want, you call!

Nice component `StatusAddress`! Thanks for sharing it @PascalPrecht!

Next time I’ll consider it for my implementation. In this case, I should redesign the component itself to use it so I’ll keep it as it is, if you don’t mind.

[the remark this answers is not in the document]
noeliaSD: > Can we rename this? Maybe I not used to the print naming, but it could be rather compose SubTitile? anyway, just my feeling.

Sure! I'll rename it!

noeliaSD:
> Also, since we now finally get such a component, would you mind making use of it in the demo application? We have a member list there as well that uses some rather rudimentary implementation at the moment.

noeliaSD:
> Also, since we now finally get such a component, would you mind making use of it in the demo application? We have a member list there as well that uses some rather rudimentary implementation at the moment.

Could you tell me where do you have this member list you mean exactly?

noeliaSD:
Oh! I accidentally closed it!!

src/StatusQ/Components/StatusMemberListItem.qml
noeliaSD: Ok, so I'll only set the following to always have, at least, a letter as default:
```icon.isLetterIdenticon: !root.image.source.toString()```

**Comment by noeliaSD** [comment]
> Oh, and... should the member list item not be smaller? Based on what I see in sandbox, it seems to have the same default size as the StatusListItem. I don't think they should be that big default no?

Sure! I'll modified it according to the design! Thanks

**Comment by noeliaSD** [comment]
@PascalPrecht and @Khushboo-dev-cpp, I've pushed an update with the requested changes. Please check them again and let me know what you think. Thanks!

File: src/StatusQ/Components/StatusListItem.qml (hunk 1 above)
Comment (noeliaSD) [reply src/StatusQ/Components/StatusListItem.qml:170]: Here the created issue [#540](https://app.zenhub.com/workspaces/status-desktop-5ea70d29c12ee928df5ae848/issues/status-im/statusq/540)

File: src/StatusQ/Components/StatusMemberListItem.qml (hunk 8 above)
Comment (noeliaSD) [reply src/StatusQ/Components/StatusMemberListItem.qml:16]: Created new task for that: [#541](https://github.com/status-im/StatusQ/issues/541)

File: src/StatusQ/Components/StatusMemberListItem.qml (hunk 3 above)
Comment (PascalPrecht) [reply src/StatusQ/Components/StatusMemberListItem.qml]: One possibility although, we might as well default it to `false` because I don't think members will ever have a letter identicon. Those are used for channels in communities.

I think the default of `isLetterIdenticon` is even `false`. If not, you can default it to `false` inside of this component.

File: src/StatusQ/Components/StatusSmartIdenticon.qml [hunk 9]
```diff
@@ -7,6 +7,9 @@ Loader {
 
     property string name: ""
 
+    // Badge color properties must be set if badgeItem.visible = true
+    property alias statusSmartIdenticonBadge: statusBadge
```
PascalPrecht, line 11: Can just call this `badge` instead. We already know we're in a smart identicon

src/StatusQ/Components/StatusMemberListItem.qml
```diff
@@ -0,0 +1,74 @@
...
+    statusListItemIcon.statusSmartIdenticonBadge.color: root.isOnline ? Theme.palette.successColor1 : Theme.palette.baseColor1
+
+    // Default sizes/positions by design
+    implicitWidth: 256
+    leftPadding: 8
+    image.width: 32
+    image.height: 32
+    icon.width: 32
+    icon.height: 32
+    statusListItemIcon.anchors.verticalCenter: sensor.verticalCenter
```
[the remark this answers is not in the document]
PascalPrecht: That's a very valid point and one that we've considered a few times in the past (until we eventually ended up occasionally exposing/aliasing entire components).

I guess at the end of the day it's a matter of balance. I fully agree that generally you don't want to expose too much control to consumers, especially if they could break the components. 

But then, I guess, you also expect it to be of the consumer's interest to use the component correctly. As for maintenance, I'd argue it's actually more work to maintain if we introduce individual properties that already exist on some object.

For now I think I'd vote for just going with exposing the entire thing and trusting users they won't mess it up and also for consistency's sake. 

Before we cut a 1.0 release, we can revisit these points to finalize APIs.

src/StatusQ/Components/StatusMemberListItem.qml
noeliaSD: I don't think it should be false as default. I understand from the design that each member in the list should have or an image/identicon or a letter identifying him/her. If it is set to false, the StatusMemberList item didn't display the circle with the member identification and would be only its nickName, userName and/or chatKey, and what would happend with the connect/disconnect indicator? 

From the design, the possibilities I understand are, at least, showing a letter: 
image

PascalPrecht: > I understand from the design that each member in the list should have or an image/identicon or a letter identifying him/her.

Ah, this is new and wasn't the case before. Users could only have either an identicon or a profile picture. I see letter identicons being used for accounts the first time here.

So then, I'd agree with your and probably would even say that it should default to `true` **however**, I wonder how this is done on a model/data basis. Right now every account has an `imageSource` because every account has an identicon by default. 

So if we go down that road, we probably need to discuss (with backend possibly) how we detect that indeed a letteridenticon should be rendered and not the normal identicon

src/StatusQ/Components/StatusSmartIdenticon.qml
noeliaSD: I agree! Done!

src/StatusQ/Components/StatusMemberListItem.qml
noeliaSD: > So if we go down that road, we probably need to discuss (with backend possibly) how we detect that indeed a letteridenticon should be rendered and not the normal identicon

Yes, maybe a discussion with the backend is needed or there is already a task to do it. Think we are now doing just the qml components while we wait for `base_bc` to be integrated to master.

noeliaSD: > So then, I'd agree with you and probably would even say that it should default to `true` **however**, I wonder how this is done on a model/data basis. Right now every account has an `imageSource` because every account has an identicon by default.

We cannot set it true as default because of the way the `StatusSmartIdenticon` is implemented as setting it to true will force the letter option always. 

See the piece of code where `StatusSmartIdenticon` decides about the **sourceComponent** to be loaded:

```
sourceComponent: statusSmartIdenticon.icon.isLetterIdenticon ? letterIdenticon :
                 !!statusSmartIdenticon.image.source.toString() ? roundedImage :
                 !!statusSmartIdenticon.icon.name.toString() ? roundedIcon : letterIdenticon```

PascalPrecht: Ah good find! That let's stick with the expression you've proposed. We might want to adjust `StatusSmartIdenticon` in the future to be a bit more flexible on that regard

noeliaSD:
> Also, since we now finally get such a component, would you mind making use of it in the demo application? We have a member list there as well that uses some rather rudimentary implementation at the moment.

Here an screenshot:
image

PascalPrecht (approved):
Very nice work @noeliaSD !

This seems to be ready for merge. There's still a conflict though, would you mind updating the PR?

PR #545, opened 2022-01-31
fix: Icon components had wrong color in sandbox app

### LINKED ISSUE #538: Icon components lost their color in sandbox application
Seems like this is a regression introduced in https://github.com/status-im/StatusQ/commit/0056df517d33f44dbf7ec1e07c116f48159c7ed1

The icons in the icons component page (sandbox) have lost their color property. This needs to be restored.

alexandraB99:
Icon color in Icons page was broken as setting it
in main.qml was mistakenly removed. Set it back to
Theme.palette.primaryColor1

Closes #538
wtheme
dtheme

### DIFF
--- a/sandbox/main.qml
+++ b/sandbox/main.qml
@@ -273,7 +273,12 @@ StatusWindow {
                         Loader {
                             id: viewLoader
                             anchors.centerIn: parent
-                            source: control("Icons")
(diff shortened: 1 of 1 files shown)


jrainville (approved):

PascalPrecht:
Nice fix @alexandraB99 !!

PR #546, opened 2022-02-01
Chore/restore 0.22.0 release

alexandraB99:


alexandraB99:
@noeliaSD I have updated Models.qml as well as DemoApp and main.qml, should be fine now. Please have a look and let me know if you observed anything else, if not we could squash the commits and merge this PR. Thanks for initiating this!

PascalPrecht (changes requested):
Cool, this seems to fix #547 - thanks for that!!

Would be wonderful if we could squash these commits into one, following the commit message conventions.

noeliaSD (approved):
Squashed commits and ready to merge! Great job @alexandraB99, thank you!!

@PascalPrecht if you don't have any objection, we can now merge it.

PascalPrecht (approved):
Great great work you two!

PR #548, opened 2022-02-01
feat(StatusQ.Components): Adding StatusToastMessage

### LINKED ISSUE #521: Implement `StatusToastMessage` component
figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=2090%3A252386
figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=1838%3A360271
figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=1939%3A218046

Screen Shot 2022-01-12 at 3 19 20 PM

alexandraB99:
Added StatusToastMessage component, an example page
as well as a demonstration in chat view

Closes #521

https://user-images.githubusercontent.com/31625338/152238220-b533f300-8da8-49a0-a262-125ed7b990da.mov

### DIFF
--- a/src/StatusQ/Components/StatusToastMessage.qml
+++ b/src/StatusQ/Components/StatusToastMessage.qml
@@ -1,0 +1,179 @@
+import QtQuick 2.14
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+
+import QtGraphicalEffects 1.13
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+
+Control {
+    id: root
+    width: 343
+    height: !!secondaryText ? 68 : 48
+    anchors.right: parent.right
+
+    property bool open: false
+    property string primaryText: ""
+    property string secondaryText: ""
+    property bool loading: false
+    property string iconName: ""
+    property string linkUrl: ""
+
+    property StatusIconSettings icon: StatusIconSettings {
+        width: 23
+        height: 23
+    }
+
+    property int type: StatusToastMessage.Type.Default
+    enum Type {
+        Default,
+        Success
+    }
+
+    function open(title, subTitle, iconName, type, loading, url) {
+        root.primaryText = title;
+        root.secondaryText = subTitle;
+        root.icon.name = iconName;
+        root.type = type;
+        root.loading = loading;
+        root.linkUrl = url;
+        root.open = true;
+    }
+
+    signal close()
+    signal linkActivated(var link)
+
+    states: [
+        State {
+            name: "opened"
+            when: root.open
+            PropertyChanges {
+                target: root
+                anchors.rightMargin: 0
+                opacity: 1.0
+            }
+        },
+        State {
+            name: "closed"
+            when: !root.open
+            PropertyChanges {
+                target: root
+                anchors.rightMargin: -width
+                opacity: 0.0
+            }
+            StateChangeScript {
+                script: { root.close(); }
+            }
+        }
+    ]
+
+    transitions: [
+        Transition {
+            to: "*"
+            NumberAnimation { properties: "anchors.rightMargin,opacity"; duration: 400 }
+        }
+    ]
+
+    background: Rectangle {
+        id: background
+        anchors.fill: parent
+        color: Theme.palette.statusToastMessage.backgroundColor
+        radius: 8
+        border.color: Theme.palette.baseColor2
+        layer.enabled: true
+        layer.effect: DropShadow {
+            verticalOffset: 3
+            radius: 8
+            samples: 15
+            fast: true
+            cached: true
+            color: Theme.palette.dropShadow
+        }
+    }
+
+    contentItem: Item {
+        anchors.left: parent.left
+        anchors.leftMargin: 16
+        anchors.right: parent.right
+        anchors.rightMargin: 4
+        height: parent.height
+        MouseArea {
+            anchors.fill: parent
+            onMouseXChanged: {
+                root.open = (mouseX < (root.width/3));
+            }
+        }
+        RowLayout {
+            anchors.fill: parent
+            spacing: 16
+            Rectangle {
+                implicitWidth: 32
+                implicitHeight: 32
+                Layout.alignment: Qt.AlignVCenter
+                radius: (root.width/2)
+                color: (root.type === StatusToastMessage.Type.Success) ?
+                        Theme.palette.successColor2 : Theme.palette.primaryColor3
+                Loader {
+                    anchors.centerIn: parent
+                    sourceComponent: root.loading ? loadingInd : statusIcon
+                    Component {
+                        id: loadingInd
+                        StatusLoadingIndicator {
+                            color: (root.type === StatusToastMessage.Type.Success) ?
+                                   Theme.palette.successColor1 : Theme.palette.primaryColor1
+                        }
+                    }
+                    Component {
+                        id: statusIcon
+                        StatusIcon {
+                            anchors.centerIn: parent
+                            width: root.icon.width
+                            height: root.icon.height
+                            color: (root.type === StatusToastMessage.Type.Success) ?
+                                   Theme.palette.successColor1 : Theme.palette.primaryColor1
+                            icon: root.icon.name
+                        }
+                    }
+                }
+            }
+            Column {
+                Layout.fillWidth: true
+                Layout.alignment: Qt.AlignVCenter
+                StatusBaseText {
+                    width: parent.width
+                    font.pixelSize: 13
+                    color: Theme.palette.directColor1
+                    elide: Text.ElideRight
+                    text: root.primaryText
+                }
+                StatusBaseText {
+                    width: parent.width
+                    visible: (!root.linkUrl && !!root.secondaryText)
+                    height: visible ? contentHeight : 0
+                    font.pixelSize: 13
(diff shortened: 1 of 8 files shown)


src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,110 @@
...
+        root.secondaryText = subTitle;
+        root.open = true;
+    }
+
+    signal close()
+
+    background: Rectangle {
+        id: background
+        anchors.fill: parent
+        color: Theme.palette.statusSelect.menuItemBackgroundColor
```
PascalPrecht, line 38: Since this is not a select component, we shouldn't use that palette color.
Let's rather introduce a `statusToastMessage` namespace for that.

src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,110 @@
...
+        }
+        RowLayout {
+            anchors.fill: parent
+            spacing: 16
+            Rectangle {
+                implicitWidth: 32
+                implicitHeight: 32
+                Layout.alignment: Qt.AlignVCenter
+                radius: (width/2)
+                color: "#4EBC60"
```
PascalPrecht, line 85: Not sure what this is about, but if this is some sort of default color for this component, it should live in n`Theme`.

PascalPrecht (commented):
Very nice and clean @alexandraB99 !
Just left some comments related to colors.

Will review again once the "don't" merge label has been removed.

alexandraB99:
@PascalPrecht took care of the comments, you can have a look now. I added also a page in the API documentation. That said I am not sure how far we should go with this task, since it's only about creating the component. The mechanism to show/hide notifications lacks specification and I assume it will also depend on how the backend will be designed (if it's not there already). So does it make sense to implement also the mechanism and show it in demoapp in addition to the component page, or it's sufficient to just demonstrate the component in API documentation for this task?

alexandraB99:
after discussion with @iurimatias, this PR fulfils the task. The open/close mechanism should be tackled in a different task.

src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,166 @@
...
+                    font.pixelSize: 13
+                    color: Theme.palette.baseColor1
+                    text: root.secondaryText
+                    elide: Text.ElideRight
+                }
+                StatusSelectableText {
+                    visible: (!!root.linkUrl)
+                    height: visible ? implicitHeight : 0
+                    font.pixelSize: 13
+                    hoveredLinkColor: Theme.palette.primaryColor1
```
PascalPrecht, line 150: Judging by the designs, this text should always have the link color, not only when hovered.

src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,166 @@
...
+    id: root
+    width: 343
+    height: !!secondaryText ? 68 : 48
+    anchors.right: parent.right
+
+    property bool open: false
+    property string primaryText: ""
+    property string secondaryText: ""
+    property bool loading: false
+    property string iconName: ""
```
PascalPrecht, line 20: I think it'd be nicer if we used `StatusIconSettings` here instead and then give it some decent defaults. This way, we stay consistent with other components that render icons.

src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,166 @@
...
+    property bool success: false
```
PascalPrecht, line 21: I'd expect that there's toast messages that are neither "success" nor "loading" but probably "failure".

How about we rather introduce an enum type, similar to many other components in StatusQ that can have these different states?

src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,166 @@
...
+                    height: visible ? implicitHeight : 0
+                    font.pixelSize: 13
+                    hoveredLinkColor: Theme.palette.primaryColor1
+                    text: "<p><a style=\"text-decoration:none\" href=\'" + root.linkUrl + " \'>" + root.secondaryText + "</a></p>"
+                    onLinkActivated: {
+                        root.linkActivated(root.linkUrl);
+                    }
+                }
+            }
+            StatusIconTabButton {
```
PascalPrecht, line 157: This should rather be a 

```qml
        StatusFlatRoundButton {
            ...
        }
```

PascalPrecht (changes requested):
> you can have a look now. I added also a page in the API documentation.

Wonderful @alexandraB99 !

I've found a few more things that need adjustment, but nothing major.

Re: adding it to demo app: 
Generally, API documentation page is a must. Demo app scenario is a nice to have.

Once the outstanding things are addressed, this is gtg!

src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,166 @@
...
+import QtGraphicalEffects 1.13
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+
+Control {
+    id: root
+    width: 343
+    height: !!secondaryText ? 68 : 48
+    anchors.right: parent.right
```
jrainville, line 14: Should the anchors be handled by the component that uses it instead?

jrainville (commented):
Looks pretty good to me. I'm not used to reviewing StatusQ components, so I'll trust Pascal's opinion on the matter

src/StatusQ/Components/StatusToastMessage.qml
alexandraB99: this is not clear to me as it's not really specified, only one state is shown so I can't judge. @[John-44](https://github.com/John-44) could you maybe clarify?

src/StatusQ/Components/StatusToastMessage.qml
alexandraB99: this is to handle the animation for show/hide. Atm I don't have a view of how that should be working since very limited info is available on Figma. This should be re-visited when we know how toast messages should work exactly

src/StatusQ/Components/StatusToastMessage.qml
alexandraB99: I thought about this, but since is nothing specified yet, I don't know whether it will be only the icon name needed, more or nothing (maybe we have fixed icons). I'd avoid adding complexity if not needed. nevertheless this is something we can always add later when more details are available.

src/StatusQ/Components/StatusToastMessage.qml
PascalPrecht: If we keep it like it is now, you basically don't have an indicator that this text is clickable in the first place and it looks like the other case where there's indeed only info text. 

I assume that's why the "View on Etherscan" text has the primary color (blue) by default. I'd also then re-add the text-decoration: underline on hover to get the hover indicator back.

src/StatusQ/Components/StatusToastMessage.qml
PascalPrecht: I understand your reasoning. Just be aware that, if we do need to change it afterwards, it will introduce a breaking change (which I'd avoid as much as possible) unless we keep the `iconName` property around.

I'll let this be your call.

src/StatusQ/Components/StatusToastMessage.qml
PascalPrecht: Same as with the icon settings: We can add it later, but it would require us to change the API so this will be a breaking change.

> I thought about this, but since is nothing specified yet I don't know whether it will be only 2 types or more

We already have "failure" toast messages in Status Desktop today (for failing transactions). The new design just seems to be incomplete on that regard. I guess we want to preserve existing functionality.

alexandraB99:
@PascalPrecht it's not that I disagree, but that I can't decide whether type and icon settings are needed or not since a lot of details are still missing. Anyway, I took care of all comments so you could have a second look. When we have more details we can update accordingly.

src/StatusQ/Components/StatusToastMessage.qml
```diff
@@ -0,0 +1,179 @@
...
+    property int type: StatusToastMessage.Type.Default
+    enum Type {
+        Default,
+        Success
+    }
+
+    function open(title, subTitle, iconName, type, loading, url) {
+        root.primaryText = title;
+        root.secondaryText = subTitle;
+        root.iconName = iconName;
```
PascalPrecht, line 37: If you went for `StatusIconSettings`, this should probably now be `root.icon.name = iconName`

alexandraB99: you are right, updated!

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/7b290ddd56bb228c5341147699246a0fe7680b70

@alexandraB99 I made a few smaller changes:

1. The close button didn't have the correct variation (needed to be a secondary type variant)
2. On hover, the close button size was way too big, so I've updated as well to match what we have in the rest of the design system

Here's a screenshot:
Screenshot 2022-02-08 at 09 36 24

Massive work! Happy to see this has landed :)

Issue #499, opened 2021-12-14
Introduce message component(s)


PascalPrecht:
This issue serves a ground to discuss the new API(s) for message components provided by StatusQ. 
The discussion has originally started in https://github.com/status-im/status-desktop/issues/4272.

Comments in the original issue have been echo'd here.

- [x] create skeleton with profile image and dummy model
- [x] https://github.com/status-im/StatusQ/issues/501
- [x] https://github.com/status-im/StatusQ/issues/502
- [x] https://github.com/status-im/StatusQ/issues/505
- [x] https://github.com/status-im/StatusQ/issues/506
- [x] https://github.com/status-im/StatusQ/issues/504
- [x] https://github.com/status-im/StatusQ/issues/507
- [x] https://github.com/status-im/StatusQ/issues/508
- [x] https://github.com/status-im/StatusQ/issues/511
- [x] https://github.com/status-im/StatusQ/issues/512
- [ ] invitation bubble
- [ ] links
- [ ] transaction
- [ ] time info ?
- [ ] Audio Message new design with audio graph

PascalPrecht:
[Comment](https://github.com/status-im/status-desktop/issues/4272#issuecomment-992500744) by @Khushboo-dev-cpp 


This issue needs to be analysed and be divided into several subtasks -

API should be something like below

1. Sender Name
2. has mention
3. Time info
4. Profile Picture info
5. Pinned message
6. Message Content (Create separate Components for each of the below in StatusQ
  a. Text message
  b. Image message (open in full screen should be self contained and also the various actions of right click of button?)
  c. Sticker
  d. Gif
  e. Audio
7. Edit Mode
8. is a reply and the reply component itself.
9. Quick Actions button panel

PascalPrecht:
[Comment](https://github.com/status-im/status-desktop/issues/4272#issuecomment-992754961) by @Khushboo-dev-cpp 

Noting down here is an initial version of what the API for the Message component should look like

```qml
StatusMessage {
    messageFromUser: true
    senderName: "Sender of Message"
    senderProfilePicture: image||identicon
    contentType: StatusMessage.Type.Text ||
                 StatusMessage.Type.Image ||
                 StatusMessage.Type.Sticker ||
                 StatusMessage.Type.Audio
    messageText: "This is a message"
    imageMsgSource: "img"
    stickerData: sticker
    timeInfo: ""
    isMessagePinned: true
    pinnedBy: someone
    isAReply: true
[...]

PR #549, opened 2022-02-04
feat(StatusQ.Components): Adding StatusWizardSteper component

### LINKED ISSUE #522: Implement `StatusWizardStepper` component
figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=2459%3A328034

Screen Shot 2022-01-12 at 3 22 32 PM

alexandraB99:
Added StatusWizardStepper component

Also added corresponding page in API documentation

Closes #522

https://user-images.githubusercontent.com/31625338/152873883-601b29a3-3ac2-4aab-97a3-e245e0c484c3.mov

### DIFF
--- a/sandbox/pages/StatusWizardStepperPage.qml
+++ b/sandbox/pages/StatusWizardStepperPage.qml
@@ -1,0 +1,100 @@
+import QtQuick 2.14
+
+import StatusQ.Core 0.1
+import StatusQ.Components 0.1
+import StatusQ.Core.Theme 0.1
+
+Item {
+    id: root
+    width: 800
+    height: 100
+
+    //Simulate animation between steps
+    property bool stepsCompleted: false
+    function reset() {
+        for (var i = 0; i < stepsListModel.count; i++) {
+            stepsListModel.setProperty(i, "loading", false);
+            stepsListModel.setProperty(i, "stepCompleted", false);
+            step1.loadingTime = 0;
+            step2.loadingTime = 0;
+            root.stepsCompleted = false;
+        }
+    }
+
+    ListModel {
+        id: stepsListModel
+        ListElement {description:"Send Request"; loadingTime: 0; stepCompleted: false}
+        ListElement {description:"Receive Response"; loadingTime: 0; stepCompleted: false}
+        ListElement {description:"Confirm Identity"; loadingTime: 0; stepCompleted: false}
+    }
+
+    Shortcut {
+        sequence: "Ctrl+R"
+        onActivated: {
+            if (!root.stepsCompleted) {
+                animate.running = true;
+            } else {
+                root.reset();
+            }
+        }
+    }
+
+    StatusBaseText {
+        id: indicLabel
+        anchors.horizontalCenter: parent.horizontalCenter
+        color: Theme.palette.directColor1
+        text: "Press Ctrl+R to run the animation"
+        font.pixelSize: 17
+    }
+
+    SequentialAnimation {
+        id: animate
+        ScriptAction {
+            id: step1
+            property int loadingTime: 0
+            Behavior on loadingTime { NumberAnimation { duration: 2000 }}
+            onLoadingTimeChanged: {
+                stepsListModel.setProperty(1, "loadingTime", step1.loadingTime);
+            }
+            script: {
+                step1.loadingTime = 2000;
+                stepsListModel.setProperty(0, "loadingTime", step1.loadingTime);
+                stepsListModel.setProperty(0, "stepCompleted", true);
+            }
+        }
+        PauseAnimation {
+            duration: 2100
+        }
+        ScriptAction {
+            id: step2
+            property int loadingTime: 0
+            Behavior on loadingTime { NumberAnimation { duration: 2000 } }
+            onLoadingTimeChanged: {
+                stepsListModel.setProperty(2, "loadingTime", step2.loadingTime);
+            }
+            script: {
+                step2.loadingTime = 2000;
+                stepsListModel.setProperty(1, "stepCompleted", true);
+            }
+        }
+        PauseAnimation {
+            duration: 2100
+        }
+        ScriptAction {
+            script: {
+                stepsListModel.setProperty(2, "stepCompleted", true);
+                root.stepsCompleted = true;
+            }
+        }
+    }
+    //simulation code
+
+    StatusWizardStepper {
+        id: wizardStepper
+        width: (parent.width - 50)
+        anchors.top: indicLabel.bottom
+        anchors.topMargin: 16
+        anchors.horizontalCenter: parent.horizontalCenter
+        stepsModel: stepsListModel
+    }
+}
--- a/src/StatusQ/Components/StatusWizardStepper.qml
+++ b/src/StatusQ/Components/StatusWizardStepper.qml
@@ -1,0 +1,78 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.12
+import QtQuick.Controls 2.14
+
+import StatusQ.Core 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Core.Theme 0.1
+
+Item {
+    id: wizardWrapper
+    width: parent.width
+    height: 56
+    property ListModel stepsModel: ListModel { }
+
+    ListView {
+        id: repeat
+        width: childrenRect.width
+        height: parent.height
+        anchors.horizontalCenter: parent.horizontalCenter
+        orientation: ListView.Horizontal
+        model: stepsModel
+        delegate: Item {
+            id: wrapperItem
+            width: (index === 0) ? descriptionLabel.contentWidth : (wizardWrapper.width / repeat.count)
+            height: 56
+            onXChanged: {
+                //as x changes while delegates are created, direct assignment doesn't work
+                x = (index === (repeat.count-1)) ? (width+32) : (label.width/2) + 8
+            }
+            StatusProgressBar {
+                id: barBorder
+                width: visible ? (parent.width - (label.width/2 + 24)) : 0
+                height: visible ? 8 : 0
+                visible: (index > 0)
+                anchors.verticalCenter: parent.verticalCenter
+                anchors.verticalCenterOffset: -12
+                from: 0
+                //TODO what is to: here?
+                to: 2000
+                value: loadingTime
+                backgroundColor: "transparent"
+                backgroundBorderColor: Theme.palette.primaryColor1
+                fillColor: Theme.palette.primaryColor1
+            }
+            Item {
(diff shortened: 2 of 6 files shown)


src/StatusQ/Components/StatusWizardStepper.qml
```diff
@@ -0,0 +1,79 @@
...
+        model: stepsModel
+        delegate: Item {
+            id: wrapperItem
+            width: (index === 0) ? descriptionLabel.contentWidth : (wizardWrapper.width / repeat.count)
+            height: 56
+            onXChanged: {
+                //as x changes while delegates are created, direct assignment doesn't work
+                x = (index === (repeat.count-1)) ? (width+32) : (label.width/2) + 8
+            }
+            Rectangle {
```
noeliaSD, line 29: Just a suggestion... We have the `StatusProgressBar` control which inherits from `ProgressBar` which could maybe be used here (instead of using Rectangles) and that component could take advantage of its behavior. Please take a look at it and let me know your thoughts!

noeliaSD:
Great component abstraction @alexandraB99 ! 
As a general comment, I was wondering if we could create a specific component, maybe not in StatusQ (or yes), since it could be desktop specific, which includes the specific ListModel created in `StatusWizardStepperPage`. I don't know if it could make sense, as maybe it should be used in different views.

sandbox/pages/StatusWizardStepperPage.qml
```diff
@@ -0,0 +1,82 @@
...
+            stepsListModel.setProperty(i, "stepCompleted", false);
+        }
+    }
+
+    ListModel {
+        id: stepsListModel
+        ListElement {description:"Send Request"; loadingTime: 0; loading: false; stepCompleted: false}
+        ListElement {description:"Receive Response"; loadingTime: 2000; loading: false; stepCompleted: false}
+        ListElement {description:"Confirm Identity"; loadingTime: 2000; loading: false; stepCompleted: false}
+    }
```
PascalPrecht, line 25: How does this behave when the first and third item are `stepCompleted: true` ?

Shouldn't the component rather track at which step it is? E.g. being at step 2 implies that step 1 is completed. Or are we also dealing with cases where one can be at a certain step without having completed a prior one?

src/StatusQ/Components/StatusWizardStepper.qml
PascalPrecht: Fully agree!

src/StatusQ/Components/StatusWizardStepper.qml
```diff
@@ -0,0 +1,79 @@
...
+                        text: index+1
+                        font.pixelSize: 17
+                        color: (stepCompleted && bar.barLoaded) ? Theme.palette.indirectColor1 : Theme.palette.primaryColor1
+                    }
+                }
+                StatusBaseText {
+                    id: descriptionLabel
+                    anchors.bottom: parent.bottom
+                    text: description
+                    font.pixelSize: 13
```
PascalPrecht, line 74: This will need a `color` to work properly in dark mode

sandbox/pages/StatusWizardStepperPage.qml
```diff
@@ -0,0 +1,82 @@
...
+    //simulation code
+
+    StatusWizardStepper {
+        id: wizardStepper
+        width: (parent.width - 50)
+        anchors.top: indicLabel.bottom
+        anchors.topMargin: 16
+        anchors.horizontalCenter: parent.horizontalCenter
+        stepsModel: stepsListModel
```
PascalPrecht, line 80: As mentioned above, I think it'd make sense for it to have an API to know at which step the stepper is. Maybe we can then also have dedicated components to `.complete(i)` and `uncomplete(i)` 

Wdyt?

PascalPrecht (changes requested):
That's a really cool component!
I left some comments as we might want to fine-tune the API

alexandraB99:
> Great component abstraction @alexandraB99 !
> 
> As a general comment, I was wondering if we could create a specific component, maybe not in StatusQ (or yes), since it could be desktop specific, which includes the specific ListModel created in `StatusWizardStepperPage`. I don't know if it could make sense, as maybe it should be used in different views.

Maybe, currently I guess we don't have the use case for desktop fully defined, so the idea is to prepare the component and then use in desktop in the way that would be specified. I used the list only to simulate the behavior of the component, nevertheless this is only my assumption, I have no idea how it should be in reality and what the backend will look like.

sandbox/pages/StatusWizardStepperPage.qml
```diff
@@ -0,0 +1,82 @@
...
+                root.reset();
+            }
+        }
+    }
+
+    StatusBaseText {
+        id: indicLabel
+        anchors.horizontalCenter: parent.horizontalCenter
+        text: "Press Ctrl+R to run the animation"
+        font.pixelSize: 17
```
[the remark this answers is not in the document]
alexandraB99: done

src/StatusQ/Components/StatusWizardStepper.qml
alexandraB99: done. I used `directColor1` as well here.

sandbox/pages/StatusWizardStepperPage.qml
alexandraB99: I don't know as I don't know how that should really work. We agreed with John to make it a fix 3 step for now and change it later if needed. Other than this the completeness could/should come from the backend which is the one to really validate the steps, not the component. This is why I used a ListModel for the simulation, that said, this is only my assumption as there is not much specified or any real use case yet. This should also respond to your other questions regarding `stepCompleted: true`, I'd expect all these should come from the backend.

src/StatusQ/Components/StatusWizardStepper.qml
alexandraB99: ok, nevertheless I had to add another property as the component was not 100% fitting.

alexandraB99:
@PascalPrecht I used `Theme.palette.statusAppLayout.rightPanelBackgroundColor` as the background color for the step number circle as it is the same color. Do we have any other more generic color for this?

sandbox/pages/StatusWizardStepperPage.qml
PascalPrecht: Thanks for your insights @alexandraB99 !
That makes sense for now. 
Regardless of what the backend does, I think it's also okay to propose an API for you **you** think this component **should** work and then take it from there.

I personally like to refer to Angular Material components that I've worked on in the past. They usually have nice thought-through APIs (although, it's very well possible that not everything works as nicely to QML).

If you need any inspiration for designing the API together with the backend, here the stepper component from Angular Material, maybe this is useful: https://material.angular.io/components/stepper/overview

src/StatusQ/Components/StatusWizardStepper.qml
PascalPrecht: Are there any other things that would make using `StatusProgressBar` hard to use here?
Cause then it might be worth to explore teaching `StatusProgressBar` a custom border color, so the component can be reused here.

At least visually, it really does look like this is essentially a progress bar.

sandbox/pages/StatusWizardStepperPage.qml
alexandraB99: so you mean to invent or adapt an API without really knowing what the component should be doing exactly in order to just have something in place? what if we add stuff and we end up using only some percent of all these at the end? I am fine to add something, however I prefer to have the real use case to take reference from as I believe we're creating a component specific to status and not a generic framework that could be used anyrywhere like Angular.

src/StatusQ/Components/StatusWizardStepper.qml
alexandraB99: I think progress bar is ok just by adding a new `backgroundBorderColor` property (which I already did).

sandbox/pages/StatusWizardStepperPage.qml
PascalPrecht: > so you mean to invent or adapt an API without really knowing what the component should be doing exactly in order to just have something in place?

No, I was making a **suggestion** and provide some reference material in case you want look into this.

We're not building anything here "just to have something in place" , which is why we have this discussion in the first place.

What I was trying to say is, that it's fine that you don't know what some backend is doing, because these StatusQ components should come with a well defined API regardless. So when authoring a new component based on designs, you have the freedom to come up with an API that makes sense (which is a good thing!) (see: https://github.com/status-im/StatusQ/issues/499#issuecomment-993302089 as an example). 

If you're not sure what a component **should** be doing, then usually common sense is a good place to start and also looking at what other component libraries with similar components have done (which is why I've linked the Angular Material as a reference). And of course, you can reach out and get opinions (which at latest stage happens here in a PR anyway).

> however I prefer to have the real use case to take reference from as I believe we're creating a component specific to status and not a generic framework that could be used anyrywhere like Angular

The use case reference is in the figma design linked in the original issue as of now. If more designs of the same component are needed to know what the component should be doing (other than representing "steps" in some wizard-like process), **and** common sense isn't enough, then we'll have to wait with implementing the thing.  

re: we're not a generic framework: 
Not sure what your point is here. The stepper component (just like any other) should come with an API that makes sense and is easy to use. You don't have to be a "framework" for that.

Look, I'm just trying to help here and make constructive suggestions. I'm not asking you to do things that waste your time. If you're not happy with certain suggestions for solving a problem that's perfectly fine, as long as we can discuss these things and clarify **why** things shouldn't be that way. 

But if you say the "why" is a missing backend and additional designs for the component are missing to understand what it should be doing, then all I'm saying is: "okay, no problem. It's perfectly fine to use common sense and come up with your own proposal. Don't let this stop you."

If you don't want to, that's fine too. Not trying to force anything here. Again, just suggestions.

PascalPrecht:
Landed!

sandbox/pages/StatusWizardStepperPage.qml
alexandraB99: oh yeah please don't get me wrong either! I was just trying to understand why should I be adding an API that I think makes sense to me when I should be really waiting for the specification. That said, for example, to me might make sense that I expose step circle and bar colors, but at the end it's not necessary. This way, I would be adding complexity (+lines of code) in the component and they wouldn't be used at the end. What I meant with we're not a generic framework is exactly this, for example if that was the case, I would be exposing these properties just in case are needed in some specific use case of a specific project that makes use of my component but now since only status desktop client will be using this component, I would probably choose to keep the API to the necessary level according to the specifications, which are non in place yet :/ . I hope that makes sense.

PR #552, opened 2022-02-08
fix(Status/TagSelector) Results popup component position

### LINKED ISSUE #531: Add support for autocompletion in `StatusTagSelector`, move suggestions popup inside component
As discussed in #526, we need to add support for autocompletion, so that users get suggestions of things to be tagged. It's already implemented outside of `StatusTagSelector` in the demo app and in Status Desktop. This issue is about adding the ability to render the suggestions popup inside the `StatusTagSelector`.
Feel free to put an API proposal here.

alexandraB99:
Fixed alignment when results are being filtered.
If the x position that the popup should be aligned
goes off the screen, the popup appears back on
position 0.

Closes #531

https://user-images.githubusercontent.com/31625338/153080246-34a60ee0-bf9e-4234-8c53-d0bd00efc9ea.mov

### DIFF
--- a/sandbox/demoapp/CreateChatView.qml
+++ b/sandbox/demoapp/CreateChatView.qml
@@ -71,22 +71,27 @@ Page {
             StatusBaseText {
                 id: contactsLabel
                 font.pixelSize: 15
+                anchors.left: parent.left
+                anchors.leftMargin: 8
                 color: Theme.palette.baseColor1
-                text: "Contacts"
+                text: qsTr("Contacts")
             }
             Control {
                 width: 360
                 anchors {
                     top: contactsLabel.bottom
                     topMargin: 8//Style.current.padding
-                    bottom: parent.bottom
+                    bottom: !statusPopupMenuBackgroundContent.visible ?  parent.bottom : undefined
                     bottomMargin: 20//Style.current.bigPadding
                 }
+                height: 16 + (!statusPopupMenuBackgroundContent.visible ? parent.height :
+                        (((userListView.count * 64) > parent.height) ? parent.height : (userListView.count * 64)))
+                x: (statusPopupMenuBackgroundContent.visible && (tagSelector.namesModel.count > 0) &&
+                   ((tagSelector.textEdit.x + 24 + statusPopupMenuBackgroundContent.width) < parent.width))
+                   ? (tagSelector.textEdit.x + 24) : 0
                 background: Rectangle {
                     id: statusPopupMenuBackgroundContent
-                    anchors.left: parent.left
-                    anchors.right: parent.right
-                    height: (userListView.height + 8)
+                    anchors.fill: parent
                     visible: (tagSelector.sortedList.count > 0)
                     color: Theme.palette.statusPopupMenu.backgroundColor
                     radius: 8
@@ -107,9 +112,9 @@ Page {
                 }
                 contentItem: ListView {
                     id: userListView
-                    anchors.left: parent.left
-                    anchors.right: parent.right
-                    height: (count * 64) > parent.height ? parent.height : (count * 64)
+                    anchors.fill: parent
+                    anchors.topMargin: 8
+                    anchors.bottomMargin: 8
                     clip: true
                     model: contactsModel
                     ScrollBar.vertical: ScrollBar {
@@ -125,7 +130,6 @@ Page {
                         Rectangle {
                             id: rectangle
                             anchors.fill: parent
-                            anchors.topMargin: 8
                             anchors.rightMargin: 8
                             anchors.leftMargin: 8
                             radius: 8


PascalPrecht:
That's a nice fine-tuning in the demo app! However, it doesn't actually address what I was referring to (sorry for the confusion!). See this comment: https://github.com/status-im/StatusQ/issues/531#issuecomment-1033425597

I'd still love to merge this. Can you update the commit message as this is not actually a fix in `StatusTagSelector` ?

alexandraB99:
> That's a nice fine-tuning in the demo app! However, it doesn't actually address what I was referring to (sorry for the confusion!). See this comment: [#531 (comment)](https://github.com/status-im/StatusQ/issues/531#issuecomment-1033425597)
> 
> I'd still love to merge this. Can you update the commit message as this is not actually a fix in `StatusTagSelector` ?

@PascalPrecht alright! This fix is also included now and I updated the commit message as well. now it looks like:
results

PascalPrecht:
Thanks @alexandraB99 !
Looking at the changes, I was surprised to see that the changes are still all in demo app view.
I thought `StatusTagSelector` already comes with support for the auto completion (as in, it will take care of rendering the popup accordingly). That's also why I was a bit confused here: https://github.com/status-im/StatusQ/issues/531#issuecomment-1030053140

Just for me to understand: if people want to use `StatusTagSelector` and enable auto completion (getting that popup with the suggestions), do they have to rebuild that popup every time from scratch? If that's the case we probably still want to keep #531 open until `StatusTagSelector` comes with such support built-in.

What do you think?

PascalPrecht:
> I thought about this when starting with this component but since once again, not much is specified when it comes to usage

The design shows clearly that there's a popup rendered when one types into the tag selector input. So from that we can derive that the tag selector will need a way to render a popup with suggestions (whatever those are). This was the reason why #531 was created in addition to #526 in the first place (First build a StatusTagSelector that renders tags, **then** add support for autocompletion/suggestions). This should be a feature of that existing StatusTagSelector component.

What other specifications are needed for this to be true? There's very little point in closing #531 if the component doesn't actually implement that.

> Secondly it's not clear to me whether the tag selector would be used elsewhere with different specifications, e.g. different popup (or not a popup at all)

I think it's unlikely for it to have a different popup, because we use the same throughout the entire system. If that will be the case, we'll have to touch many more components. For the "not a popup at all" case – that's just the default that was implemented in #526 already.

> That said, ideally filtering should be done in the backend, not in the component itself.

I agree. That's also why I've mentioned point 3) here https://github.com/status-im/StatusQ/issues/531#issuecomment-1029280143

> ~~Loading state, shows StatusLoadingIndicator until results arrive~~ Discussed with @John-44 that this isn't needed. We just filter the static list.

^ As you can see, we kinda concluded however, that it's fine to filter on client-side for now. Running a filter on a few hundred elements isn't actually that slow, plus accounts will most likely rarely have hundreds of contacts (more like 40). And lastly, I think we already keep a full list of contacts in the front-end via some model already, so there's no need to then request it again asynchronously.

That being said, I do agree that it should support asynchronous suggestions. Ideally it will support both, but I'm happy to take it step by step and only do the static list for now and create a separate issue for async loading if this is still desired.

So I think the only thing missing really is to have the popup moved **inside** StatusTagSelector and provide an API to hydrate it with data that it can run the filter on.

alexandraB99:
> > I thought about this when starting with this component but since once again, not much is specified when it comes to usage
> 
> The design shows clearly that there's a popup rendered when one types into the tag selector input. So from that we can derive that the tag selector will need a way to render a popup with suggestions (whatever those are). This was the reason why #531 was created in addition to #526 in the first place (First build a StatusTagSelector that renders tags, **then** add support for autocompletion/suggestions). This should be a feature of that existing StatusTagSelector component.
> 
> What other specifications are needed for this to be true? There's very little point in closing #531 if the component doesn't actually implement that.
> 
> > Secondly it's not clear to me whether the tag selector would be used elsewhere with different specifications, e.g. different popup (or not a popup at all)
> 
> I think it's unlikely for it to have a different popup, because we use the same throughout the entire system. If that will be the case, we'll have to touch many more components. For the "not a popup at all" case – that's just the default that was implemented in #526 already.
> 
> > That said, ideally filtering should be done in the backend, not in the component itself.
> 
> I agree. That's also why I've mentioned point 3) here [#531 (comment)](https://github.com/status-im/StatusQ/issues/531#issuecomment-1029280143)
> 
> > ~Loading state, shows StatusLoadingIndicator until results arrive~ Discussed with @John-44 that this isn't needed. We just filter the static list.
> 
> ^ As you can see, we kinda concluded however, that it's fine to filter on client-side for now. Running a filter on a few hundred elements isn't actually that slow, plus accounts will most likely rarely have hundreds of contacts (more like 40). And lastly, I think we already keep a full list of contacts in the front-end via some model already, so there's no need to then request it again asynchronously.
> 
> That being said, I do agree that it should support asynchronous suggestions. Ideally it will support both, but I'm happy to take it step by step and only do the static list for now and create a separate issue for async loading if this is still desired.
> 
> So I think the only thing missing really is to have the popup moved **inside** StatusTagSelector and provide an API to hydrate it with data that it can run the filter on (so it'll actually close #531).
> 
> It doesn't have to be part of this PR then if you don't feel like doing it now. Just need to make sure we don't close #531 accidentally.

ok, as you like :) I'd probably take care of it as part of https://github.com/status-im/status-desktop/issues/4781. Not sure whether you still want to merge this or keep it open until popup is moved in.

PascalPrecht:
Definitely want to merge this as mentioned above. Just wanted to clarify why this isn't closing #531, so we'll have to keep that one open.

PascalPrecht:
This has landed as: https://github.com/status-im/StatusQ/commit/cdf51fefbdec10c84c3987677b806634652c86d4

PR #553, opened 2022-02-09
feat(@StatusSmartIdenticon): Add support for color rings in StatusSmartIdenticon

### LINKED ISSUE #517: Add support for color rings in `StatusSmartIdenticon`
relevant spec: https://github.com/status-im/feature-specs/pull/42/files
relevant figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=1837%3A335128

32 colors in 5 segments of various lengths, please refer to spec for more details

Screen Shot 2022-01-12 at 2 47 42 PM

noeliaSD:
Closes #517
### What does the PR do

It has been created a new control `StatusIdenticonRing` and added it into `StatusSmartIdenticon` component.

To test the component, the needed properties have been added in sandbox models to display the `StatusIdenticonRing` control.

**NOTE:** `StatusIdenticonRing` provides the following API as a first approach (since we don't have the back-end specs clearly defined yet, when they have it clear we can decide if it is better to manage the data in another format):

- `property var ringSpec`: Expected `ListModel` object to iterate through ring segments. Each segment (`ListElement`) contains: 

    1. The `colorId`: from 0 to 31 corresponding with the 32 distinctive segment defined in [figma](https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop-(rolled-back-version-to-recover-a-screen%2C-ignore)?node-id=1837%3A335130) and with the order defined there (it should be defined as well in specs).
    2. The `segmentType`: from 1 to 5 being these integer values the number of units the segment contains.

- `property real totalRingUnits`: Total number of units in the segment
- `property real initalAngleRad`: Initial angle value in radians
- `property real ringPxSize`: Ring thickness

### Affected areas
`StatusSmartIdenticon` component

Added example in Sandbox:
Screenshot 2022-02-09 at 11 01 48


src/StatusQ/Controls/StatusIdenticonRing.qml
```diff
@@ -0,0 +1,59 @@
...
+        context.beginPath()
+        context.arc(xPos, yPos, radius, startAngle, startAngle + arcAngleLen, false/*anticlockwise*/)
+        context.strokeStyle = color
+        context.lineWidth = lineWidth
+        context.stroke()
+    }
+
+    implicitWidth: 40
+    implicitHeight: 40
+    Layout.alignment: Qt.AlignHCenter | Qt.AlignVCenter
```
alexandraB99, line 31: wondering if this is needed here?

sandbox/controls/ListItems.qml
```diff
@@ -357,6 +368,8 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
                       nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2ImYgiNITTlTdG1nUZ5a92VITQxITFiJmIIjSE0htAYQrMHAAD//+wwFVpz+yqXAAAAAElFTkSuQmCC"
         image.isIdenticon: true
         isOnline: true
+        statusListItemIcon.identiconRing.ringSpec: identiconRing1
+        statusListItemIcon.identiconRing.totalRingUnits: 25
```
alexandraB99, line 372: just for the sake of beautifying: I imagine that `statusListItemIcon` is part of `StatusListItem` and not of `StatusMemberListItem` because of Mouse events and z order reasons? I was wondering, is there any case we could avoid 'dotdotting' like `statusListItemIcon.identiconRing.totalRingUnits:` and have something more straight forward like `totalRingUnits:` instead?

sandbox/controls/ListItems.qml
```diff
@@ -347,6 +347,17 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
         requestsCount: 3
     }
 
+    ListModel {
+        id: identiconRing1
```
alexandraB99, line 351: is this model used / going to be used elsewhere? If not, could we assign it directly to the component that's using it?

alexandraB99 (commented):
lovely PR @noeliaSD ! :) just left a few comment, mostly stuff related to beautification :)

sandbox/controls/ListItems.qml
noeliaSD: Yes, I agree with your point of view, however... I don't know which is the best point to expose these properties but giving as much flexibility as possible to the component. `StatusListItem` is a super flexible component that exposes, I don't know if all, but most of its components and allows you to design a super custom list item. 

I'm open to proposals! Maybe it could have sense to add all `StatusIdenticonRing`properties as well in `StatusSmartIdenticon` so in this particular case we'd have something like this:

- `statusListItemIcon.ringSpecs` or `statusListItemIcon.totalRingUnits`

WDYT?

sandbox/controls/ListItems.qml
noeliaSD: Yes, it could be assigned directly!

src/StatusQ/Controls/StatusIdenticonRing.qml
noeliaSD: No needed! Removed!

PascalPrecht (changes requested):
Super cool @noeliaSD !

As @alexandraB99 there's probably room for improvement with regards to developer ergonomics. I've left some thoughts there on how we solved this in other components.

In addition, I think the color ring colors should live in `Theme` for a few reasons: a) that's just the place where colors are defined and b) we want to ensure these colors work across themes, so every theme should probably be able to define their own ring colors (needed to ensure things like contrast ratio)

src/StatusQ/Controls/StatusIdenticonRing.qml
```diff
@@ -0,0 +1,56 @@
...
+    property real ringPxSize: 1.5
+
+    readonly property real segmentRadLen: 2 * Math.PI / identiconRing.totalRingUnits
+    // 32 distinctive segment: They are expected to be used using the array index from 0 to 31 and with the order defined in figma (it should be defined as well in specs.)
+    readonly property var distinctiveColors: ["#000000", "#726F6F", "#C4C4C4", "#E7E7E7", "#FFFFFF", "#00FF00",
+                                              "#009800", "#B8FFBB", "#FFC413", "#9F5947", "#FFFF00", "#A8AC00",
+                                              "#FFFFB0", "#FF5733", "#FF0000", "#9A0000", "#FF9D9D", "#FF0099",
+                                              "#C80078", "#FF00FF", "#900090", "#FFB0FF", "#9E00FF", "#0000FF",
+                                              "#000086", "#9B81FF", "#3FAEF9", "#9A6600", "#00FFFF", "#008694",
+                                              "#C2FFFF", "#00F0B6"]
```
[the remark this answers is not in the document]
noeliaSD: Yes, it is something we should ask @John-44 because as far as I know, colored ring should be unique so don't know if it could be changed, in the perspective of identification, depending on the theme. But surely, these colores are not correct for both views, so yes, it makes more sense to have a pair for each.

sandbox/controls/ListItems.qml
noeliaSD: Ok! It sound like a good idea. Going for that! :)

noeliaSD:
Pushed some updates adding a ring settings object. Please review @alexandraB99 and @PascalPrecht!

noeliaSD:
> it looks like blurred, is this an artifact of the screenshot?

I'm not sure.. It is true that I can see a considerable difference between looking at it in my laptop screen (that it is perfectly defined) but, in the other support screen I have, it seems the pixels are more visible. What about you @alexandraB99 @PascalPrecht ?

Here a test with a ring px size of 2 (instead of the previous one that is 1.5):
Screenshot 2022-02-11 at 10 58 17
Do you appreciate any improvement?

alexandraB99 (commented):
Looks good to me! (don't hate me for this please :) ) should we already add documentation?

noeliaSD:
> Looks good to me! (don't hate me for this please :) ) should we already add documentation?

:) I think it is the best way! Yes I'll check how to do it and add docu, as well, here!

src/StatusQ/Components/StatusSmartIdenticon.qml
```diff
@@ -20,6 +22,13 @@ Loader {
         height: 40
     }
 
+    property StatusIdenticonRingSettings ringSettings: StatusIdenticonRingSettings {
+        ringSpec: undefined
```
PascalPrecht, line 26: Just a nitpic: There's no need to define it as `undefined` here. Leave it out and it'll be undefined as well.

PascalPrecht:
Changes look good to me now! Just need to clarify https://github.com/status-im/StatusQ/pull/553#discussion_r803597455

src/StatusQ/Controls/StatusIdenticonRing.qml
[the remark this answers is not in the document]
John-44: @noeliaSD the colours should be the exactly the same across themes e.g. the ring colours shouldn't change when the user changes themes.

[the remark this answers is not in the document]
noeliaSD: > Oh and they should be added to the design system figma then. We already have a color palette system there. Could introduce an additional section for "verified" colors.

I think @davholik is working on introducing these colours into the design system file. We can keep it now just as it is and when he finishes, we can create a new task for doing the refactor! WDYT?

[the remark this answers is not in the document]
davholik: I just discussed this with @John-44, ideally, we want to try to mirror the identicon color palette in the dark theme with slightly tweaked properties, so it works in a dark environment, but still keeps the unique properties. We need to experiment with this.

So what I would do right now is, that prepare the code in a way that can support separate dark theme variant of this palette. For now, the dark variation of the palette will be identical to the light one but it will be prepared for any future changes.

I am including this identicon palette in the new branch of the design system. It is definitely gonna sit in it's separate group. It will be either part of the data visualization color group or its own. Bring it into the code as a separate group for sure.

src/StatusQ/Core/StatusIdenticonRingSettings.qml
```diff
@@ -0,0 +1,11 @@
+import QtQuick 2.13
+import StatusQ.Core 0.1
+
+QtObject {
+    id: statusIdenticonRingSettings
+
+    property var ringSpec // Expected ListModel object to iterate through ring segments
```
osmaczko, line 7:
```suggestion
    property var ringSpecModel // [ListElement {colorId: 0; segmentLength: 1}, ...]
```
Wouldn't it be more readable than a plain text comment? Also added `Model` postfix to the name of the variable.

src/StatusQ/Controls/StatusIdenticonRing.qml
```diff
@@ -0,0 +1,60 @@
...
+    property StatusIdenticonRingSettings settings: StatusIdenticonRingSettings {
+        ringSpec: undefined
+        totalRingUnits: 1
+        initalAngleRad: 0
+        ringPxSize: 1.5
+    }
+
+    readonly property real segmentRadLen: 2 * Math.PI / settings.totalRingUnits
+    // 32 distinctive segment: They are expected to be used using the array index from 0 to 31 and with the order defined in figma (it should be defined as well in specs.)
+    readonly property var distinctiveColors: ["#000000", "#726F6F", "#C4C4C4", "#E7E7E7", "#FFFFFF", "#00FF00",
```
osmaczko, line 18: I would NOT make it read-only. Clients of StatusQ may want to define their own colors.

src/StatusQ/Controls/StatusIdenticonRing.qml
```diff
@@ -0,0 +1,60 @@
...
+                                              "#009800", "#B8FFBB", "#FFC413", "#9F5947", "#FFFF00", "#A8AC00",
+                                              "#FFFFB0", "#FF5733", "#FF0000", "#9A0000", "#FF9D9D", "#FF0099",
+                                              "#C80078", "#FF00FF", "#900090", "#FFB0FF", "#9E00FF", "#0000FF",
+                                              "#000086", "#9B81FF", "#3FAEF9", "#9A6600", "#00FFFF", "#008694",
+                                              "#C2FFFF", "#00F0B6"]
+
+    function printArcSegment(context, xPos, yPos, radius, color, startAngle, arcAngleLen, lineWidth) {
```
osmaczko, line 25: It does not make much sense to expose it as a public API. Please consider putting this private function into:
```
QtObject {
    id: d
    function printArcSegment(...) {...}
}
```
id: d - detail

or defining it inside `Canvas` component.

osmaczko (changes requested):
Looks great, good job! There are only some small things that may be improved, see comments below.

src/StatusQ/Controls/StatusIdenticonRing.qml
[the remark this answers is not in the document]
osmaczko: @John-44 @davholik 

In order to have all client apps aligned, these colors should come from status-go. Otherwise, we will end up either with different identicon ring colors on mobile and desktop or with colors definition duplication in qml and react, which will be hard to maintain and bug-prone. If we really want to tweak the properties by theme it should be done by appropriate client's transform function, e.g. Qt.darker(identiconRingColors[0], 1.15).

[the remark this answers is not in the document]
PascalPrecht: > If we really want to tweak the properties by theme it should be done by appropriate client's transform function, e.g. Qt.darker(identiconRingColors[0], 1.15).

So the thing with this is that it really depends on what's the impact of tweaks needed to have those colors work in different themes. Lemme elaborate:

1. Often times, a color in one theme can't be mapped to another color in a different theme using calculated rules (.e.g `Qt.darker(...)`). In those cases, the color `red` in light mode isn't just the same `red` in a different shade in darkmode, it's actually a *different* `red`. We have a good amount of such cases in our color palette which is why our `Theme` system is so explicit. 
2. I think there's plans to eventually support custom themes, so users will be able to install themes for Status Desktop, in which case the colors provided by status-go might not go well with the theme at all.

I think we have a handful of options with their trade-offs:

1. We make color rings look exactly the same regardless of what the theme looks like (proposed by @John-44), at the cost of it possibly not being pleasant to the eye in any other theme that wasn't designed with those colors in mind (even in darkmode already those colors look too sharp, which is probably why @davholik wants to tweak them on a per theme basis).
2. We let the theme decide what those ring colors are, ensuring all colors work well in every theme, however, at the cost of the rings to be hard to recognize/verify across themes (as they might look different)
3. We drop the idea of supporting custom themes in the future entirely, at which point we'll only need to support light and darkmode colors which can then be either defined in our themes on the client side, or come from status-go (doesn't really matter)

The argument that this can cause inconsistencies across apps (desktop, mobile, etc..) is a good one. Which probably means we'll have to go with 1) or 3)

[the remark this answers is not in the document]
osmaczko: > 1. Often times, a color in one theme can't be mapped to another color in a different theme using calculated rules (.e.g `Qt.darker(...)`).

Yes, that's obvious, but it is not what I meant. I put it in the context, what I wanted to state is:
All colors should come from status-go (and so be the same for every theme) **ONLY THEN**
>If we really want to tweak the properties by theme it should be done by appropriate client's transform function, e.g. Qt.darker(identiconRingColors[0], 1.15).

src/StatusQ/Core/StatusIdenticonRingSettings.qml
noeliaSD: I like it! :)

src/StatusQ/Controls/StatusIdenticonRing.qml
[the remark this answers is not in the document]
osmaczko: > 1. We make color rings look exactly the same regardless of what the theme looks like (proposed by @John-44), at the cost of it possibly not being pleasant to the eye in any other theme that wasn't designed with those colors in mind (even in darkmode already those colors look too sharp, which is probably why @davholik wants to tweak them on a per theme basis).

I opt for this one. Identicon ring is a mechanism for visual Identity representation, it is meant to be consumed subconsciously by the Users. In my opinion, if colors will differ per Theme human brain will not handle it.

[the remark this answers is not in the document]
noeliaSD: Taking into account all your comments and since we still have some solutions open, I suggest the following:
- To make `StatusIdenticonRing` the most flexible _StatusQ_ component, we can expose the `distinctiveColors` property as a required property to be set by the consumer. Then we can close the component itself as it is.
- To manage colors/themes, we can now add the list of colors (all 32 defined in figma) in the Themes (now the same for light and dark) and when we use the component in the demo app, just like we are configuring the model in case of _ringSpecsModel property_, we can set up an array of the 32 colors (using Themes) we've now defined. So, we will have prepared the definition of colors in different themes (just waiting for the dark mode updates) and we will leave the door open to give the responsibility to the backend to decide which colors are passed to the ring.

WDYT?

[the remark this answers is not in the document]
PascalPrecht: Sounds like a good "best-of-both-worlds" solution - I like it

noeliaSD:
Pushed all changes following your comments. I've also added documentation for the new components. Please @alexandraB99, @PascalPrecht  and @osmaczko, take a look!!

PascalPrecht (approved):
Very cool, also nice that it has docs!

osmaczko (approved):
Great docs!

PR #556, opened 2022-02-10
feat(@StatusPinInput): Introduce `StatusPinInput` control

### LINKED ISSUE #524: Introduce `StatusPinInput`
figma: https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=1621%3A367655

Screen Shot 2022-01-12 at 3 29 36 PM

noeliaSD:
Closes #524

### What does the PR do
It has been created `StatusPinInput` control that allows customzing its **circle diameter**, **circle spacing** and **pin length**. It contains a `TextInput` object that will provide the component, input text management like _validation rules_.

It has been added new page in **sandbox** to play with `StatusPinInput` control.

### Affected areas
`StatusPinInput` control

Added example in Sandbox:

https://user-images.githubusercontent.com/97019400/153446460-f67f876f-b6a5-4c0d-892d-01b5fd57410a.mov


src/StatusQ/Controls/StatusPinInput.qml
```diff
@@ -0,0 +1,137 @@
...
+    width: (root.circleDiameter + root.circleSpacing) * root.pinLen
+    height: root.circleDiameter
+
+    Component.onCompleted: {
+        statesInitialization()
+        inputText.forceActiveFocus()
+    }
+
+    // Pin input data management object:
+    TextInput {
```
alexandraB99, line 35: could this be a `StatusBaseInput` or `StatusInput`?

src/StatusQ/Controls/StatusPinInput.qml
```diff
@@ -0,0 +1,137 @@
...
+                     state: background.innerState
+                     anchors.centerIn: parent
+                     height: width
+                     color: Theme.palette.primaryColor1
+                     radius: 0.5 * width
+
+                     states: [
+                         State {
+                             name: "NEXT"
+                             PropertyChanges {target: inner; width: root.circleDiameter / 2}
```
alexandraB99, line 94: since all states are changing the same property and using the same duration (for transitions) would it make sense to use inline javascript instead and have `Behavior on width { NumberAnimation { duration: 200 } }` do the job?

src/StatusQ/Controls/StatusPinInput.qml
noeliaSD: I'm not sure if it is necessary to use these controls as they are more complex and in the end this object is only used to manage data internally and take advantage of its validation rules (unless one of them implements more suitable or specific validation rules specific for Status, which I have not been able to see).

If we use, for example, `StatusBaseInput`, every time we instantiate a `StatusPinInput` object, we will also be creating a **Rectangle** with a **StatusIcon**, **StatusBaseText**, etc.. and all of this, only for managing data.. 

But if you can see another advantage on using them here, please let me know and, yes, for sure we can modify it!

src/StatusQ/Controls/StatusPinInput.qml
noeliaSD: Totally right! Changed!

sandbox/pages/StatusPinInputPage.qml
```diff
@@ -0,0 +1,76 @@
...
+        font.pixelSize: 30
+        font.bold: true
+    }
+
+    StatusPinInput {
+        id: regexPinInput
+        anchors.horizontalCenter: parent.horizontalCenter
+        validator:  RegularExpressionValidator { regularExpression: /[0-9A-Za-z@]+/ }
+        circleDiameter: 22
+        circleSpacing: 22
```
PascalPrecht, line 60: Not a requirement but nice-to-have: let users define how many fields are allowed for the pin (right now it seems to default to six, not sure if this can change in the future).

Maybe it'd make sense to make this configurable and still default to six. Gives us flexibility without a big cost.

src/StatusQ/Controls/StatusPinInput.qml
```diff
@@ -0,0 +1,115 @@
+import QtQuick 2.0
+import StatusQ.Core.Theme 0.1
+
+Item {
+    id: root
+
+    property string pinInput: inputText.text
+    property var validator: inputText.validator
+    property int pinLen: 6
```
PascalPrecht, line 9: Ah, here's the property that lets us configure the field count. Ignore the comment above!

src/StatusQ/Controls/StatusPinInput.qml
PascalPrecht: > But if you can see another advantage on using them here, please let me know and, yes, for sure we can modify it!

I think you're bringing good arguments that speak against using StatusBaseInput. Here's another one that would speak for it:

The validation system used for `StatusPinInput` is different from StatusQ's validation system, which is more flexible that QML's built-in solution.

I wonder if we could find a way to have StatusPinInput leverage our validation system as well, not because we need more flexibility or validation options, but because we now have inconsistent APIs.

For example, users of StatusQ will know that there's `StatusValidator` types which can be attached to StatusQ controls via `validators` property. However, this component uses `validator` and the validators are of different type/syntax as well, which can be confusing.

You'd need to keep track of which components of StatusQ don't use the StatusQ validation system.

If adding a `StatusBaseInput` here is problematic because of complexity, the only other way I see this work is to add the same validation mechanism to a simpler type like `TextInput`, if we aim for API consistency.

WDYT?

PascalPrecht (commented):
Great work @noeliaSD !

Like @alexandraB99 I think there's reasons to use `StatusBaseInput` inside of this component. Curious to here your thoughts.

One minor nitpic: Can you remove the `@` from the commit message?

noeliaSD:
I'll also add documentation here!

PascalPrecht:
There's one more thing I forgot to add:

Right now there's no "focus" indicator in the component. When I was testing the component in sandbox, I didn't know which of the instance were focused. Would be cool if we could come up with a solution to that.

noeliaSD:
> There's one more thing I forgot to add:
> 
> Right now there's no "focus" indicator in the component. When I was testing the component in sandbox, I didn't know which of the instance were focused. Would be cool if we could come up with a solution to that.

Do you mean something visual?

PascalPrecht:
Yes exactly. Actually there's a few things missing in terms of visual feedback but the critical ones are:

1. No cursor feedback when hovering the control (I don't actually know whether I can click the thing or not)
2. If I do click, I get no visual feedback that I put focus on that control so no idea anything happened

Obviously 2) is mainly an issue because there's multiple of such component in the demo page which won't really happen in practice, but I should still get some feedback to know that I can activate the control or now that I'm **in** that control and can start typing.

noeliaSD:
> Yes exactly. Actually there's a few things missing in terms of visual feedback but the critical ones are:
> 
> 1. No cursor feedback when hovering the control (I don't actually know whether I can click the thing or not)
> 2. If I do click, I get no visual feedback that I put focus on that control so no idea anything happened
> 
> Obviously 2) is mainly an issue because there's multiple of such component in the demo page which won't really happen in practice, but I should still get some feedback to know that I can activate the control or now that I'm **in** that control and can start typing.

The proposal for focus, and agreed with @John-44 is that we could incorporate an animation in the first circle, expanding and minimizing it, when the focus is on the component, just as if a cursor was blinking.

noeliaSD:
> > Yes exactly. Actually there's a few things missing in terms of visual feedback but the critical ones are:
> > 
> > 1. No cursor feedback when hovering the control (I don't actually know whether I can click the thing or not)
> > 2. If I do click, I get no visual feedback that I put focus on that control so no idea anything happened
> > 
> > Obviously 2) is mainly an issue because there's multiple of such component in the demo page which won't really happen in practice, but I should still get some feedback to know that I can activate the control or now that I'm **in** that control and can start typing.
> 
> The proposal for focus, and agreed with @John-44 is that we could incorporate an animation in the first circle, expanding and minimizing it, when the focus is on the component, just as if a cursor was blinking.

Pushed the animation when focused and feedback when hovering. Please, play with it and let me know your feelings @PascalPrecht, @alexandraB99 !

Here a video:

https://user-images.githubusercontent.com/97019400/154424898-49e266b6-9a28-45c0-b699-2be6e885a336.mov

noeliaSD:
Added component documentation.

src/StatusQ/Controls/StatusPinInput.qml
noeliaSD: After looking at `StatusValidator` and taking into account 2 factors (to keep API consistency and use `TextInput` to manage data), I've created the following 2 new `StatusValidator` controls with their corresponding documentation:

- `StatusRegularExpressionValidator` which wraps a QML type `RegularExpressionValidator`.

- `StatusIntValidator` which wraps a QML type `IntValidator`.

@alexandraB99 and @PascalPrecht take a look at the solution and I'll be glad to read your opinions! I hope you feel it works! :)

PascalPrecht:
This has landed as https://github.com/status-im/StatusQ/commit/1217771fd058e3f9c1b669e4ecce3965f6d4ed3c

Great work!

PR #558, opened 2022-02-14
feat(StatusQ): Moving docs outside sandbox

alexandraB99:
To generate docs, from `doc` folder run
/path/to/Qt/installation/bin/qdoc statusq.qdocconf

A folder named "html" will be generated under `doc`,
open the statusq-index.html to get to the home page.


PascalPrecht:
This is really cool!
Probably time to start thinking about automating the release process so we generate a website everytime we out put a release.

Massive work!

PR #560, opened 2022-02-17
feat(StatusChatList): Highlight chat item at creation time

borismelnik:
Part of: https://github.com/status-im/status-desktop/issues/4767

### DIFF
--- a/src/StatusQ/Components/StatusChatList.qml
+++ b/src/StatusQ/Components/StatusChatList.qml
@@ -101,6 +101,7 @@ Column {
                     muted: model.muted
                     hasUnreadMessages: model.hasUnreadMessages
                     notificationsCount: model.notificationsCount
+                    highlightWhenCreated: model.highlight
                     selected: model.active
                     icon.color: model.color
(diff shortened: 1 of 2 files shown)


PascalPrecht (changes requested):
I'm not sure I fully understand why this is needed. I'd assume the data provided will highlight the correct chat list item no?

If this change is still needed, can you please update the commit message to be a feature commit for StatusChatList?

borismelnik:
> I'm not sure I fully understand why this is needed. I'd assume the data provided will highlight the correct chat list item no?
> 
> If this change is still needed, can you please update the commit message to be a feature commit for StatusChatList?

@PascalPrecht yeah that property used for `highlight chat item at creation time ` and  should be `invalidated` after first click.

borismelnik:
it happens when app received message with `chatDto` with property `highlight: true`, which means app should tell user about some changes. Unfortunately, the `chatDto` has zero unread messages or notifications.

PascalPrecht (approved):

PR #561, opened 2022-02-17
fix(StatusListItem): add propagateTitleClicks property to StatusListItem

jrainville:
Makes it possible disable the propagation of the mouse event when it's not wanted, for example in the search popup, where if the user clicks the title, we want to show the profile popup, but not navigate to the channel

The original isn't changed  for other components using StatusListItem, since the property start at true

This was needed to fix the issue found by Abhi here: https://github.com/status-im/status-desktop/pull/4803#issuecomment-1043064425

### DIFF
--- a/src/StatusQ/Components/StatusListItem.qml
+++ b/src/StatusQ/Components/StatusListItem.qml
@@ -20,6 +20,7 @@ Rectangle {
     property real rightPadding: 16
     property bool enabled: true
     property bool highlighted: false
+    property bool propagateTitleClicks: true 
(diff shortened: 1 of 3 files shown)


PascalPrecht (approved):
Approving although I'm not super happy with this change.  Would prefer some event delegation-like solution. I think @borismelnik had suggested a similar thing. Let's wait for his review/thoughts.

jrainville:
@borismelnik can you review when you have time

borismelnik (approved):

PR #562, opened 2022-02-18
fix(Components/StatusTagSelector): Updates and fixes in the component 

alexandraB99:
Moved `remove` function below remove contact signal
also removed online status badge from CreateChatView

Fixed as well name tags to be adapting on parent's
width and scrolling the list to the end when this is
bigger than the available width. Switcehd to use
nameCountLimit property where needed

### DIFF
--- a/sandbox/demoapp/CreateChatView.qml
+++ b/sandbox/demoapp/CreateChatView.qml
@@ -170,27 +170,6 @@ Page {
                             font.pixelSize: 15
                         }
 
-                        StatusBadge {
-                            id: statusBadge
-                            width: 15
-                            height: 15
-                            anchors.left: contactImage.right
-                            anchors.leftMargin: -8
-                            anchors.bottom: contactImage.bottom
-                            border.width: 3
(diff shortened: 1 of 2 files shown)


PascalPrecht (commented):
This LGTM!
Can you please update the commit message though to actually say what it does? "Various fixes" isn't very meaningful.

Also, the scope is `StatusTagSelector`

alexandraB99:
> This LGTM! Can you please update the commit message though to actually say what it does? "Various fixes" isn't very meaningful.
> 
> Also, the scope is `StatusTagSelector`

@PascalPrecht I updated a bit the title however the commit fixes/improves various things which cannot fit in the title, everything is explained in the actual commit message.
commit

Let me know if you come up with a better wording for the title!

PascalPrecht:
Updated your commit and landed it as https://github.com/status-im/StatusQ/commit/07a5dc09ac2a9ede38fffcfda00a02e3ea6b7c5e

PR #566, opened 2022-03-02
feat(StatusChatListItem): enable assigning emoji to chat item

jrainville:
image


Khushboo-dev-cpp (approved):

jrainville:
@PascalPrecht I added the sandbox example, and doing it made me realize that using the `utils` doesn't work there, since it's from the StatusDesktop app itself, so accessing `Emoji` wasn't possible.

Therefore, I had to move all the Emoji code from the main app's utils to StatusQ's Utils. Now the component has access to the Emoji singleton and the main app still has access.

I think it's also cleaner, since now the path to access emojis in the app is `qrc:/StatusQ/src/assets/twemoji` instead of using relative paths.

Let me know if you think it doesn't belong in StatusQ and if so, how I should handle the emoji parsing then, since I get the emoji as unicode from te model itself.

PascalPrecht:
> Let me know if you think it doesn't belong in StatusQ and if so, how I should handle the emoji parsing then, since I get the emoji as unicode from te model itself.

I think this is the easiest way forward. We also have Backpressure in here.
Thanks for the PR!

PR #569, opened 2022-03-04
Feat/status list item tags

alaibe:
image

### DIFF
--- a/sandbox/controls/ListItems.qml
+++ b/sandbox/controls/ListItems.qml
@@ -322,6 +322,17 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
         badge.icon.isLetterIdenticon: true
     }
 
+    StatusListItem {
+        title: "List Item with Tags"
+        icon.isLetterIdenticon: true
+        tags: [
+            StatusListItemTag {
+                title: qsTr("Tag 1")
+                icon.isLetterIdenticon: true
+            }
+        ]
+    }
+
     StatusDescriptionListItem {
         title: "Title"
         subTitle: "Subtitle"
--- a/src/StatusQ/Components/StatusListItem.qml
+++ b/src/StatusQ/Components/StatusListItem.qml
@@ -23,6 +23,7 @@ Rectangle {
     property bool propagateTitleClicks: true 
     property int type: StatusListItem.Type.Primary
     property list<Item> components
+    property list<StatusListItemTag> tags
 
     property StatusIconSettings icon: StatusIconSettings {
         height: isLetterIdenticon ? 40 : 20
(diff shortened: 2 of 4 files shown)


sandbox/controls/ListItems.qml
```diff
@@ -322,6 +322,19 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
...
     }
 
+    StatusListItem {
+        title: "List Item with Tags"
+        icon.isLetterIdenticon: true
+        tags: [
+            StatusListItemTag {
+                title: qsTr("Tag 1")
+                icon.isLetterIdenticon: true
+                onClick: {
```
alexandraB99, line 332: Please add a console.log or something or delete it if not needed :)

src/StatusQ/Components/StatusListItemTag.qml
```diff
@@ -0,0 +1,77 @@
...
+
+Rectangle {
+    id: root
+    width: titleText.contentWidth + 60
+    height: 30
+    color: Theme.palette.primaryColor2
+    radius: 15
+
+    property string title: ""
+    signal click()
```
alexandraB99, line 15: would it make sense to rename this to "clicked"? just because when using the component, is more likely that someone types `onClicked`

sandbox/controls/ListItems.qml
alaibe: removed

src/StatusQ/Components/StatusListItemTag.qml
alaibe: changed to clicked, habits with web but indeed qml use clicked

noeliaSD:
LGTM!!

noeliaSD (approved):

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -21,8 +21,10 @@ Rectangle {
     property bool enabled: true
     property bool highlighted: false
     property bool propagateTitleClicks: true 
+    property bool clickable: true
```
PascalPrecht, line 24: We can remove this. You get access to `sensor` which can be `enabled`. So within and outside of this component you can check for `sensor.enabled`

PascalPrecht (changes requested):
Changes LGTM, but we don't need the `clickable` property :)

src/StatusQ/Components/StatusListItem.qml
alaibe: @PascalPrecht  even when I disable the sensor, there is still one part which have a mouse area enable:

https://github.com/status-im/StatusQ/blob/master/src/StatusQ/Components/StatusListItem.qml#L175-L184

Do you think this is a bug?

alaibe: synced offline, this was a bug and it should be fixed now

PascalPrecht:
Sweet!

PR #572, opened 2022-03-09
fix(@StatusListItem): Use tags model

alaibe:
Tags are meant to be dynamic, so switch from static declaration to a model

### DIFF
--- a/sandbox/controls/ListItems.qml
+++ b/sandbox/controls/ListItems.qml
@@ -331,12 +331,11 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
     StatusListItem {
         title: "List Item with Tags"
         icon.isLetterIdenticon: true
-        tags: [
-            StatusListItemTag {
-                title: qsTr("Tag 1")
-                icon.isLetterIdenticon: true
-            }
-        ]
+        bottomModel: 3
+        bottomDelegate: StatusListItemTag {
+            title: "tag"
+            icon.isLetterIdenticon: true
+        }
     }
 
     StatusDescriptionListItem {
--- a/src/StatusQ/Components/StatusListItem.qml
+++ b/src/StatusQ/Components/StatusListItem.qml
@@ -23,7 +23,8 @@ Rectangle {
     property bool propagateTitleClicks: true 
     property int type: StatusListItem.Type.Primary
     property list<Item> components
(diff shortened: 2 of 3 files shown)


sandbox/controls/ListItems.qml
```diff
@@ -331,12 +331,11 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
     StatusListItem {
         title: "List Item with Tags"
         icon.isLetterIdenticon: true
-        tags: [
-            StatusListItemTag {
-                title: qsTr("Tag 1")
-                icon.isLetterIdenticon: true
-            }
-        ]
+        tagsModel: 3
```
PascalPrecht, line 334: Shouldn't model be a `ListModel` here instead of a number?

sandbox/controls/ListItems.qml
```diff
@@ -331,12 +331,11 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
...
+        tagsDelegate: StatusListItemTag {
+            title: "tag"
+            icon.isLetterIdenticon: true
+        }
```
PascalPrecht, line 338: If we have a `tagsModel` property, we already know we want to render tag elements. So we might as well put it in the component, but I like it that you made it a `delegate` so one can do something else.

Maybe we should then no longer call it `tag*` because we could also render other elements? Then, `tagsModel` would be called something else as well I believe...

Curious to hear what @noeliaSD and @alexandraB99 think

alaibe: I could call that `bottomComponentModel` ? 
I don t think I had much choice for doing a delegate because off needed customization

Anyway, happy to rename whatever you prefer

sandbox/controls/ListItems.qml
alaibe: model also accept number as per the docs

sandbox/controls/ListItems.qml
PascalPrecht: `bottomComponentModel` or `bottomSlotModel` sounds good to me. Then also `bottomComponentDelegate`/`bottomSlotDelegate`.

noeliaSD: Yeah! bottom*Model/Delegate souds good to me. It is more generic!

PR #573, opened 2022-03-09
fix(StatusQ.Validators): fix regular expression test in validator

jrainville:


src/StatusQ/Controls/Validators/StatusEnodeValidator.qml
```diff
@@ -0,0 +1,17 @@
+import StatusQ.Core.Utils 0.1
+import StatusQ.Controls 0.1
+
+StatusValidator {
+
+    name: "url"
```
PascalPrecht, line 6: Should probably be called `enode` or `enodeUrl`?

PascalPrecht (approved):
Minor comment, great addition.

Can you make the scope of the message `StatusQ.Validators` ?

noeliaSD (commented):
Just only 2 minor suggestions / questions:

1. Could it make sense / be possible to define the `validatorObj` property as a specific `RegularExpressionValidator` QML Type for that? Maybe the best way is using the existing `StatusRegularExpressionValidator` instead of derive from `StatusValidator` and just only fixing the `regularExpression` property to your specific one.
2. Could you, as it is a new StatusQ component, add the necessary QML documentation? You can take as an example, `StatusRegularExpressionValidator` or `StatusIntValidator` controls.

jrainville:
@noeliaSD my bad, I didn't know about that Validator, I'm now using that one, though I found that it didn't work correctly. It failed with 
```
qrc:/StatusQ/src/StatusQ/Controls/Validators/StatusRegularExpressionValidator.qml:59: TypeError: Property 'validate' of object QRegularExpressionValidator(0x55601f018130) is not a function
qrc:/StatusQ/src/StatusQ/Controls/Validators/StatusRegularExpressionValidator.qml:59: TypeError: Type error
```
With the fix I did, now everything works.

src/StatusQ/Controls/Validators/StatusRegularExpressionValidator.qml
```diff
@@ -51,10 +51,9 @@ StatusValidator {
 
     name: "regex"
     errorMessage: "Please enter a valid regular expression."
-    validatorObj: RegularExpressionValidator { regularExpression: root.regularExpression }
 
     validate: function (t) {
         // Basic validation management
-        return root.validatorObj.validate() === RegularExpressionValidator.Acceptable
+        return new RegExp(root.regularExpression).test(t)
```
PascalPrecht, line 57: Seems like @stefandunca has done a similar fix here https://github.com/status-im/StatusQ/pull/576/files#diff-ffa848f7eb38f5a1379b105eab7afff658a8558240d4b5a0161b85ed0ed36d19R58

PascalPrecht: Nooooo.... :D 
Your PR is scoped smaller, so it's easier to land. Could've updated yours to match those changes and Stefan could've rebased on top of yours

jrainville: lol, ok then, reopening 😆

PR #574, opened 2022-03-09
feat(StatusQ/Controls): Adding StatusSeedPhraseInput control

### LINKED ISSUE #567: Implement StatusSeedPhraseInput component
seed

Notes
- the autocomplete only searches the BIP39 wordlist
- the moment a user types the 3rd letter of a word, the word should be automatically completed and the focus should jump to the next word entry field.

alexandraB99:
Closes #567

Also added an example in Inputs page

https://user-images.githubusercontent.com/31625338/157976842-d629dc9a-d8f1-43b9-af7f-6040ca120d9f.mov

### DIFF
--- a/src/StatusQ/Controls/StatusSeedPhraseInput.qml
+++ b/src/StatusQ/Controls/StatusSeedPhraseInput.qml
@@ -1,0 +1,125 @@
+import QtQuick 2.12
+import QtGraphicalEffects 1.13
+
+import StatusQ.Controls 0.1
+import StatusQ.Popups 0.1
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+Item {
+    id: root
+    width: 162
+    height: 44
+    property alias textEdit: seedWordInput
+    property alias text: seedWordInput.text
+    property string leftComponentText: ""
+    property ListModel inputList: ListModel { }
+    property ListModel filteredList: ListModel { }
+    signal doneInsertingWord(string word)
+
+    StatusInput {
+        id: seedWordInput
+        implicitWidth: parent.width
+        implicitHeight: parent.height
+        input.anchors.topMargin: 0
+        input.anchors.leftMargin: 0
+        input.anchors.rightMargin: 0
+        input.leftComponent: StatusBaseText {
+            text: root.leftComponentText
+            color: seedWordInput.input.edit.activeFocus ?
+                   Theme.palette.primaryColor1 : Theme.palette.baseColor1
+            font.pixelSize: 15
+        }
+        onTextChanged: {
+            filteredList.clear();
+            if (text !== "") {
+                for (var i = 0; i < inputList.count;i++) {
+                    if (inputList.get(i).seedWord.startsWith(text)) {
+                        filteredList.insert(filteredList.count, {"seedWord": inputList.get(i).seedWord});
+                    }
+                }
+                seedSuggestionsList.model = filteredList;
+                if ((text.length === 3) && (filteredList.count === 1) &&
+                    ((input.edit.keyEvent !== Qt.Key_Backspace) && (input.edit.keyEvent !== Qt.Key_Delete))) {
+                    seedWordInput.text = filteredList.get(0).seedWord;
+                    seedWordInput.input.edit.cursorPosition = seedWordInput.text.length;
+                    seedSuggestionsList.model = 0;
+                    root.doneInsertingWord(seedWordInput.text);
+                }
+            } else {
+                seedSuggestionsList.model = 0;
+            }
+        }
+    }
+
+    Item {
+        id: suggListContainer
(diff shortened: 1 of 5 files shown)


src/StatusQ/Components/StatusSeedPhraseInput.qml
```diff
@@ -0,0 +1,154 @@
...
+        text: ((grid.count > 12) && !grid.secondPage) ? qsTr("Next") : qsTr("Import")
+        onClicked: {
+            if ((grid.count > 12) && !grid.secondPage) {
+               grid.positionViewAtEnd();
+            } else {
+                //import seed word to text input
+            }
+        }
+    }
+}
```
PascalPrecht, line 154: Hey @alexandraB99 ,
sorry don't wanna get ahead of myself here, but just noticing: I don't think we want the seed phrase input the entire page, but really just the input + its suggestions thingy.

The page will be built in desktop. This also allows for using StatusSeedPhraseInput in different ways.

src/StatusQ/Controls/StatusSeedPhraseInput copy.qml
```diff
@@ -0,0 +1,109 @@
...
+                    onClicked: {
+                        seedWordInput.text = seedWord;
+                        seedWordInput.input.edit.cursorPosition = seedWordInput.text.length;
+                        seedSuggestionsList.model = 0;
+                    }
+                }
+            }
+        }
+    }
+}
```
PascalPrecht, line 109: This file has been committed mistakenly I assume?

src/StatusQ/Controls/StatusSeedPhraseInput.qml
```diff
@@ -0,0 +1,110 @@
...
+            color: seedWordInput.input.edit.activeFocus ?
+                   Theme.palette.primaryColor1 : Theme.palette.baseColor1
+            font.pixelSize: 15
+        }
+        onTextChanged: {
+            filteredList.clear();
+            if (text !== "") {
+                for (var i = 0; i < helloModel.count;i++) {
+                    if (helloModel.get(i).seedWord.startsWith(text)) {
+                        filteredList.insert(filteredList.count, {"seedWord": helloModel.get(i).seedWord});
```
PascalPrecht, line 31: I might be wrong, but I don't think we can rely on `helloModel` here..

PascalPrecht (commented):
Nice work @alexandraB99 !
I think I found 1-2 thinks that need change, lemme know what you think!

alexandraB99:
> Nice work @alexandraB99 !
> I think I found 1-2 thinks that need change, lemme know what you think!

thanks @PascalPrecht, really funny issues :D I overlooked at those as I was moving files around..I have removed both now, please have a second look!

src/StatusQ/Controls/StatusSeedPhraseInput.qml
```diff
@@ -0,0 +1,110 @@
...
+                    height: parent.height
+                    color: sensor.containsMouse? Theme.palette.primaryColor1 : "transparent"
+                }
+                StatusBaseText {
+                    id: suggWord
+                    anchors.left: parent.left
+                    anchors.leftMargin: 14
+                    anchors.verticalCenter: parent.verticalCenter
+                    text: seedWord
+                    color: Theme.palette.directColor1
```
PascalPrecht, line 91: This color should probably change when hover state of the item is active. If there's nothing about it in the design, common sense would be to make it behave exactly like `StatusMenuItem`s. (I think they have white font color on hover, but don't know from the top of my head)

PascalPrecht (changes requested):
Nice @alexandraB99 

This looks good to me all in all! Just that the popup item text color probably shouldn't stay black all the time

src/StatusQ/Controls/StatusSeedPhraseInput.qml
alexandraB99: I don't see any color change when hovering the Items of a PopupMenu

noeliaSD (changes requested):
Nice component!!! :)

Just one issue when trying to remove a word, there is a point (4 chars¿?) that it does not allow to remove and autocompletes the word again. See the video to better understanding:

https://user-images.githubusercontent.com/97019400/158543032-418c6160-197c-4ffa-b806-8a6ca2cc3780.mov

noeliaSD:
> > Nice component!!! :)
> > Just one issue when trying to remove a word, there is a point (4 chars¿?) that it does not allow to remove and autocompletes the word again. See the video to better understanding:
> > Screen.Recording.2022-03-16.at.08.56.05.mov
> 
> good catch @noeliaSD , fixed!

Great!

noeliaSD (approved):

PascalPrecht (approved):
Great work @alexandraB99 !

PR #576, opened 2022-03-10
feat(StatusInput): Add ignore invalid input option to StatusInput

stefandunca:
## Changes

- [x] test the validator
- [x] add an optional feature to StatusInput not to allow typing characters that are not validated
- [x] add a test entry in the sandbox app.
- [x] add basic documentation

Required for merging #5010

## Screenshots

New entry
image

Allow typing only what matches regex
image

updates #4961

### DIFF
--- a/src/StatusQ/Controls/StatusInput.qml
+++ b/src/StatusQ/Controls/StatusInput.qml
@@ -44,7 +44,8 @@ Item {
 
     enum ValidationMode {
         OnlyWhenDirty, // validates input only after it has become dirty
-        Always // validates input even before it has become dirty
+        Always, // validates input even before it has become dirty
+        IgnoreInvalidInput // ignore the new content if it doesn't match
     }
 
     property var errors: ({})
@@ -57,7 +58,9 @@ Item {
         root.errorMessage = ""
     }
 
+    property string _previousText: text
     function validate() {
+
         if (!statusBaseInput.dirty && validationMode === StatusInput.ValidationMode.OnlyWhenDirty) {
             return
         }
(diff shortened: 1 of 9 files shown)


src/StatusQ/Controls/StatusInput.qml
```diff
@@ -37,6 +37,7 @@ Item {
     property list<StatusAsyncValidator> asyncValidators
     property int validationMode: StatusInput.ValidationMode.OnlyWhenDirty
     property string validatedValue
+    property bool ignoreInvalidInput: false
```
PascalPrecht, line 40: This is a cool feature. Could consider introducing a new `ValidationMode` for that instead of introducing a new prop, but I'm happy to keep this.

PascalPrecht (changes requested):
This was marked as WIP, but I still went ahead to take a look.

Few things:

There's two things happening here 1) a fix in the regex validator and 2) a new feature in `StatusInput`.

Can you make this two atomic commits for both of these things, taking into account that each of them have their scope (`fix` and `feat` for `StatusRegexValidator` and `StatusInput` respectively)?

Otherwise, great stuff!
Also @noeliaSD might want to see some documentation for the newly introduced feature.

PascalPrecht:
@stefandunca this can be rebased on top of https://github.com/status-im/StatusQ/pull/573 which has just landed in master, which means you only need the feature commit

stefandunca:
Thanks for the heads up, will make the required changes

PascalPrecht (approved):
Nice job! Love that it includes tests and docs :)
This just needs a rebase

noeliaSD:
Late answer but just only want to say.. Super nice that it includes unit tests!! Thanks @stefandunca for introducing them in the project!! :)

PR #579, opened 2022-03-14
feat(StatusIconSetting): add emojiSize setting for icons

jrainville:
Also enables emojjis in the ChatIdentifier and chat header

### DIFF
--- a/src/StatusQ/Components/StatusChatListItem.qml
+++ b/src/StatusQ/Components/StatusChatListItem.qml
@@ -28,8 +28,9 @@ Rectangle {
         width: 24
         height: 24
         color: Theme.palette.miscColor5
-        letterSize: emoji ? 14 : 15
+        letterSize: emoji ? 10 : 15
         emoji: ""
+        emojiSize: Emoji.size.verySmall
         charactersLen: 1
     }
     property int type: StatusChatListItem.Type.PublicChat
--- a/src/StatusQ/Components/StatusLetterIdenticon.qml
+++ b/src/StatusQ/Components/StatusLetterIdenticon.qml
@@ -9,6 +9,7 @@ Rectangle {
     property alias identiconText: identiconText
     property string name
     property string emoji
+    property string emojiSize
     property int letterSize: 21
     property int charactersLen: 1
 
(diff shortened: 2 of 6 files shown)


src/StatusQ/Components/StatusChatListItem.qml
```diff
@@ -28,8 +28,9 @@ Rectangle {
         width: 24
         height: 24
         color: Theme.palette.miscColor5
-        letterSize: emoji ? 14 : 15
+        letterSize: emoji ? 10 : 15
         emoji: ""
+        emojiSize: "16x16"
```
PascalPrecht, line 33: Now that we get more emoji related properties, it'd be cool to have them similar to `icon` and `image` properties so that we could do: `emoji.name: ""`, `emoji.size: ...` etc.

I'll let you decide if we want to do this as part of this PR or keep it for the future because it'd be a breaking change.

PascalPrecht (approved):

src/StatusQ/Components/StatusChatListItem.qml
Khushboo-dev-cpp, line 33: Hey I notice in the Emoji.qml, we have predefined emoji sizes 
    readonly property var size: {
        "big": "72x72",
        "middle": "32x32",
        "small": "18x18"
    }
maybe we can added 16*16 here as well?  also by default it was already 18*18 what going wrong visually in the chat items?

Khushboo-dev-cpp (commented):
just a question

src/StatusQ/Components/StatusChatListItem.qml
jrainville: I can add it to the sizes. The problem was that `18x18;  was too big in the channel list.

Khushboo-dev-cpp: got it :)

jrainville: Good idea. Like you said though, it would be a breaking change, so I'll wait until the wallet and chat changes have settled so I don't conflict with concurrent changes.

jrainville: Updated

PR #580, opened 2022-03-15
fix(StatusContactRequestIndicatorListItem): hide badge when value is 0

jrainville:


src/assets/img/icons/edit_pencil.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
PascalPrecht, line 1: One thing to note: All icons so far have been created with spacing so that they are always 24x24. This is to ensure they behave predictably and can be centered properly regardless of their actual icon size (because they live in an invisible samve-sized box).

This icon seems to be 16x16. Is there a chance we can get a 24x24 of this as well (not making the icon itself bigger but the padding it comes with?

jrainville: Scratch that, there is an issue with the sizes in my PR. The icons in that new design are smaller. I'll try to figure it out

jrainville: Ok, I fixed my display with `icon.width: 16`. 
So, again, do you think I should ask the designer for a 24x24 icon?
Design: https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=1657%3A268542

PascalPrecht: For consistency's sake it'd be great if all icons are 24x24 and then use whatever their natural size is inside of that 24x24 box. But I don't want this to be a blocker here.

Your call.

PascalPrecht (approved):

src/assets/img/icons/edit_pencil.svg
jrainville: I got a 24x24 icon from David. All good now!

PR #583, opened 2022-03-18
feat(StatusQ.Components/Controls): New `StatusListPicker` and `StatusItemPicker` selector component / control

### LINKED ISSUE #563: New selector component
see designs

https://www.figma.com/file/idUoxN7OIW2Jpp3PMJ1Rl8/%E2%9A%99%EF%B8%8F-Settings-%7C-Desktop?node-id=973%3A105672

Picker button

Screen Shot 2022-02-26 at 1 22 06 PM

Selector

Screen Shot 2022-02-26 at 1 13 53 PM

noeliaSD:
Closes status-im/StatusQ#563

### What does the PR do
It adds the `StatusItemPicker` control and `StatusListPicker` component and its corresponding documentation.

The new selector component is composed by a `StatusButtonPicker` and a drop-down list.

It adds sections in `ListView` by using `category` model property.

It incorporates a searcher in header.

The selector type can be customised as a `StatusRadioButton` or a `StatusCheckbox`.

It adds dynamic selection mechanism.

It adds a dynamic text fit between `name` and `shortName` text components.

It adds a specific page in `sandbox` and some models to play with that.

It extends `StatusPickerButton` to allow more than one type.

### Affected areas

`StatusPickerButton`, `StatusListPicker` component and `StatusItemPicker` control

https://user-images.githubusercontent.com/97019400/160801734-8d5780fe-f793-404a-b375-21107c3325fd.mov

### Checklist

- [x] update sandbox app


alexandraB99 (commented):
LGTM! I am just wondering whether it makes sense to move the Delegate in a separate file to shorten a bit StatusListPicker? it's 400 lines currently

src/StatusQ/Components/StatusListPicker.qml
```diff
@@ -0,0 +1,400 @@
...
+       \qmlproperty string StatusListPicker::printSymbol
+       This property holds whether the selected items will display a composition of the role `symbol` in the StatusPickerButton text or just only the role `shortName`.
+    */
+    property bool printSymbol: false
+
+    /*!
+       \qmlproperty string StatusListPicker::maxPickerHeight
+       This property holds the maximum drop-down list height allowed. The drop-down list height will be set as the minimum value between the list content height and the maxPickerHeight property.
+    */
+    property int maxPickerHeight: 718
```
stefandunca, line 97: Great catch. `maxPickerHeight` can be used to contain the popup in parent's/window boundaries so the user can access all the items by scrolling. I can't see how we control the popup position or width for the same use-case,  when it is spanned partially outside its parent view and obstruct checkboxes. A note in the doc header for a user to consider this on a case by case basis would do.

src/StatusQ/Components/StatusListPicker.qml
```diff
@@ -0,0 +1,400 @@
...
+                height: root.height
+                width: root.width
+                onClicked: { currencyPicker.close() }
+            }
+       \endqml
+    */
+    function close() { if(picker.visible) picker.visible = false }
+
+    QtObject {
+        id: d
```
stefandunca, line 114: Nice approach in hiding implementation

src/StatusQ/Components/StatusListPicker.qml
```diff
@@ -0,0 +1,400 @@
...
+   For a list of components available see StatusQ.
+*/
+Item {
+    id: root
+
+    /*!
+       \qmlproperty string StatusListPicker::inputList
+       This property holds the data that will be populated in the list picker.
+
+       NOTE: It is an in / out property so the `selected` role will be dynamically changed according to the user actions.
```
stefandunca, line 45: How about adding a note for users that the model should not change. Will this help?

src/StatusQ/Components/StatusListPicker.qml
```diff
@@ -0,0 +1,400 @@
...
+       \qmlproperty string StatusListPicker::searchText
+       This property holds the placeholder text the searcher input displays by default.
+    */
+    property string searchText: qsTr("Search")
+
+    /*!
+       \qmlproperty string StatusListPicker::isRadioBtn
+       This property holds whether the list picker is a single or a multiple picker option by displaying a StatusRadioButton or StatusCheckBox instead.
+    */
+    property bool isRadioBtn: true
```
stefandunca, line 85: Just a thought: `isRadioBtn` looks to me like a side effect of the main aspect of multi-selection vs. single selection. What do you think of having this as `multiSelection` and note in docs the radiobuttons vs. checkboxes side effect? Will this approach provide more self-documenting?

src/StatusQ/Components/StatusListPicker.qml
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    width: 110
+    height: 38
+
+    Component.onCompleted: d.initialize()
+
+    StatusPickerButton {
+        id: btn
+        width: parent.width
```
stefandunca, line 198: Do we need to control the position? Otherwise `anchors.fill: parent` will do the same and also show explicitly the position

src/StatusQ/Components/StatusListPicker.qml
```diff
@@ -0,0 +1,400 @@
...
+        width: content.itemWidth
+        height: Math.min(content.contentHeight + content.anchors.topMargin + content.anchors.bottomMargin, root.maxPickerHeight)
+        anchors.left: btn.left
+        anchors.top: btn.bottom
+        anchors.topMargin: 4
+        visible: false
+        color: Theme.palette.statusPopupMenu.backgroundColor
+        radius: 8
+        layer.enabled: true
+        layer.effect: DropShadow {
```
stefandunca, line 228: Nice touch.

Just a thought for the future: DropShadow is costly and forces layering which makes the entire component to be drawn twice every frame (one in the offscreen buffer, blending then on screen). Also DropShadow is done using high-quality gaussian blur and documentation says that for 60 FPS full image you need an advanced GPU. Also graphical effects are moved to compatibility module in Qt6.

It will be great to experiment in the future with a generic component shadow based on `Gradient` applied in a rectangle, reusable `RectangleWithShadow` StatusQ.Control. It will do the job for rectangular backgrounds like most of the cases.

stefandunca (approved):
This looks and feels as a great complete user experience. Good job.

src/StatusQ/Components/StatusListPicker.qml
noeliaSD: Yeah, I decided only to expose the height because the width and position seemed to me too much but, yes, it could be considered as well.

What do you think if we add in the header something like this:
"Make sure to position the component inside its parent also when it is spanned in order to achieve a better user experience."

src/StatusQ/Components/StatusListPicker.qml
noeliaSD: Could it work something like this?

"This model property should not change so it is an in / out property where the `selected` role will be dynamically changed according to user actions"

src/StatusQ/Components/StatusListPicker.qml
noeliaSD: I agree! It makes more sense to rename it to `multiSelection`. I will do the change. Thanks :)

src/StatusQ/Components/StatusListPicker.qml
noeliaSD: Thanks for that info. I'll create a `niceToHave` task to play with that in the future!

noeliaSD:
> 

Could it work creating a `StatusItemPicker` and replacing all the content on the `ListView` delegate item?

noeliaSD:
> 

@alexandraB99  Could it work creating a `StatusItemPicker` and replacing all the content on the `ListView` delegate item?

src/StatusQ/Components/StatusListPicker.qml
stefandunca: Sounds great.

src/StatusQ/Components/StatusListPicker.qml
stefandunca: Very clear, thanks

noeliaSD:
Just pushed changes according to your comments. Please take a look at them @alexandraB99, @stefandunca and @PascalPrecht. Thanks!!

src/StatusQ/Controls/StatusItemPicker.qml
```diff
@@ -0,0 +1,183 @@
...
+   For a list of components available see StatusQ.
+*/
+Item {
+    id: root
+
+    /*!
+       \qmlproperty string StatusItemPicker::imageSource
+       This property holds the image source information.
+    */
+    property string imageSource
```
PascalPrecht, line 45: Would it make sense to make this `StatusImageSettings`?

src/StatusQ/Controls/StatusPickerButton.qml
```diff
@@ -11,7 +11,13 @@ Button {
...
     property color bgColor: Theme.palette.baseColor2
     property color contentColor: Theme.palette.baseColor1
-    signal clicked()
+    property var type: StatusPickerButton.Type.NEXT
+    property int lateralMargins: 16
+
+    enum Type {
+        NEXT,
+        DOWN
+    }
```
PascalPrecht, line 20: Just cosmetics: I think so far we've not used all-caps for type enums. If we want to stay consistent, we don't do it here either. But again, just cosmetics, no strong opinion there from my side.

0x-r4bbit (approved):

PR #584, opened 2022-03-18
refactor(StatusBaseInput): use RowLayout for position management

osmaczko:
- used RowLayout for position management which simplifies code a lot
- fixed implicitHight for "Multiline with max/min" scenario, see video:

https://user-images.githubusercontent.com/33099791/159000891-10a234dc-457a-4708-b68f-e274a9fab123.mp4

### DIFF
--- a/sandbox/pages/StatusInputPage.qml
+++ b/sandbox/pages/StatusInputPage.qml
@@ -146,7 +146,7 @@ Column {
     StatusInput {
         label: "Label"
         input.placeholderText: "Input width component (right side)"
-        input.component: StatusIcon {
+        input.rightComponent: StatusIcon {
             icon: "cancel"
             height: 16
             color: Theme.palette.dangerColor1
@@ -173,25 +173,27 @@ Column {
     }
 
     StatusInput {
+        property bool toggled: true
         label: "Input with emoji icon"
         input.placeholderText: "Enter Name"
-        input.icon.emoji: "😁"
+        input.icon.emoji: toggled ? "😁" : "🧸"
         input.icon.color: "blue"
         input.isIconSelectable: true
         onIconClicked: {
-            // launch emoji popup
+            toggled = !toggled
         }
     }
 
     StatusInput {
+        property bool toggled: true
         label: "Input with selectable icon which is not an emoji"
         input.placeholderText: "Enter Name"
         input.icon.emoji: ""
-        input.icon.name: "filled-account"
+        input.icon.name: toggled ? "filled-account" : "image"
         input.icon.color: "blue"
         input.isIconSelectable: true
         onIconClicked: {
-            // launch emoji popup
+            toggled = !toggled
         }
     }
 }
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -161,6 +161,7 @@ Item {
                         id: edit
 
                         property string previousText: text
+                        property var keyEvent
 
                         width: flick.width
                         height: flick.height
@@ -176,7 +177,8 @@ Item {
 
                         Keys.onReturnPressed: event.accepted = !multiline
                         Keys.onEnterPressed: event.accepted = !multiline
-                        Keys.forwardTo: [statusBaseInput]
+                        Keys.forwardTo: [root]
+                        Keys.onPressed: edit.keyEvent = event.key
 
                         onCursorRectangleChanged: flick.ensureVisible(cursorRectangle)
                         onActiveFocusChanged: if (root.pristine) root.pristine = false
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -1,15 +1,15 @@
 import QtQuick 2.14
 
 import QtQuick.Controls 2.14 as QC
+import QtQuick.Layouts 1.14
 
 import StatusQ.Components 0.1
 import StatusQ.Controls 0.1
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 
-
 Item {
-    id: statusBaseInput
+    id: root
 
     property bool multiline: false
 
@@ -66,40 +66,24 @@ Item {
         }
     }
 
-    property Item leftComponent
-    property Item component
+    property Component leftComponent
+    property Component rightComponent
 
-    signal iconClicked()
-
-    onClearableChanged: {
-        if (clearable && !component) {
-            clearButtonLoader.active = true
-            clearButtonLoader.parent = statusBaseInputComponentSlot
-        } else {
-            clearButtonLoader.active = false
-        }
-    }
-
-    onLeftComponentChanged: {
-        if (!!leftComponent) {
-            leftComponent.parent = statusBaseInputLeftComponentSlot;
-        }
-    }
-
-    onComponentChanged: {
-        if (!!component) {
-            component.parent = statusBaseInputComponentSlot
-        }
-    }
+    signal iconClicked
 
     implicitWidth: 448
-    implicitHeight: multiline ? Math.max((edit.implicitHeight + topPadding + bottomPadding), 44) : 44
+    implicitHeight: multiline ? Math.min(Math.max(
+                                    (edit.implicitHeight + topPadding + bottomPadding),
+                                    44, root.minimumHeight), root.maximumHeight) : 44
 
     Rectangle {
         width: parent.width
         height: maximumHeight != 0 ? Math.min(
-                                         minimumHeight != 0 ? Math.max(statusBaseInput.implicitHeight, minimumHeight)
-                                         : statusBaseInput.implicitHeight, maximumHeight) : parent.height
+                                         minimumHeight
+                                         != 0 ? Math.max(
+                                                    root.implicitHeight,
+                                                    minimumHeight) : root.implicitHeight,
+                                         maximumHeight) : parent.height
         color: Theme.palette.baseColor2
         radius: 8
 
@@ -107,15 +91,14 @@ Item {
 
         border.width: 1
         border.color: {
-            if (!statusBaseInput.valid && statusBaseInput.dirty) {
-                return Theme.palette.dangerColor1;
+            if (!root.valid && root.dirty) {
+                return Theme.palette.dangerColor1
             }
             if (edit.activeFocus) {
-                return Theme.palette.primaryColor1;
+                return Theme.palette.primaryColor1
             }
             return sensor.containsMouse ? Theme.palette.primaryColor2 : "transparent"
         }
-
 
         MouseArea {
             id: sensor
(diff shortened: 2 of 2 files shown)


osmaczko:
It has become extremely easy to add new components there, instead of adding yet another condition to anchors left, right, etc. it is just enough to add the component to RowLayout:
```
diff --git a/src/StatusQ/Controls/StatusBaseInput.qml b/src/StatusQ/Controls/StatusBaseInput.qml
index 54bbc5d..45fa5d9 100644
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -133,6 +133,11 @@ Item {
 
                 clip: true
 
+                Text {
+                    Layout.alignment: Qt.AlignVCenter
+                    text: qsTr("%1 %2").arg(root.implicitHeight).arg(edit.implicitHeight)
+                }
+
                 StatusSmartIdenticon {
                     id: emoji
```

https://user-images.githubusercontent.com/33099791/159001626-e1d7ba60-1b5e-4db0-abb5-85f8c50a63ad.mp4

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -120,155 +124,142 @@ Item {
...
-                color: statusBaseInput.icon.color
-                visible: !!statusBaseInput.icon.name && !statusBaseInput.isIconSelectable
-            }
+                clip: true
 
-            Flickable {
-                id: flick
-                anchors.top: parent.top
-                anchors.bottom: parent.bottom
-                anchors.left: (statusIcon.visible && statusBaseInput.leftIcon) ?
```
osmaczko, line 165: That is no longer required, it was not maintainable at all.

osmaczko:
Real-world example (I had to reapply changes required by `StatusSeedPhraseInput` due to conflicts):
- changes required to `StatusBaseInput` before refactor using manual positioning:
```diff
diff --git a/src/StatusQ/Controls/StatusBaseInput.qml b/src/StatusQ/Controls/StatusBaseInput.qml
index a14df16..4326900 100644
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -66,6 +66,7 @@ Item {
         }
     }
 
+    property Item leftComponent
     property Item component
 
     signal iconClicked()
@@ -79,6 +80,12 @@ Item {
         }
     }
 
+    onLeftComponentChanged: {
+        if (!!leftComponent) {
+            leftComponent.parent = statusBaseInputLeftComponentSlot;
+        }
+    }
+
     onComponentChanged: {
         if (!!component) {
             component.parent = statusBaseInputComponentSlot
@@ -158,19 +165,38 @@ Item {
                 visible: !!statusBaseInput.icon.name && !statusBaseInput.isIconSelectable
             }
 
+            Item {
+                id: statusBaseInputLeftComponentSlot
+                anchors.left: parent.left
+                anchors.leftMargin: 12
+                width: childrenRect.width
+                height: childrenRect.height
+                anchors.verticalCenter: parent.verticalCenter
+            }
+
             Flickable {
                 id: flick
                 anchors.top: parent.top
                 anchors.bottom: parent.bottom
-                anchors.left: (statusIcon.visible && statusBaseInput.leftIcon) ?
-                              statusIcon.right : emoji.visible ? emoji.right: parent.left
+                anchors.left: {
+                    if (statusIcon.visible && statusBaseInput.leftIcon) {
+                        return statusIcon.right;
+                    } else if (emoji.visible) {
+                        return emoji.right;
+                    } else if (!!statusBaseInput.leftComponent) {
+                        return statusBaseInputLeftComponentSlot.right;
+                    } else {
+                        return parent.left;
+                    }
+                }
                 anchors.right: {
                     if (!!statusBaseInput.component) {
                         return statusBaseInputComponentSlot.left
                     }
                     return statusIcon.visible && !statusBaseInput.leftIcon ? statusIcon.left : parent.right
                 }
-                anchors.leftMargin: statusIcon.visible && statusBaseInput.leftIcon ? 8
+                anchors.leftMargin: (statusIcon.visible && statusBaseInput.leftIcon)
+                                    || !!statusBaseInput.leftComponent ? 8
                                     : statusBaseInput.leftPadding
                 anchors.rightMargin: {
                     return clearable ? clearButtonLoader.width + 12 : 
@@ -194,6 +220,7 @@ Item {
                 TextEdit {
                     id: edit
                     property string previousText: text
+                    property var keyEvent
                     width: flick.width
                     height: flick.height
                     verticalAlignment: Text.AlignVCenter
@@ -212,6 +239,10 @@ Item {
                         }
                     }
 
+                    Keys.onPressed: {
+                        edit.keyEvent = event.key;
+                    }
+
                     Keys.onReturnPressed: {
                         if (multiline) {
                             event.accepted = false
```

- changes required to `StatusBaseInput` after refactor using layouts:
```diff
diff --git a/src/StatusQ/Controls/StatusBaseInput.qml b/src/StatusQ/Controls/StatusBaseInput.qml
index 54bbc5d..a2cb94e 100644
--- a/src/StatusQ/Controls/StatusBaseInput.qml
+++ b/src/StatusQ/Controls/StatusBaseInput.qml
@@ -66,6 +66,7 @@ Item {
         }
     }
 
+    property Component leftComponent
     property Item component
 
     signal iconClicked
@@ -133,6 +134,10 @@ Item {
 
                 clip: true
 
+                Loader {
+                    sourceComponent: root.leftComponent
+                }
+
                 StatusSmartIdenticon {
                     id: emoji
 
@@ -202,6 +207,7 @@ Item {
                         id: edit
 
                         property string previousText: text
+                        property var keyEvent
 
                         width: flick.width
                         height: flick.height
@@ -218,6 +224,7 @@ Item {
                         Keys.onReturnPressed: event.accepted = !multiline
                         Keys.onEnterPressed: event.accepted = !multiline
                         Keys.forwardTo: [statusBaseInput]
+                        Keys.onPressed: edit.keyEvent = event.key
 
                         onCursorRectangleChanged: flick.ensureVisible(cursorRectangle)
                         onActiveFocusChanged: if (root.pristine) root.pristine = false
```

sandbox/pages/StatusInputPage.qml
```diff
@@ -146,7 +146,7 @@ Column {
     StatusInput {
         label: "Label"
         input.placeholderText: "Input width component (right side)"
-        input.component: StatusIcon {
+        input.rightComponent: StatusIcon {
```
PascalPrecht, line 149: This is a breaking change. We need to be very careful with merging this.
I'd rather ensure `component` is still working and introduce a `rightComponent` property that acts as an alias.

PascalPrecht (approved):
Tested it locally, everything seems to work just fine.

I'd like @alexandraB99 and @noeliaSD to have a closer at the changes.
@osmaczko you've introduced a breaking change in this PR that I'd vote for reconsidering. 

Lemme know what you thinkn

sandbox/pages/StatusInputPage.qml
osmaczko: Yes, that's right, it's a breaking change. Thanks for pointing that out. The status app needs to adapt.

Current StatusInput interface is:
```
property Item leftComponent
property Item component
```
it's inconsistent, `component` says nothing about its position, how could clients possibly interpret that? How will they know that `component` is what is meant to be `rightComponent`?

I could make an alias for that to not break clients' code. How long are we going to maintain this misleading property? Do we have any process for deprecation in place? Cause for me it would be a candidate for deprecation for sure.

PascalPrecht: > it's inconsistent, component says nothing about its position, how could clients possibly interpret that? How will they know that component is what is meant to be rightComponent?

This is because `component` was introduced way before we knew `leftComponent` will be needed. So originally there was only `component`. One of the common things when software historically grows, even more important that we avoid breaking consumers (which is why I haven't changed `component` back then).

> I could make an alias for that to not break clients' code. How long are we going to maintain this misleading property? Do we have any process for deprecation in place? Cause for me it would be a candidate for deprecation for sure.

We don't have an official deprecation process in place, partly also because we're still in 0.x so theoretically, breaking changes are allowed without further notice. I'd still like to keep them at a minimum/push them back because StatusQ is now heavily used inside Desktop and it'd be bad if we'd break it.

I agree it's good to think about a deprecation process. Could imagine, once a thing is marked as deprecated it can be removed in the next major version, giving consumers time to adapt. 

But again, probably only makes sense once we've reached 1.x.

I don't think at least, that introducing a property alias here is going to hurt maintenance that much.

osmaczko: > I don't think at least, that introducing a property alias here is going to hurt maintenance that much.

It doesn't hurt maintenance that much, it hurts the interface readability.

```
StatusInput {
  component: WhatToPutHere{}
}
```
If you were the client, what would the `component` mean to you? What would you need to do to figure that out?

> I'd still like to keep them at a minimum/push them back because StatusQ is now heavily used inside Desktop and it'd be bad if we'd break it.

At some point, we will need to remove these deprecated components/properties anyway. In my opinion, it's better to do it now, since the number of clients can only grow in time (and so the number of changes required to adapt).

> We don't have an official deprecation process in place, partly also because we're still in 0.x so theoretically, breaking changes are allowed without further notice. I'd still like to keep them at a minimum/push them back because StatusQ is now heavily used inside Desktop and it'd be bad if we'd break it.

If we have that possibility right now, we should use it as much as possible (when something is clearly wrong).

PascalPrecht: > If you were the client, what would the component mean to you? What would you need to do to figure that out?

Ideally this is solved by documentation. Naming is hard.

noeliaSD: If I've not missed something, I've made a quick find in the desktop project and I think it will be easy to update if we remove the `component` property so.. if it does not require so much effort, we can remove it and leave the component clean.

osmaczko:
@PascalPrecht adapted in https://github.com/status-im/status-desktop/pull/5099

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -127,191 +110,130 @@ Item {
...
-                    } else {
-                        return parent.left;
-                    }
-                }
-                anchors.right: {
-                    if (!!statusBaseInput.component) {
-                        return statusBaseInputComponentSlot.left
+                Loader {
+                    sourceComponent: {
+                        if (root.leftComponent) return root.leftComponent
```
stefandunca, line 124: A nice approach in simplifying mutually exclusive components

src/StatusQ/Controls/StatusBaseInput.qml
```diff
@@ -127,191 +110,130 @@ Item {
...
+                        selectedTextColor: color
+                        focus: true
+                        font.pixelSize: 15
+                        font.family: Theme.palette.baseFont.name
+                        color: Theme.palette.directColor1
+                        wrapMode: root.multiline ? Text.WrapAtWordBoundaryOrAnywhere : TextEdit.NoWrap
+
+                        Keys.onReturnPressed: event.accepted = !multiline
+                        Keys.onEnterPressed: event.accepted = !multiline
+                        Keys.forwardTo: [statusBaseInput]
```
stefandunca, line 180: This looks like an assumption. Even if the control is designed to be embedded in StatusInput not referencing instances outside this component would make it clear.

Also, this generates a QML warning used on its own for which I propose the following test to catch it (note that it is based on this PR): https://github.com/status-im/StatusQ/pull/590
```suggestion
                        Keys.forwardTo: [root.input]
```

stefandunca (approved):
Great job. Please have a look at my suggestion and let me know what you think.

src/StatusQ/Controls/StatusBaseInput.qml
osmaczko: Great catch!

stefandunca: @osmaczko thanks for pointing out that the correct solution is
```suggestion
                        Keys.forwardTo: [root]
```

PascalPrecht:
@osmaczko this looks like @stefandunca 's comments still need to be addressed, correct?

Khushboo-dev-cpp:
Hey @osmaczko I can see that the changes done here dont match the design here https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=4205%3A175746

the spacing between emoji icon from the left and spacing between icon  and text seems different than the one in design. 

Did you update this as per some system design.?

osmaczko:
> Hey @osmaczko I can see that the changes done here dont match the design here https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=4205%3A175746
> 
> the spacing between emoji icon from the left and spacing between icon and text seems different than the one in design.
> 
> Did you update this as per some system design.?

Oh, I haven't noticed the difference after my changes.

before my changes:
image
after my changes:
image

---

fix: https://github.com/status-im/StatusQ/pull/605

Khushboo-dev-cpp:
Hey in the screenshot before and after for the emoji do you notice the difference in left padding from border? there is a difference :)

osmaczko:
> Hey in the screenshot before and after for the emoji do you notice the difference in left padding from border? there is a difference :)

yes of course ;p see the fix: https://github.com/status-im/StatusQ/pull/605

PR #588, opened 2022-03-21
feat(assets): Add new icon for community settings

borismelnik:
Part of: https://github.com/status-im/status-desktop/issues/4932


src/assets/img/icons/invite-users.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
PascalPrecht, line 1: Would be good if this was 24x24 like all other icons.

PascalPrecht (commented):
Can you update the commit message to be `feat(StatusQ.Core): ..` or `feat(assets): ...` at least?

src/assets/img/icons/invite-users.svg
borismelnik: okay :)

PascalPrecht: This is so that all icons have predicable positioning no matter how big they are inside of the 24x24 box

borismelnik: @PascalPrecht so what is the best way to do it? Should I change just view box?

PascalPrecht: Usually, the designer has prepared all icons in the designs so that they are surrounded by such a box and can then be exported that way.

Maybe you can reach out to the designer in charge and ask the same.

PR #590, opened 2022-03-21
test(StatusInput): add regression unit test for StatusInput

stefandunca:
Adds regression test for #584

Add test helpers to monitor QML output using a declarative approach.
Improves the QML tests structure.

### DIFF
--- a/tests/TestControls/TestHelpers/MonitorQtOutput.h
+++ b/tests/TestControls/TestHelpers/MonitorQtOutput.h
@@ -1,0 +1,32 @@
+#pragma once
+
+#include <QQuickItem>
+#include <QtGlobal>
+
+#include <memory>
+#include <mutex>
+
+///
+/// \brief Monitor output for tests and declarativelly control message handler availability
+/// \todo Check that QML doesn't keep instance between test runs
+///
+class MonitorQtOutput : public QQuickItem
+{
+    Q_OBJECT
+public:
+    MonitorQtOutput();
+    ~MonitorQtOutput();
+
+    Q_INVOKABLE QString qtOuput();
+
+signals:
+
+private:
+    static void qtMessageOutput(QtMsgType type, const QMessageLogContext &context, const QString &msg);
+
+    // Use it to keep track of qInstallMessageHandler call
+    static std::weak_ptr<QString> m_qtMessageOutputForSharing;
+    static std::mutex m_mutex;
+    std::shared_ptr<QString> m_thisMessageOutput;
+    int m_start = 0;
+};
--- a/tests/TestControls/tst_test-StatusBaseInput.qml
+++ b/tests/TestControls/tst_test-StatusBaseInput.qml
@@ -1,0 +1,52 @@
+import QtQuick 2.0
+import QtTest 1.0
+
+import StatusQ.Controls 0.1
+import StatusQ.Controls.Validators 0.1
+
+import StatusQ.TestHelpers 0.1
+
+Item {
+    width: 300
+    height: 100
+
+    StatusBaseInput {
+        id: statusInput
+        text: "Control under test"
+        placeholderText: "Placeholder"
+        focus: true
+    }
+
+    TestCase {
+        id: testCase
+        name: "CheckQmlWarnings"
+
+        when: windowShown
+
+        //
+        // Test guards
+
+        function initTestCase() {
+        }
+
+        function cleanup() {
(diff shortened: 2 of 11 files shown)


PascalPrecht:
I'm a little late to the party, but can you tell me how I can verify that these tests work? We should update the README with instructions as well (can be part of another PR)

tests/TestControls/tst_test-statusinput.qml
```diff
@@ -32,6 +34,13 @@ Item {
 
         when: windowShown
 
+        //> Simulate key and wait for side effects
+        function pressKeyAndWait(key, time) {
```
alaibe, line 38: will this be usefull for many test? I think we just duplicated this function, maybe we can extract this?

alaibe (approved):

stefandunca:
> I'm a little late to the party, but can you tell me how I can verify that these tests work? We should update the README with instructions as well (can be part of another PR)

@PascalPrecht I added this readme file: https://github.com/status-im/StatusQ/pull/590/files#diff-4cfdda6d53eac90934bf621814cd6f2166f91078fac58c2f56f53224a3c5abaa

Let me know if clear enough

tests/TestControls/tst_test-statusinput.qml
stefandunca: Done. Thanks

tests/TestControls/TestHelpers/MonitorQtOutput.h
```diff
@@ -0,0 +1,31 @@
...
+    Q_INVOKABLE QString qtOuput();
+
+signals:
+
+private:
+    static void qtMessageOutput(QtMsgType type, const QMessageLogContext &context, const QString &msg);
+    // Use it to keep track of qInstallMessageHandler call
+    static std::weak_ptr<QString> _qtMessageOutputForSharing;
+    static std::mutex _mutex;
+    std::shared_ptr<QString> _thisMessageOutput;
```
osmaczko, line 29: We don't have coding guidelines in place, but for naming conventions, I think we should follow: https://github.com/status-im/status-desktop/tree/fix/cpp-structure. Or we can agree on using "_" prefix for private members. All I want to say is that we should stay consistent across all cpp source code across Status projects.

tests/TestControls/tst_test-StatusBaseInput.qml
```diff
@@ -0,0 +1,52 @@
...
+        //
+        // Tests
+
+        function test_initial_empty_is_valid() {
+            mouseClick(statusInput)
+            // Do some editing
+            TestUtils.pressKeyAndWait(testCase, statusInput, Qt.Key_B)
+            TestUtils.pressKeyAndWait(testCase, statusInput, Qt.Key_Left)
+            TestUtils.pressKeyAndWait(testCase, statusInput, Qt.Key_A)
+            verify(qtOuput.qtOuput().length === 0, "No output expected. Check qtOuput.qtOuput()")
```
osmaczko, line 45: Maybe we could print here the unexpected output directly?

osmaczko (approved):
That's great! Thanks to that we'll be able to catch any warnings for StatusQ reusable components in an automated way.

In the future we should:
- integrate it with CI
- add 'warnings free' tests for all StatusQ components

tests/TestControls/TestHelpers/MonitorQtOutput.h
stefandunca: Thanks for the heads up. I see Hungarian notation and no decoration approach in the same file: https://github.com/status-im/status-desktop/blob/185457d94fb3f4718eed3df12d3af61d739aaaa8/src-cpp-structure/projects/App/Core/StatusSyntaxHighlighter.h ; Wondering which is preferred, I would go fo no decoration.

@iurimatias codding convention looks like a good topic to address in one of the next C++ architecture meetings

tests/TestControls/tst_test-StatusBaseInput.qml
stefandunca: Will have to check.

tests/TestControls/tst_test-statusinput.qml
```diff
@@ -4,6 +4,8 @@ import QtTest 1.0
 import StatusQ.Controls 0.1
```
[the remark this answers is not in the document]
stefandunca: Will fix

PR #593, opened 2022-03-22
feat(StatusQ.Controls): add StatusTokenInlineSelector

osmaczko:
Closes: status-im/status-desktop#493

https://user-images.githubusercontent.com/33099791/159451255-d75dd9ce-9b03-46b6-930f-c14b89d5644b.mp4

### Checklist

- [x] update sandbox app


src/StatusQ/Controls/StatusTokenInlineSelector.qml
```diff
@@ -0,0 +1,170 @@
...
+       This is a property function used to acquire image source for given token.
+    */
+    property var tokenImageSourceGetter: function (token) {
+        return "../../assets/img/icons/%1.png".arg(token)
+    }
+
+    signal triggered(real amount, string token)
+
+    StatusBaseText {
+        text: "Hold"
```
PascalPrecht, line 45: This needs to be internationalized.
In fact, I'm not sure the text in the component combined with the token component works across languages

src/StatusQ/Controls/StatusTokenInlineSelector.qml
```diff
@@ -0,0 +1,170 @@
...
+    Repeater {
+        model: Math.max(0, root.tokens.length - 1)
+        delegate: Loader {
+            sourceComponent: tokenComponent
+            onLoaded: d.assignItemProperties(item, index, root.tokens[index])
+        }
+    }
+
+    StatusBaseText {
+        text: "or"
```
PascalPrecht, line 59: Same here

src/StatusQ/Controls/StatusTokenInlineSelector.qml
```diff
@@ -0,0 +1,170 @@
...
+        active: root.tokens.length > 1
+        sourceComponent: tokenComponent
+        onLoaded: {
+            const index = root.tokens.length - 1
+            d.assignItemProperties(item, index, root.tokens[index])
+        }
+    }
+
+    StatusBaseText {
+        text:  "to post"
```
PascalPrecht, line 74: And here

PascalPrecht (commented):
This looks nice!
Just a few minor comments regarding i18n.

Also, I think we need to double check the colors in darktheme. The SNT token doesn't look right to me (in terms of contrast ratio). Can you confirm this is correct?
Screenshot 2022-03-23 at 07 36 17

osmaczko:
> This looks nice!

Thanks!

>  Just a few minor comments regarding i18n.

Unfortunately, we do not translate strings in StatusQ, because of the [issue with the translation process](https://github.com/status-im/status-desktop/issues/4544) we have in status-desktop. I don't see much sense in exposing these texts as properties (so they could be overridden and translated by status-desktop). I will use qsTr() here. Mentioned issue is assigned to this iteration and the outcome would be `add qsTr() for all the strings in StatusQ` anyway. What do you think?

> Also, I think we need to double check the colors in darktheme. The SNT token doesn't look right to me (in terms of contrast ratio). Can you confirm this is correct? 

It does not look good to me as well. I reused the same colors as used in `StatusButton`:
image
There is a subtle difference though, it seems colors are blended differently because of different backgrounds, let me see what I can do about it.

PascalPrecht:
>  What do you think?

Sounds very good to me!

> There is a subtle difference though, it seems colors are blended differently because of different backgrounds, let me see what I can do about it.

Correct, the button colors have been designed to work well on certain backgrounds. It just wasn't foreseen that those colors will be used inside of an input. So what I suggest doing is clarify with the designer what colors should be used (most likely a different alpha value). Also, this is assuming that these inline tokens will then exclusively used inside of such an input where the  a certain background color can be expected.

osmaczko:
> Also, this is assuming that these inline tokens will then exclusively used inside of such an input where the a certain background color can be expected.

Maybe for such cases we could use an opaque background to blend with. That would not require any assumptions, and component may be used everywhere with that same color output.

Token.qml
```
Rectangle {
  color: Theme.palette.opaqueBackgroundColor
  
  Rectangle {
     id: background
     anchors.fill: parent
     color: Theme.palette.primaryColor1 // nonopaque
  }
}
```

PascalPrecht:
Yeap, can do that too. Will just have the same side effect that it probably won't behave "as expected" outside of an input. The reason being that in the case of alpha channel colors, they are supposed to behave differently given certain backgrounds. 

So while we can give it a solid color instead, it might still not look as you would want it to look when you use it outside of an input (or whatever the context it was designed for). This is hard to get right and I think the compromise will always be to have dedicated colors for certain contexts..

osmaczko:
The real issue was the color assignment instead of binding, making colors not adapt themselves to the theme. This is how it looks now:
image
image

src/StatusQ/Controls/StatusTokenInlineSelector.qml
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    StatusBaseText {
+        text: qsTr("to post")
+        color: Theme.palette.directColor6
+        font.pixelSize: 15
+    }
+
+    QtObject {
```
noeliaSD, line 79: Just a question here, just to agree with coding conventions... I use to create that `QtObject` which is used to create "private" things here (this is qt's proposed order, so I write it after the functions definition, since normally this object contains properties and functions that will be used in the following lines and they are also properties and functions, but we can't expose them as an interface):

- id
- property declarations
- signal declarations
- JavaScript functions
- -->> PRIVATE STUFF DEFINITION <<--
- object properties
- child objects
- states
- transitions 

WDYT @osmaczko ?

noeliaSD:
I'm late for the review but it looks nice!! :)

src/StatusQ/Controls/StatusTokenInlineSelector.qml
osmaczko: Yeah, it's a matter of agreement on coding conventions, which we need to define, see: https://github.com/status-im/StatusQ/pull/594#discussion_r832579471

I don't have a strong preference, I tend to put private stuff/details at the end.

noeliaSD: > Yeah, it's a matter of agreement on coding conventions, that we need to define, see: [#594 (comment)](https://github.com/status-im/StatusQ/pull/594#discussion_r832579471)
> 
> I don't have a strong preference, I tend to put private stuff/details at the end.
Ok! I will check it there.. still not reviewed!!

PR #594, opened 2022-03-22
doc(General): initial proposal on developer documentation and checklists

stefandunca:
# Scope

Developer targeted checklist and reminders to cover the most common source of errors and tech-debt (maintenance burden).

In environments like UX and UI, it is impossible to consider and follow the implications of all the variables and dependencies. This PR attempts to create checklists that guide QML/Qt developers in their work.

# Future

The plan is to add reminders and entries for the common source of oversight found in development which hopefully will help newcomers and seasoned `StatusQ` developers.
If the page/file grows too much, it can be split into sections and addressed in other files by link indirection.

### DIFF
--- a/src/contributing.md
+++ b/src/contributing.md
@@ -1,0 +1,156 @@
+# Contributing guidelines
+
+## Generic Reusable Component Design
+
+Define the component's purpose. Narrow down the scope as much as possible.
+
+Avoid assumptions about its parent/owner
+
+- See [broken behavior example](./code-examples.md#Dont-assume-parents-context-bad-example)
+- See [well-behaved behavior example](./code-examples.md#Give-user-choice)
+
+Define and make clear the interface/contract that defines requirements.
+
+To simplify maintenance consider the following options for implementing custom controls, generic containers or views:
+
+- Use QML2's customization mechanics in place if the change is simple: [Check docs for examples](https://doc.qt.io/qt-6/qtquickcontrols2-customize.html)
+- If the control should be generic or the changes are extensive and the design maps to a QML control, use QML2's customization mechanisms in a generic QML control: [Check docs for examples](https://doc.qt.io/qt-6/qtquickcontrols2-customize.html) instead of starting from scratch.
+- Use Qt controls instead of defining our own. Defining and implementing a complex control from scratch should be the last resort and there should be an excellent reason to do that.
+- For keyboard input item, ensure focus properties are set; see [Qt docs](https://doc.qt.io/qt-6/qtquick-input-focus.html)
+
+When unsure, check Qt's excellent documentation.
+
+- For main components/controls, there is a good overview that is worth revisiting from time to time
+- Functions/Properties usually have a short on-point description
+- In `QtCreator` you can quickly open the doc panel using `F1` key while having a cursor on component or function/property and recheck its invariants, pre and post conditions
+
+Consider that design follows user-friendlier principles.
+
+- Have commonly used items at hand
+- Have transition if possible
+  - Use Qt's property animation for states
+- Avoid often and radical size changes to items in views/layouts. Radical size change confuses users.
+  - If the content is not available, consider having placeholders using the estimated size 
+  (delegates, dummy items)
+
+Have the base control as the component's root. This way, control inherits all the interface and reduce the code.
+If it doesn't map to an existing one, use a base control like `Item`.
+
+- Don't use layouts or positioners as base for controls, they don't have the same behavior when used in layouts/containers
+  - Layouts have `fillWidth`/`fillHeight` as true by default and they will be extended. Controls don't and they will follow implicit sizes
+
+## QML well-behaved components checklist
+
+### Sizing support
+
+[Positioners and Layouts In QML](https://doc.qt.io/qt-6/qtquick-usecase-layouts.html)
+
+- Define size hints appropriately. Define implicit properties (`impicitWidth` and `implicitHeight`)
+  - They are used to break polishing binding loops for adaptive control
+  - All the containers use the implicit size for deriving the initial size (Layouts, Popups, Delegates, GridView, ListView ...). Size hints combined with resize adaptability are the appropriate way to have reactive controls.
+  - For complex controls, look if layouts are the choice as the main position driver. [Layouts](###Layouts)
+
+  ```qml
+  Item {
+    id: root
+    implicitWidth: mainLayout.implicitWidth
+    implicitHeight: mainLayout.implicitHeight
+    RowLayout {	// Column, Grid
+       id: mainLayout
+       // ...
+    }
+  ```
+
+- Adapt to the requested size if control can scale or it make sense to be extensible. For sizes bigger than content follow QML way for similar controls. e.g. Text leaves empty space around content
+- If the control is not adaptable and makes sense only in its full size, define default sizes and make it work by default with positioners (`Row`, `Column`, `Grid`).
+  - [Item Positioners in QML](https://doc.qt.io/qt-5/qtquick-positioning-layouts.html)
+- Don't mix hints with sizes, it will create binding loops
+
+## Topics
+
+### QML Coding Conventions
+
+[Follow Qt's way if appropriate](https://doc.qt.io/qt-6/qml-codingconventions.html)
+
+### Layouts
+
+[Qt Quick Layouts Overview](https://doc.qt.io/qt-6/qtquicklayouts-overview.html)
+
+Hierarchically sources hints from children. Implicit properties will have the recommended aggregated size.
+
+Layouts as children of other layouts have `fillWidth` and `fillHeight` enabled; controls don't.
+
+Use `Layout.preferredWidth` and `Layout.preferredHeight` attached properties to overwrite item's `implicitWidth` if they are not appropriate to the design.
+
+### Scope
+
+Follow [Qt's recommendations](https://doc.qt.io/qt-5/qtqml-documents-scope.html) if appropriate
+
+- Consider dynamic scoping. Component `id`s are not accessible outside the component's scope. If required, the component can expose the instance through properties binding. E.g. `readonly property ComponentType exposedComponent: privateInstanceId`
+  - Example
+
+    ```qml
+    Item {
+      // Probably won't work as intended. If another `testComp` instance is defined in QML's document hierarchy model and has a `testProp` property, that will be sourced instead
+      property bool booleanProp: testComp.testProp
+      // Same behavior if the TestComponent is defined in a file TestComponent.qml
+      component TestComponent: Item {
+        id: testComp
+        property bool testProp: false
+      }
+    }
+    ```
+
+- If in doubt, explicitly use an instance variable to access properties
+- If the scope is clear and there is no ambiguity, use the property directly for readability
+
+## Testing
+
+Test in isolation
+
+- Use `qmlproject`s and `qml` for quick complex scenarios
+- Use [`QtQuick.TestCase`](https://doc.qt.io/qt-5/qtquicktest-index.html)
+
+Integration tests
+
+- Use sandbox test app
+- Use QML `qmlproject`s and `qmlscene` for quick debugging with layouts
+
+Try scenarios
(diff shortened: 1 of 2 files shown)


src/readme.md
```diff
@@ -0,0 +1,140 @@
+# Developer notes
```
alaibe, line 1: I think this file is generally named CONTRIBUTING.md on github shall we just use that?

alaibe (approved):

src/readme.md
stefandunca: Done, thanks.

saledjenic (approved):

src/contributing.md
```diff
@@ -0,0 +1,140 @@
+# Contributing guidelines
+
+## Generic Reusable Component Design
+
+Define the component's purpose. Narrow down the scope as much as possible.
+
+Avoid assumptions about its parent/owner.
+
+Define the interface/contract. Make explicit what characteristics can be set by the owner and which are read-only or undefined behavior. (prefix with `_` private internal properties?)
```
jrainville, line 9: Since we never did the `_` prefix before, maybe it's not worth adding here.

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+
+If control maps to a QML control, use QML2's customization mechanisms. [Check docs for examples](https://doc.qt.io/qt-6/qtquickcontrols2-customize.html)
+
+Ensure proper Keyboard support and focus-scope handling.
```
jrainville, line 13: I agree that keyboard support is very important and we should have it, but we never did before, so I wonder if maybe we should add some kind of baseline for it before advocating for it (to potential new contributors)

jrainville (approved):

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+       // ...
+    }
+  ```
+
+- Define a default size to work with positioners (`Row`, `Column`, `Grid`). If appropriate, source them from implicit dimensions and expect the owner to overwrite them.
+  - Don't mix hints with sizes, it will create binding loops
+- Adapt to the requested size.
+Have the base control as the component's root. This way, control inherits all the interface and reduce the code.
+If it doesn't map to an existing one, use `Item` (never use layout or positioners as base control, they don't behave as expected)
+Have public properties at the top and make them explicit
```
osmaczko, line 58: I think it should land in a separate section named "QML Coding Conventions". We should probably start with [Qt recommended ones](https://doc.qt.io/qt-5/qml-codingconventions.html) and adapt/extend it to our needs iteratively.

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+       id: mainLayout
+       // ...
+    }
+  ```
+
+- Define a default size to work with positioners (`Row`, `Column`, `Grid`). If appropriate, source them from implicit dimensions and expect the owner to overwrite them.
+  - Don't mix hints with sizes, it will create binding loops
+- Adapt to the requested size.
+Have the base control as the component's root. This way, control inherits all the interface and reduce the code.
+If it doesn't map to an existing one, use `Item` (never use layout or positioners as base control, they don't behave as expected)
```
osmaczko, line 57: > never use layout or positioners as base control, they don't behave as expected

That's interesting, could you provide clarification or an example (case I used to do that [before](https://github.com/status-im/StatusQ/blob/c951c8e6f9425af78400be8c917c3d310457e2a1/src/StatusQ/Controls/StatusTokenInlineSelector.qml))?

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+
+Hierarchically sources hints from children. Implicit properties will have the recommended aggregated size.
+
+Layouts as children of other layouts have `fillWidth` and `fillHeight` enabled; controls don't.
+
+Use `Layout.prefferedWidth` and `Layout.prefferedHeight` attached properties to overwrite item's implicitWidth if they are not appropriate to the design.
+
+### Scope
+
+Component 'id's are not accessible outside component's scope
```
osmaczko, line 72: Maybe it would be worth mentioning the term  `dynamic scoping` here.

> Dynamic scoping is very powerful, but it must be used cautiously to prevent the behavior of QML code from becoming difficult to predict. In general it should only be used in cases where the two components are already tightly coupled in another way. When building reusable components, it is preferable to use property interfaces.
https://doc.qt.io/qt-5/qtqml-documents-scope.html#component-instance-hierarchy

src/contributing.md
```diff
@@ -0,0 +1,140 @@
+# Contributing guidelines
+
+## Generic Reusable Component Design
+
+Define the component's purpose. Narrow down the scope as much as possible.
+
+Avoid assumptions about its parent/owner.
```
osmaczko, line 7: Maybe we could provide an example:

SomeItem.qml
```qml
Item {
  Layout.fillWidth: true
  Layout.fillHeight: true
}
```
If `SomeItem` is not used in layout it would end up with warnings. I've seen it in some places in our code already.

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+  ```
+
+- If in doubt, explicitly use an instance variable to access properties
+- If the scope is clear and there is no ambiguity, use directly the property for readability
+
+## Testing
+
+Test in isolation
+
+- Use `qmlproject`s for quick complex scenarios
```
osmaczko, line 95: What is `qmlproject`?

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+- Use [`QtQuick.TestCase`](https://doc.qt.io/qt-5/qtquicktest-index.html)
+
+Integration tests
+
+- Use sandbox test app
+- Use QML `qmlproject`s for quick debugging with layouts
+
+Try scenarios
+
+- Embed in Layouts with different properties: fixed size, min-max, not size set
```
osmaczko, line 105: I like this approach, maybe we should make it a part of the sandbox test app? We can even try to generalize it, so that the required code would be:
```
TestItemInLayout {
   componentSrc: "TestControl.qml"
}
```

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+    id: root
+    implicitWidth: mainLayout.implicitWidth
+    implicitHeight: mainLayout.implicitHeight
+    RowLayout {	// Column, Grid
+       id: mainLayout
+       // ...
+    }
+  ```
+
+- Define a default size to work with positioners (`Row`, `Column`, `Grid`). If appropriate, source them from implicit dimensions and expect the owner to overwrite them.
```
osmaczko, line 53: Shouldn't we prefer Layouts over positioners? Layouts are adaptive, positioners are not.

osmaczko (approved):
Great work! :+1:

src/contributing.md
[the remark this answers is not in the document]
alexandraB99: yes, please no underscores :D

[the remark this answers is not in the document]
noeliaSD: I also use @osmaczko approach. [QML Convention from Qt - Private properties](https://doc.qt.io/archives/qt-4.8/qml-coding-conventions.html) proposes it as well.

src/contributing.md
noeliaSD: Just to agree on convention: https://github.com/status-im/StatusQ/pull/593/files/9c8d2f16e2bd03bc835e3903d28e787ae629c651#r833172954

noeliaSD (approved):
@stefandunca great job!

src/contributing.md
stefandunca: Agree. I changed to state the importance of having a clear interface.

src/contributing.md
stefandunca: I removed it and keep it for a later discussion.

src/contributing.md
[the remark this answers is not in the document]
stefandunca: Thanks for the heads up. I agree that it doesn't bring any enforcement, therefore there is no reason for it outside little readability, a habit from past projects.

Not sure about the `d` approach. I haven't seen forcing good behavior make a considerable difference as a drawback to performance and code readability. Will love to see this in a follow-up PR.

Also having comments that guard the public-private sections is also a quick win and doesn't require any redirection or decoration. I removed this note and add to my list for future discussion.

src/contributing.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Generic Reusable Component Design
+
+Define the component's purpose. Narrow down the scope as much as possible.
+
+Avoid assumptions about its parent/owner.
+
+Define the interface/contract. Make explicit what characteristics can be set by the owner and which are read-only or undefined behavior. (prefix with `_` private internal properties?)
+
+If control maps to a QML control, use QML2's customization mechanisms. [Check docs for examples](https://doc.qt.io/qt-6/qtquickcontrols2-customize.html)
```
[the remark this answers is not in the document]
stefandunca: I meant `B`: creating a custom control should be first considered using the QML2's way as it can be seen in the documentation from the link provided. That means replacing defaults with custom items like markers, popups, backgrounds, etc.

This recommendation advocates against starting to build one from scratch without considering the alternatives. Most of the time the cost of creating complex controls and the effort to cover all the corner cases is high. QML2 provides the basics, and make sense to use it as long as the design follows the basic rules enforced by the QML's mechanism.

> we should use Qt controls instead of defining our own unless we have a very good reason to do that.
I integrated this great suggestion, thanks.

src/contributing.md
stefandunca: > I think it should land in a separate section named "QML Coding Conventions". We should probably start with [Qt recommended ones](https://doc.qt.io/qt-5/qml-codingconventions.html) and adapt/extend it to our needs iteratively.

Done, please check the new structure

> Just to agree on convention: https://github.com/status-im/StatusQ/pull/593/files/9c8d2f16e2bd03bc835e3903d28e787ae629c651#r833172954

Great suggestion, will love to see it in a follow-up PR.

src/contributing.md
stefandunca: I mentioned differences between controls and layouts as the default behavior of `Layout.fill*` attached properties.

src/contributing.md
stefandunca: Done, thanks for the suggestion

src/contributing.md
stefandunca: I provided examples in a separate file to keep this as a practical quick checklist and added redirection links, please have a look at the new structure.

src/contributing.md
stefandunca: They are QML only projects and offer a convenient way to quickly import QML modules, index all `qml` and `js` files. See https://doc.qt.io/qtcreator/quick-projects.html and https://doc.qt.io/archives/qt-5.12/qtquick-views-views-qmlproject.html for a typical example

src/contributing.md
stefandunca: Great idea, let's bring this experiment.

src/contributing.md
stefandunca: There are cases where Positioners provide a quick alternative and their characteristics should be considered. I updated this part, hope it brings more clarity.

stefandunca:
> @stefandunca great job!

Thanks for the great input and support. Feel free to propose improvements and corrections.

jrainville (approved):
It all looks good to me!

src/contributing.md
```diff
@@ -0,0 +1,156 @@
...
+
+Layouts as children of other layouts have `fillWidth` and `fillHeight` enabled; controls don't.
+
+Use `Layout.preferredWidth` and `Layout.preferredHeight` attached properties to overwrite item's `implicitWidth` if they are not appropriate to the design.
+
+### Scope
+
+Follow [Qt's recommendations](https://doc.qt.io/qt-5/qtqml-documents-scope.html) if appropriate
+
+- Consider dynamic scoping. Component `id`s are not accessible outside the component's scope. If required, the component can expose the instance through properties binding. E.g. `readonly property ComponentType exposedComponent: privateInstanceId`
```
osmaczko, line 89: Hm, what I meant is that we should **avoid** dynamic scoping. It makes QML code difficult to predict.

PR #599, opened 2022-03-24
feat(StatusSeedPhraseInput): exposed keys on pressed signal

alexandraB99:
Needed for task https://github.com/status-im/status-desktop/issues/4955

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Controls/StatusSeedPhraseInput.qml
+++ b/src/StatusQ/Controls/StatusSeedPhraseInput.qml
@@ -7,6 +7,42 @@ import StatusQ.Popups 0.1
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 
+/*!
+   \qmltype StatusSeedPhraseInput
+   \inherits Item
+   \inqmlmodule StatusQ.Controls
+   \since StatusQ.Controls 0.1
+   \brief Displays a text input with suggestions. Inherits \l{https://doc.qt.io/qt-5/qml-qtquick-item.html}{Item}.
+
+   The \c StatusSeedPhraseInput item displays a text input with suggestions filtered based on the typed text.
+   For example:
+
+   \qml
+       StatusSeedPhraseInput {
+           id: statusSeedInput
+           anchors.left: parent.left
+           anchors.right: parent.right
+           textEdit.input.anchors.leftMargin: 16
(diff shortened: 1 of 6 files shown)


src/StatusQ/Controls/StatusSeedPhraseInput.qml
```diff
@@ -17,6 +17,7 @@ Item {
     property ListModel inputList: ListModel { }
     property ListModel filteredList: ListModel { }
     signal doneInsertingWord(string word)
+    signal keyPressed(var event)
```
PascalPrecht, line 20: I might be mistaking but this looks like an addition of a new API to me no?
Or did `StatusSeedPhraseInput` have an `onKeyPressed` before?
If not, the component should have documentation for this new handler.

Also, if the above is true, this looks like a feature to me rather than a fix...
Wdyt @alexandraB99

alexandraB99: good question, this is not actually part of the StatusSeedPhraseInput but starts from StatusBaseInput, passes on to StatusInput and then comes here so that TextEdit key events can be catched. In that case, it's not really a StatusSeedPhraseInput feature but just StatusSeedPhraseInput trying to access TextEdit (so many layers in between). I agree that some documentation is needed, but currently none of the 3 components I mentioned has any documentation, so I would see documenting those (as all the rest StatusQ components) as separate tasks. I remember it was mentioned that issues regarding this would be created and assigned accordingly but I am not sure what is currently the status there.

PascalPrecht: Hm.. okay..I'd say, if `onKeyPressed` didn't exist on `StatusSeedPhraseInput` before this commit, that this is a new API. But you can leave it as is..

> I agree that some documentation is needed, but currently none of the 3 components I mentioned has any documentation,

That's why I believe it's good to start with it. I didn't mean you should document the entire component, just the new feature you've added. But if you say this isn't actually a feature then it's fine.

PascalPrecht (approved):

src/StatusQ/Controls/StatusSeedPhraseInput.qml
alexandraB99: documentation added only for StatusSeedPhraseInput and only for its keyPressed signal.

PR #603, opened 2022-03-28
fix(StatusChatListCategory): Selecting the category menu should not toggle the category itself

noeliaSD:
Closes `status-desktop` issue https://github.com/status-im/status-desktop/issues/4914

### DIFF
--- a/src/StatusQ/Components/StatusChatListCategory.qml
+++ b/src/StatusQ/Components/StatusChatListCategory.qml
@@ -48,7 +48,6 @@ Column {
                     popupMenuSlot.item.popup(mouse.x + 4, mouse.y + 6);
                     return
                 }
-                statusChatListCategory.opened = !opened;
             }
         }
         onTitleClicked: statusChatListCategory.opened = !opened


src/StatusQ/Components/StatusChatListCategory.qml
```diff
@@ -48,7 +48,6 @@ Column {
                     popupMenuSlot.item.popup(mouse.x + 4, mouse.y + 6);
                     return
                 }
-                statusChatListCategory.opened = !opened;
```
PascalPrecht, line 51: So this seems off to me... I read the original issue and it indeed says that "selecting" a category shouldn't toggle it. However, the attached video shows how clicking the "more" button in the category causes it to toggle (which is indeed unexpected)

However, clicking the category itself should most likely still toggle the category

noeliaSD: What I understand is that it should expand / collapse, if user clicks on:
- `Category title`.
- `>` button.

And do nothing (in relation to expand / collapse) if user clicks on:
- `+` button.
- `...` button.

Please @iurimatias confirm!

PascalPrecht: That's what I think how it should work as well!
And it seems your PR does that. Was just confused by the original issue title

noeliaSD: So, do you @PascalPrecht approve it? ;)

PascalPrecht: Ah sorry yes!

PR #645, opened 2022-05-04
fix(StatusListItem): Fix title text out of the expected boundaries

MishkaRogachev:
Tiny fix to keep text always inside boundaries

listitemtooltip-2022-05-09_14 08 43

### Checklist

- [x] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusListItem.qml
+++ b/src/StatusQ/Components/StatusListItem.qml
@@ -14,13 +14,13 @@ Rectangle {
     property string title: ""
     property string titleAsideText: ""
     property string subTitle: ""
-    property string tertiaryTitle: ""    
+    property string tertiaryTitle: ""
     property string label: ""
     property real leftPadding: 16
     property real rightPadding: 16
     property bool enabled: true
     property bool highlighted: false
-    property bool propagateTitleClicks: true 
+    property bool propagateTitleClicks: true
     property int type: StatusListItem.Type.Primary
     property list<Item> components
     property var bottomModel: []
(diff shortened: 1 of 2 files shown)


PascalPrecht (approved):

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -171,7 +171,7 @@ Rectangle {
                 height: visible ? contentHeight : 0
                 wrapMode: Text.WrapAtWordBoundaryOrAnywhere
                 anchors.left: parent.left
-                anchors.right: !statusListItem.titleAsideText && !titleIconsRow.sourceComponent ? parent.right : undefined
+                anchors.right: parent.right
```
noeliaSD, line 174: By doing this fix we are not following the design specs we need. We need that `titleAsideText` and `titleIconsRow` are positioned right next to the text. See the following [PR](https://github.com/status-im/StatusQ/pull/539#discussion_r791840349) where you can see a screenshot of a specific design.

Example to clarify the needed behavior:

- We want this:

Screenshot 2022-05-04 at 14 08 42

- By the current fix we achive this:

Screenshot 2022-05-04 at 14 09 24

noeliaSD (changes requested):

src/StatusQ/Components/StatusListItem.qml
PascalPrecht: Good catch @noeliaSD !

MishkaRogachev: Oh, its a bit more complicated i've expected :) So if the text is longer than we can place on item, what should we do?
Crop text by clipping or elide text with dots?

PascalPrecht: I'd say yes, but looping in @John-44 to be sure

noeliaSD: Yes, it is not easy! ;)

What I would suggest is:
- If we configure multiline, it just continues the writting in next lines.
- If we configure as single line, it writes dots.

WDYT @PascalPrecht ?

PascalPrecht: Sure, that works too, if we can make this configurable, even better

MishkaRogachev: image

MishkaRogachev: image

MishkaRogachev: I've removed multiline property and added tooltip

PR #652, opened 2022-05-05
fix(StatusBaseButton): After loading, button keeps loader width, so the button size is bigger

noeliaSD:
When button changed from loading to their normal shape it kept the width of the loader as it was visible in the row component.

It has been added a binding between loader `visible` property and root `loading` property that fixes the issue with the width.

Also, removed some `break` leftovers.

Fixes #5606 --> https://github.com/status-im/status-desktop/issues/5606

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Controls/StatusBaseButton.qml
+++ b/src/StatusQ/Controls/StatusBaseButton.qml
@@ -40,11 +40,9 @@ Rectangle {
               return 5
             case StatusBaseButton.Size.Small:
               return 10
-              break;
             case StatusBaseButton.Size.Large:
             default:
               return 11
-              break;
         }
     }
     property real defaultBottomPadding: {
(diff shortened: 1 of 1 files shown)


PascalPrecht:
@noeliaSD Just to clarify: The original issue was about a button keeping its original size, even when switching to loading state.
I might be wrong, but judging by the commit message and the changes, I'm not sure if this commit really addresses this.

Does this commit do the right thing?

PascalPrecht:
I understand!
So coming back to the original issue then, which is:

> Buttons with longer labels change size when entering loading state

This is not tackled here right and will then probably be done in a different PR?
The issue linked in this PR in fact brings up the exact place where this happened... a button with "Finalize password creation" label (which is very long)

PR #658, opened 2022-05-09
fix(StatusTagSelector): Added up/down hard keys navigation

alexandraB99:
Needed for: https://github.com/status-im/status-desktop/issues/5640

### Checklist

- [x] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusTagSelector.qml
+++ b/src/StatusQ/Components/StatusTagSelector.qml
@@ -235,7 +235,7 @@ Item {
                 }
             }
 
-            TextEdit {
+            TextInput {
                 id: edit
                 Layout.fillWidth: true
                 Layout.preferredHeight: 44
@@ -244,7 +244,7 @@ Item {
                 enabled: visible
                 focus: true
                 font.pixelSize: 15
-                wrapMode: TextEdit.WrapAnywhere
+                wrapMode: TextEdit.NoWrap
                 font.family: Theme.palette.baseFont.name
                 color: Theme.palette.directColor1
                 Keys.onPressed: {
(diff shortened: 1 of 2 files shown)


src/StatusQ/Components/StatusTagSelector.qml
```diff
@@ -254,7 +254,13 @@ Item {
                         removeMember(namesModel.get(namesList.count-1).publicId);
                         namesModel.remove((namesList.count-1), 1);
                     }
+                    if ((event.key === Qt.Key_Return || event.key === Qt.Key_Enter) && (sortedList.count > 0)) {
+                        root.insertTag(sortedList.get(userListView.currentIndex).name, sortedList.get(userListView.currentIndex).publicId);
+                    }
                 }
+                Keys.onUpPressed: { userListView.decrementCurrentIndex(); }
+                Keys.onDownPressed: { userListView.incrementCurrentIndex(); }
+                Component.onCompleted: { forceActiveFocus(); }
```
PascalPrecht, line 263: Just one thought here: Do we really want this component to force active focus? Shouldn't the application decide which controls will need active focus in their views?

PascalPrecht (approved):
Changes LGTM!
Just one comment

src/StatusQ/Components/StatusTagSelector.qml
alexandraB99: Not really, and it's not specified either. I added it because I found it annoying to have to click on it every time in order to type, since the only point that initiates action here is the text edit, I thought adding it. But I could also remove it :) wdyt?

alexandraB99: done, removed from the component and added in the instance where it's used :)

PR #664, opened 2022-05-11
feat(@StatusAppNavBar): Add qr scan button

alaibe:
image


PascalPrecht (approved):

src/StatusQ/Layout/StatusAppNavBar.qml
```diff
@@ -162,6 +169,16 @@ Rectangle {
         model: navBarModel
     }
 
+    Item {
```
alexandraB99, line 172: Could you use a `Loader` instead?

alaibe: This is same technique as `navBarProfileButtonSlot` I guess that is more logical to keep same logic

PR #665, opened 2022-05-11
fix(StatusSlider): improve handle draging area

stefandunca:
Issues
- The slider only handles mouse events in its own control space.
- To use the entire area it needs to be centred.

Fixes

- Center the content vertically (handle and slider bar)
- Height and implicitHeight defaults to heigh of the handle
- Add a note to create awareness about the mouse handling requirements

### Checklist

- [x] update sandbox app (checked)


PascalPrecht (approved):

stefandunca:
@anastasiyaig, this PR fixes slider grabbing, and it is already integrated into the original PR [#5631](https://github.com/status-im/status-desktop/pull/5631)
When you get back to testing PR [#5631](https://github.com/status-im/status-desktop/pull/5631) please have a quick regression over the sliders in our app to be sure that none was relying on the align progress line at top previous implementation that might break some layouts

src/StatusQ/Controls/StatusSlider.qml
```diff
@@ -6,19 +6,28 @@ import StatusQ.Core 0.1
...
     leftPadding: 0
 
     background: Rectangle {
         id: bgRect
+
+        x: statusSlider.leftPadding
+        y: statusSlider.topPadding + statusSlider.availableHeight / 2 - height / 2
         implicitWidth: 100
         implicitHeight: 4
+        width: statusSlider.availableWidth
```
alexandraB99, line 28: does it make sense to keep both `implicitWidth` and `width` set differently?

stefandunca: Thanks for the heads up, I removed it.
I followed [Qt's example](https://doc.qt.io/qt-5/qtquickcontrols2-customize.html#customizing-slider), but I agree that it is not needed given the `Control`'s background requirements: `"Note: If the background item has no explicit size specified, it automatically follows the control's size. In most cases, there is no need to specify width or height for a background item."`

PR #678, opened 2022-05-17
fix(StatusTagSelector): Fix mouse hover

borismelnik:
Closes: https://github.com/status-im/status-desktop/issues/5220

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusTagSelector.qml
+++ b/src/StatusQ/Components/StatusTagSelector.qml
@@ -355,8 +355,8 @@ Item {
                     anchors.rightMargin: 8
                     anchors.leftMargin: 8
                     radius: 8
-                    visible: (root.sortedList.count > 0)
-                    color: (userListView.currentIndex === index) ? Theme.palette.baseColor2 : "transparent"
+                    visible: wrapper.ListView.isCurrentItem
+                    color: Theme.palette.baseColor2
                 }
 
                 StatusSmartIdenticon {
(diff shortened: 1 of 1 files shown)


src/StatusQ/Components/StatusTagSelector.qml
```diff
@@ -355,8 +355,7 @@ Item {
                     anchors.rightMargin: 8
                     anchors.leftMargin: 8
                     radius: 8
-                    visible: (root.sortedList.count > 0)
-                    color: (userListView.currentIndex === index) ? Theme.palette.baseColor2 : "transparent"
+                    color: sensor.containsMouse ? Theme.palette.baseColor2 : "transparent"
```
alexandraB99, line 358: hmm is this working when navigating with the up/down arrows? I'd rather set the current index to index when the mouse area 'contains mouse'

alexandraB99 (changes requested):

src/StatusQ/Components/StatusTagSelector.qml
stefandunca: This fixes the issue but I agree with @alexandraB99 that the complete fix should be:
```qml
  visible: wrapper.ListView.isCurrentItem
  color: Theme.palette.baseColor2
```

stefandunca: @borismelnik also the fix won't be completed without fixing `userListView.decrementCurrentIndex()` and `userListView.incrementCurrentIndex()` calls above. We use a visual workaround to hide elements instead of removing them from the data model. Therefore the increment/decrement calls don't take this into considerations and `currentIndex` can be one of the invisible items.

stefandunca (changes requested):

stefandunca (approved):

alexandraB99 (approved):
cool!

PR #681, opened 2022-05-18
feat(StatusModalFloatingButtonsSelector): Implementing the floating buttons selector widget

Khushboo-dev-cpp:
feat(StatusModalFloatingButtonsSelector): Implementing the floating accounts selector widget.

Needed under wallet design https://www.figma.com/file/FkFClTCYKf83RJWoifWgoX/Wallet-v2?node-id=3311%3A177552

### Checklist

- [x] update sandbox app

https://user-images.githubusercontent.com/60327365/169160811-804bee04-2b26-45ed-b082-422d3a3819c9.mov

### DIFF
--- a/src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
+++ b/src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
@@ -1,0 +1,133 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.14
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Popups 0.1
+
+/*!
+   \qmltype StatusModalFloatingButtonsSelector
+   \inherits Row
+   \inqmlmodule StatusQ.Controls
+   \since StatusQ.Controls 0.1
+   \brief The StatusModalFloatingButtonsSelector provides a template for creating a selectable buttons list
+   this list can be parially hidden and the rest of the items are shown under the more button in a popup
+
+   Example of how to use it:
+
+   \qml
+        StatusModalFloatingButtonsSelector {
+            id: floatingHeader
+            model: dummyAccountsModel
+            delegate: StatusAccountSelectorTag {
+                title: "name"
+                icon.name: "iconName"
+                isSelected: floatingHeader.currentIndex === index
+                visible: visibleIndices.includes(index)
+                onClicked: floatingHeader.currentIndex = index
+            }
+            popupMenuDelegate: StatusListItem {
+                implicitWidth: 272
+                title: "name"
+                onClicked: floatingHeader.itemSelected(index)
+                visible: !visibleIndices.includes(index)
+            }
+        }
+   \endqml
+
+   For a list of components available see StatusQ.
+*/
+Row {
+    id: floatingButtons
+
+    /*!
+       \qmlproperty delegate
+        This property represents the delegate of selectable items shown to the user.
+        Can be used to assign delegate to the buttons selector
+        \endqml
+    */
+    property alias delegate: itemSelectionRepeater.delegate
+    /*!
+       \qmlproperty popupMenuDelegate
+        This property represents the delegate of popupmenu fropm which items can be selected by the user.
+        Can be used to assign delegate to the popupmenu
+        \endqml
+    */
+    property alias popupMenuDelegate: popupMenuSelectionRepeater.delegate
+
+    /*!
+       \qmlproperty model
+        This property represents the model of selectable items shown to the user.
+        Can be used to assign selectable items in the buttons selector
+        \endqml
+    */
+    property var model
+    /*!
+       \qmlproperty visibleIndices
+        This property represents the indices from the selectable items that will visible to the user
+        Can be used to set visible items in the buttons selector
+        \endqml
+    */
+    property var visibleIndices: [0,1,2]
+    /*!
+       \qmlproperty currentIndex
+        This property represents the index of the currently selected item
+        Can be used to set the currnetly selected item in the buttons selector
+        \endqml
+    */
+    property int currentIndex: 0
+
+    function itemSelected(index) {
+        visibleIndices = [0,1,index]
+        floatingButtons.currentIndex = index
+        popupMenu.close()
+    }
+
+    height: 32
+    spacing: 12
+    clip: true
+
+    Repeater {
+        id: itemSelectionRepeater
+        model: floatingButtons.model
+    }
+
+    Rectangle {
+        width: button.width
+        height: button.height
+        radius: 8
+        visible: floatingButtons.model.count > 3
+        StatusButton {
+            id: button
+            implicitHeight: 32
+            topPadding: 8
+            bottomPadding: 0
+            defaultLeftPadding: 4
+            defaultRightPadding: 4
+            normalColor: "transparent"
+            icon.name: "more"
+            icon.background.color: "transparent"
+            onClicked: popupMenu.popup(parent.x, y + height + 8)
+        }
+    }
+
+    // Empty item to fill up empty space
+    Item {
+        Layout.preferredHeight: parent.height
+        Layout.fillWidth: true
+    }
+
+    StatusPopupMenu {
+        id: popupMenu
+        width: layout.width
+        ColumnLayout {
+            id: layout
+            Repeater {
+                id: popupMenuSelectionRepeater
+                model: floatingButtons.model
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/StatusQ/Components/StatusClickableTag.qml
```diff
@@ -0,0 +1,139 @@
+import QtQuick 2.13
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+/*!
+   \qmltype StatusAccountSelectorTag
```
PascalPrecht, line 9: Should this be `StatusClickableTag` ?

src/StatusQ/Components/StatusClickableTag.qml
```diff
@@ -0,0 +1,139 @@
...
+
+   \qml
+        StatusClickableTag {
+            title: "some name"
+            icon.name: "some icon name"
+            isSelected: false
+            onClicked: {
+                // do something
+            }
+        }
```
PascalPrecht, line 25: Maybe I'm wrong, but to me this looks like just another variation of `StatusButton`.
Or at least, the one question that comes up is: Why is this considered a "tag" and not a "button" ?

src/StatusQ/Components/StatusClickableTag.qml
```diff
@@ -0,0 +1,139 @@
...
+        Can be used to assign text to the tag
+        \endqml
+    */
+    property string title: ""
+    /*!
+       \qmlproperty isSelected
+        This property represents if the tag should be highlighted
+        \endqml
+    */
+    property  bool isSelected: false
```
PascalPrecht, line 52: We might as well call this `highlighted`? Also aligns more with what other buttons already do...

src/StatusQ/Components/StatusClickableTag.qml
```diff
@@ -0,0 +1,139 @@
...
+    /*!
+       \qmlproperty maxTextWidth
+        This property represents the max widht allowed for the title after which it will elide
+        \endqml
+    */
+    property int maxTextWidth: 93
```
PascalPrecht, line 58: Not that it makes a huge difference, but would it be okay to use some even value instead?

src/StatusQ/Components/StatusClickableTag.qml
```diff
@@ -0,0 +1,139 @@
...
+    /*!
+       \qmlproperty icon
+        This property represents the StatusIconSettings to be shown in the tag
+        Can be used to assign icon to the tag
+        \endqml
+    */
+    property StatusIconSettings icon: StatusIconSettings {
+        height: 20
+        width: 20
+        rotation: 0
```
PascalPrecht, line 79: I believe `rotation: 0` is the default, so this could be omitted.

src/StatusQ/Components/StatusClickableTag.qml
```diff
@@ -0,0 +1,139 @@
...
+    signal clicked()
+
+    width: layout.width
+    height: 32
+    color: Theme.palette.statusAppLayout.backgroundColor
+    radius: 8
+
+    Rectangle {
+        anchors.fill: parent
+        color: isSelected ? Theme.palette.primaryColor2 :"transparent"
```
PascalPrecht, line 99: This should have some hover indication as well

src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+import StatusQ.Core.Theme 0.1
+import StatusQ.Components 0.1
+import StatusQ.Controls 0.1
+import StatusQ.Popups 0.1
+
+/*!
+   \qmltype StatusModalFloatingButtonsSelector
+   \inherits Row
+   \inqmlmodule StatusQ.Controls
+   \since StatusQ.Components 0.1
```
PascalPrecht, line 14: This doesn't actually live in `StatusQ.Components`

src/StatusQ/Components/StatusClickableTag.qml
```diff
@@ -0,0 +1,139 @@
...
+import QtQuick.Layouts 1.12
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+/*!
+   \qmltype StatusAccountSelectorTag
+   \inherits Rectangle
+   \inqmlmodule StatusQ.Components
+   \since StatusQ.Components 0.1
```
PascalPrecht, line 12: Shouldn't this actually be in `Controls`? Cause essentially it's a form control..

src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+   \endqml
+
+   For a list of components available see StatusQ.
+*/
+Row {
+    id: floatingButtons
+
+    /*!
+       \qmlproperty delegate
+        This property represents the delgate of selectable items shown to the user.
```
PascalPrecht, line 47:
```suggestion
        This property represents the delegate of selectable items shown to the user.
```

src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+    */
+    property alias popupMenuDelegate: popupMenuSelectionRepeater.delegate
+
+    /*!
+       \qmlproperty itemsModel
+        This property represents the model of selectable items shown to the user.
+        Can be used to assign selectable items in the buttons selector
+        \endqml
+    */
+    property var itemsModel
```
PascalPrecht, line 66: Nit: shall we call this just `model`?

src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+    /*!
+       \qmlproperty visibleIndices
+        This property represents the indices from the selectable items that will visible to the user
+        Can be used to set visible items in the buttons selector
+        \endqml
+    */
+    property var visibleIndices: [0,1,2]
```
PascalPrecht, line 73: One thing to keep in mind here is that, depending on available viewport, this would have to change.

src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
```diff
@@ -0,0 +1,121 @@
...
+        id: itemSelectionRepeater
+        model: itemsModel
+    }
+
+    StatusClickableTag {
+        visible: itemsModel.count > 3
+        icon.name: "more"
+        icon.background.color: "transparent"
+        onClicked: popupMenu.popup(x, y + height + 8)
+    }
```
PascalPrecht, line 102: The usage of this thing makes it semantically even more a button than a tag :D (sorry)

PascalPrecht (changes requested):
Good stuff!

Just some minor comments and questions. Maybe some tiny requested changes

src/StatusQ/Components/StatusClickableTag.qml
PascalPrecht: One other thing I've noticed as well: The colors generally don't seem to match with the ones in the design. At least for the labels, probably also the background?

stefandunca, line 99: Wondering what is the advantage of this approach over changing the `visible` property? I'm thinking that `visible` is a better alternative and making colour "transparent" is a workaround for hiding it?
```suggestion
        color: Theme.palette.primaryColor2
        visible: isSelected
```

stefandunca (approved):

src/StatusQ/Components/StatusClickableTag.qml
Khushboo-dev-cpp: hmm good point, I think it could be a button just different look and feel

src/StatusQ/Components/StatusClickableTag.qml
Khushboo-dev-cpp: so the thing is I think the wallet designs are syncing 100% with other designs and I think it expected it will be synced once the component design system that David is working on will be completed

src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
Khushboo-dev-cpp: I think not, because the Modal width and height will remain constent

Khushboo-dev-cpp:
Hey @PascalPrecht , I removed the StatusClickableTag and reused the StatusButton instead.
I think the designs are inconsistent atm and it is dependent on the component design system. So for now, I have tried to use it with minimal changes

PR #687, opened 2022-05-23
feat(StatusLabeledSlider): add ability to decorate sliders and slider, decorated with labels

### LINKED ISSUE #686: Decoratable StatusSlider component
image
image

Component for https://github.com/status-im/status-desktop/issues/5809

MishkaRogachev:
Close #686

image

### Checklist

- [x] update sandbox app

### DIFF
--- a/sandbox/controls/Controls.qml
+++ b/sandbox/controls/Controls.qml
@@ -189,6 +189,16 @@ GridLayout {
         value: 40
     }
 
+    StatusLabeledSlider {
+        width: 360
+        model: [ qsTr("XS"), qsTr("S"), qsTr("M"), qsTr("L"), qsTr("XL"), qsTr("XXL")]
+    }
+
+    StatusLabeledSlider {
+        width: 360
+        model: [ qsTr("50%"), qsTr("100%"), qsTr("150%"), qsTr("200%")]
+    }
+
     StatusBanner {
         id: banner
         width: 360
@@ -198,7 +208,7 @@ GridLayout {
     }
 
     StatusProgressBar {
-        id: progressBar        
+        id: progressBar
         text: "Weak"
         value: 0.5
         fillColor : Theme.palette.pinColor1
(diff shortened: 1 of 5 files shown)


sandbox/controls/Controls.qml
```diff
@@ -189,6 +189,16 @@ GridLayout {
...
 
+    StatusLabeledSlider {
+        width: 360
+        model: [ qsTr("XS"), qsTr("S"), qsTr("M"), qsTr("L"), qsTr("XL"), qsTr("XXL")]
+    }
+
+    StatusLabeledSlider {
+        width: 360
+        model: [ qsTr("50%"), qsTr("100%"), qsTr("150%"), qsTr("200%")]
+    }
```
PascalPrecht, line 200: Can' `StatusSlider` also have a `model`?

And then instead of having `StatusLabeledSlider` and `StatusSlider`, we'd just have `StatusSlider` with a different variant? Both would be able to render labels, but the variant makes it look different.

WDYT?

MishkaRogachev: In future we probably would have more slider types e.g picking range or with some gauge with ticks. Decaration can be  different from this labels. So I decided to leave in StatusSlider just generic features, that can be specialized in child sliders.

PascalPrecht: Okay so... but that means if I want to use a `StatusSlider` (that looks like a `StatusSlider` and not a `StatusLabeledSlider`) and it should have labels (like it does in Desktop), I have to attach the labels manually somehow, even though we have a slider implementation that has a built-in API for that?

We can do it that way, just feels like this is something that wants to be reused..

MishkaRogachev: Hm, what about to have separated decaration instead like this?
```
StatusSlider {
    decoration: StatusDecarationLabeled { model: [...] }
}

StatusRangeSlider {
    decoration: StatusDecarationGauge { tickSize: ... }
}
```

PascalPrecht: Let's revisit this once it's actually becoming a problem and go with what you have so far. I don't want us to over engineer this either :D

noeliaSD (approved):
Nice improvement!! :)

sandbox/controls/Controls.qml
MishkaRogachev: So i can merge all chages to StatusSlider or you just approve this PR :)

PascalPrecht (approved):

PR #688, opened 2022-05-24
fix(StatusNavigationListItem): make click handler work again

PascalPrecht:
The click handler used inside the component broke in one of the latest
changes done in `StatusListItem` where it expects two arguments being
passed to its click handler.

This commit fixes it by supplying all needed arguments (where `itemId`
is always `''` unless set differently)


osmaczko (approved):

iurimatias:
spoke to alexandra, approved. merging this

Issue #692, opened 2022-05-25
Second iteration of `StatusCommunityCard` to fit mobile design


noeliaSD:
Agreed with @John-44 that the `Community Card` component shall look like the **mobile design** instead of the desktop one so in this task the `StatusCommunityCard` component should be modified to fit the mobil design:

![](https://cdn.discordapp.com/attachments/730410574428504166/979107336620736512/Screenshot_2022-05-25_at_19.10.57.png)

![](https://cdn.discordapp.com/attachments/730410574428504166/979107553772449792/Screenshot_2022-05-25_at_19.28.27.png)

*New figma design:* https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=8159%3A416159

*Desktop figma:* https://www.figma.com/file/17fc13UBFvInrLgNUKJJg5/Kuba%E2%8E%9CDesktop?node-id=2090%3A245814

PR #693, opened 2022-05-25
feat(StatusCommunityCard): First component iteration

noeliaSD:
Part of task: https://github.com/status-im/status-desktop/issues/4936

### What does the PR do
Normal `Community Card` includes:
- Logo
- Title
- Community members
- Description
- _Tags row*_ (that must be replaced to a new `StatusQ` component `StatusListItemTagRow`. It is out of the scope of this task so a raw implementation has been added and if the tags model is big, the component will be cut).

Loading `Community Card`.

Added support in sandbox (component pages and demo app view).

### Affected areas

- Components/StatusCommunityCard
- sandbox

Normal / contains mouse:
Screenshot 2022-05-26 at 08 05 58

Loading card:
Screenshot 2022-05-26 at 08 06 13

https://user-images.githubusercontent.com/97019400/170427666-1ce25e28-e3f0-4c9a-8a1d-d27dd7e54d63.mov

### Checklist

- [x] update sandbox app


alexandraB99:
haven't looked at figma, but spacing in sport - food - privacy (etc) badges looks a bit weird, too wide in the middle (between icon and word), too narrow at the end

src/StatusQ/Components/StatusCommunityCard.qml
```diff
@@ -0,0 +1,261 @@
...
+    // Loading Card
+    ColumnLayout {
+        visible: !root.loaded
+        anchors.fill: parent
+        anchors.margins: d.dMargins
+        clip: true
+        spacing: 4
+
+        // Filler
+        Item { height: 40}
```
alexandraB99, line 225: maybe better use `prefferedHeight` to be sure ;)

noeliaSD:
> haven't looked at figma, but spacing in sport - food - privacy (etc) badges looks a bit weird, too wide in the middle (between icon and word), too narrow at the end

Yeah, the `tags row` component will be addressed in a different task, creating a new `StatusQ` component `StatusListItemTagRow` and there it will be fixed this spacing issue and other things. `Tags row` is out of the scope of this first iteration but I added a quick / base implementation, if you are not convinced of showing it with this visual "defect", I can hidde it by the moment but I'd keep it and fix it later. WDYT?

The task that will address that is: https://github.com/status-im/status-desktop/issues/7560

noeliaSD:
Added `don't merge` tag bc I will update the `loading card` state to be more similar with the one that designers shared yesterday.  It will not be the same since second iteration of that component will be addressed in task https://github.com/status-im/StatusQ/issues/692 but to have a similar look.

noeliaSD:
Updated `loading card` similar to the new design but consistent with the current `normal card` design.

Screenshot 2022-05-26 at 08 06 13

alexandraB99:
> > haven't looked at figma, but spacing in sport - food - privacy (etc) badges looks a bit weird, too wide in the middle (between icon and word), too narrow at the end
> 
> Yeah, the `tags row` component will be addressed in a different task, creating a new `StatusQ` component `StatusListItemTagRow` and there it will be fixed this spacing issue and other things. `Tags row` is out of the scope of this first iteration but I added a quick / base implementation, if you are not convinced of showing it with this visual "defect", I can hidde it by the moment but I'd keep it and fix it later. WDYT?
> 
> The task that will address that is: status-im/status-desktop#7560

I can live with fixing it in status-im/status-desktop#7560 :)

alexandraB99 (commented):
I am getting: `sandbox/demoapp/StatusAppCommunitiesPortalView.qml:58: ReferenceError: Style is not defined`

noeliaSD:
> I am getting: `sandbox/demoapp/StatusAppCommunitiesPortalView.qml:58: ReferenceError: Style is not defined`

Solved! Thanks

alexandraB99 (approved):

src/StatusQ/Components/StatusCommunityCard.qml
```diff
@@ -0,0 +1,289 @@
...
+    /*!
+       \qmlproperty bool StatusCommunityCard::loaded
+       This property holds a boolean value that represents if the community information is loaded or not.
+    */
+    property bool loaded: true
+    /*!
+       \qmlproperty string StatusCommunityCard::logo
+       This property holds the community logo source.
+    */
+    property string logo: ""
```
PascalPrecht, line 53: Shouldn't this be of type `url` instead?

src/StatusQ/Components/StatusCommunityCard.qml
```diff
@@ -0,0 +1,289 @@
...
+            }
+       \endqml
+    */
+    property ListModel categories: ListModel {}
+    /*!
+       \qmlproperty string StatusCommunityCard::locale
+       This property holds the application locale used to give format to members number representation.
+       If not provided, default value is "en".
+    */
+    property string locale: "en"
```
PascalPrecht, line 95: Is this needed? I thought applications would be decide what the overall locale is? How would this behave if an application is set to `es` but this component receives (for whatever reason) `en`?

We don't have `locale` properties in any of the other components, so I'm wondering why this is needed now

src/StatusQ/Components/StatusCommunityCard.qml
```diff
@@ -0,0 +1,289 @@
...
+
+    /*!
+        \qmlsignal StatusCommunityCard::cardSelected(string communityId)
+        This signal is emitted when the card item is selected / clicked.
+    */
+    signal cardSelected(string communityId)
```
PascalPrecht, line 101: How do you feel about making this a `clicked` signal instead?
`selected` has a particular semantic meaning, more referred to "selecting" items (maybe even multiple ones, or at least fro multiple options), whereas `clicked` describes exactly what this is..

0x-r4bbit (commented):
Got some minor comments.

There was already discussions around the tag row so I'm not going to comment on that, other than that we'll have to ensure these tags will have proper hover indicators.

noeliaSD:
> Got some minor comments.
> 
> There was already discussions around the tag row so I'm not going to comment on that, other than that we'll have to ensure these tags will have proper hover indicators.

`Tags row` will be implemented / improved here: https://github.com/status-im/status-desktop/issues/7560

src/StatusQ/Components/StatusCommunityCard.qml
noeliaSD: Here we have 2 options:
- Exposing a more generic `string` property instead of assuming the `members` representation is a number, then the formatting of the members number is a complete responsability of the consumer.
- Exposing the `locale`, as it is currently implemented, then the formatting of the members number depends on this locale but it is done in the component implementation. 

At the end, what you said about "the application would be decide what the overall locale is" is true in both options since it is a parameter and you set it from your application consumer, the component does not decide from you. And in both options, if the component receives a wrong format (in case of exposing locale, another different from the global, or in case of direct format, a number format that is not aligned with the global locale), the behavior of the component will be the same so it will be visualized depending on the consumer input.

The response to "why this is needed now" and "why we do not have other components using it".. I'm not sure if we have a specific case where we are representing just only `numbers` in a component. Maybe our components are always more generic and a field is not restricted to a `number` bc it can really be a whatever.
Maybe the implementation of the `card` component here is not as generic as it should be and I should think of it as another more generic component with properties like this:
- Instead of `name` --> `title`
- Instead of `description` --> `subtitle`
- Instead of `members` --> `titleAsideText`
- Exposing an `icon` property instead of hardcoding the `tiny-contact` icon.
...

What are your thoughts?

PascalPrecht: Ah I see! I didn't realize the locale was used primarily for the member count. I guess this is to ensure we get either sth. like `1,000` vs `1000` vs `1.000` ?

Either way, I think then it's fine as it is. Probably not necessary to make it too generic as, by definition, that label is supposed to be a number.

0x-r4bbit (approved):

src/StatusQ/Components/StatusCommunityCard.qml
noeliaSD: Yeah! That's exactly for that!

PR #696, opened 2022-05-27
feat(Theme): Added `radius` and `bigPadding` properties

igor-sirotin:
We use `Style.current.bigPadding` kind of properties for sizing in status-desktop.
As I understand, we have deprecated the `Style` component in favor of StatusQ `Theme`.
If so, StatusQ `Theme` should also contain sizing properties.

I've only added `raidus` and `bigPadding`, which I used implementing https://github.com/status-im/status-desktop/issues/5814.

If all other sizing properties should also be moved here, I can do that.

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Core/Theme/Theme.qml
+++ b/src/StatusQ/Core/Theme/Theme.qml
@@ -6,6 +6,9 @@ QtObject {
     id: appTheme
     property QtObject palette: StatusLightTheme {}
 
+    property int bigPadding: 24
+    property int radius: 8
+
     function setTheme(theme) {
         palette = theme
     }


src/StatusQ/Core/Theme/Theme.qml
```diff
@@ -6,6 +6,9 @@ QtObject {
     id: appTheme
     property QtObject palette: StatusLightTheme {}
 
+    property int bigPadding: 24
+    property int radius: 8
```
PascalPrecht, line 10: We can add those here, but I was generally a bit skeptic about them. It's fine to have something like `padding` and `bigPadding` etc, but the primary reason to have those in variables is so that:

1. You can easily share the same value across various components
2. You can update the value in just a single place if need be

I would argue that: a) not every component will need `bigPadding` or `padding` or `halfPadding` (or whatever else there is in desktop), so we most likely still end up with hardcoded values anyways, plus, these variables are often used incorrectly just because they happen to have a value that is needed in a different context, and b) chances are low we'll want to change, say, `bigPadding`. And even if we did, we'll probably a lot of breakage in the UI because of how changing that value will affect the overall look and feel (not even considering that those variables are often used in wrong places).

Anyways, just a brain dump here on why those variables haven't been introduced in StatusQ yet. :)

PascalPrecht: As I said, it doesn't hurt to add them. We should probably then also add the others for consistency's sake.
We should just keep an eye on that these variables are used correctly, otherwise it defeats the purpose.

PR #698, opened 2022-05-30
fix(StatusRoundedImage): actual loaded image width and height updated

saledjenic:
In case you're dynamically change source image and image width and height
are way bigger than the rounded image component width and height, resizing
is not done correctly, binding `sourceSize` solves the issue.


src/StatusQ/Components/StatusRoundedImage.qml
```diff
@@ -28,11 +28,18 @@ Rectangle {
...
+        cache: false
+
+        onSourceChanged: {
+            if (sourceSize.width < width || sourceSize.height < height) {
+                sourceSize = Qt.size(width * 2, height * 2)
+            }
+            else {
+                sourceSize = undefined
+            }
+        }
```
osmaczko, line 42: It will not work properly when the image size is changed dynamically, because it's an imperative assignment, not a binding. You can use Qt.binding function to solve that.
```suggestion
            sourceSize = Qt.binding(function() {
                return sourceSize.width < width || sourceSize.height < height ? Qt.size(width * 2, height * 2) : undefined
            })
        }
```

osmaczko (changes requested):

osmaczko (approved):

PR #703, opened 2022-06-01
chore(StatusBaseText): add theme aware default color

osmaczko:
### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Core/StatusBaseText.qml
+++ b/src/StatusQ/Core/StatusBaseText.qml
@@ -16,7 +16,7 @@ import StatusQ.Core.Theme 0.1
            width: 240
            text: qsTr("Hello World!")
            font.pixelSize: 24
-           color: Theme.pallete.directColor1
+           color: Theme.palette.directColor1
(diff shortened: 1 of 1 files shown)


PascalPrecht:
This was long overdue - thanks!
We need to ensure though that we don't have usages of `StatusBaseText` in desktop that don't have an explicit `color` set and now behave differently due to set color (although, it would've been a bug anyways because they wouldn't work across themes, but maybe there's cases where it wasn't problematic)

PascalPrecht (approved):

osmaczko:
> This was long overdue - thanks! We need to ensure though that we don't have usages of `StatusBaseText` in desktop that don't have an explicit `color` set and now behave differently due to set color (although, it would've been a bug anyways because they wouldn't work across themes, but maybe there's cases where it wasn't problematic)

I took a quick tour over the app and didn't find such places (it would be very time-consuming to find them all), also as you mentioned it would've been probably a bug anyway.

PR #707, opened 2022-06-01
fix(StatusBaseButton): Mouse events ignored when button is disabled

igor-sirotin:
Previously the button was stealing mouse events even being disabled. Fixed that.
Also gave the buttons minor internal refactotring.

### Checklist

- [] update sandbox app

### DIFF
--- a/src/StatusQ/Controls/StatusBaseButton.qml
+++ b/src/StatusQ/Controls/StatusBaseButton.qml
@@ -86,14 +86,16 @@ Rectangle {
     radius: size !== StatusBaseButton.Size.Tiny ? 8 : 6
 
     color: {
-        if (statusBaseButton.enabled) {
+        if (statusBaseButton.enabled)
             return sensor.containsMouse || highlighted ? hoverColor
-                                        : normalColor
-        } else {
-            return disaledColor
-        }
+                                                       : normalColor;
+        return disaledColor
     }
 
+    QtObject {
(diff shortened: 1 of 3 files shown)


src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -102,8 +102,8 @@ Rectangle {
 
         cursorShape: loading ? Qt.ArrowCursor
                              : Qt.PointingHandCursor
-        hoverEnabled: !loading
-        enabled: !loading
+        hoverEnabled: !loading && statusBaseButton.enabled
```
PascalPrecht, line 105: Can prolly do `!hoverEnabled: enabled` then

PascalPrecht:
> It gives ability to make the Button look like enabled, but being disabled actually.

@igor-sirotin this sounds backwards to me. Are we sure we want this?

MishkaRogachev (approved):
Looks good!

PascalPrecht:
As discussed offline, the components that @igor-sirotin was looking for was `StatusRoundedIcon`, which enables us to render those icons in a circle without them being interactive buttons.

So no need to introduce a state in button components that make them look like they are enabled but in fact are disabled.

PascalPrecht (approved):

PascalPrecht:
Nice job Igor

PR #712, opened 2022-06-03
feat(StatusCommunityCard): New design

### LINKED ISSUE #692: Second iteration of `StatusCommunityCard` to fit mobile design

noeliaSD:
Closes status-im/StatusQ#692

### What does the PR do
- Added community banner.
- Added community restrictions / private.
- Modified card background shape.
- Modified logo shape and position.
- Modified text position and size.
- Modified `loading state` card.
- Added `DropShadow` effect.
- Added new icons.
- Added new community images and updated model.
- Added numbers format function.
- Added following new properties: `banner`,`activeUsers` `communityColor`, `isPrivate` and `tokenLogo`.

### Affected areas
`StatusCommunityCard` can be tested in:
- StatusCommunityCardPage
- demoapp/StatusAppCommunitiesPortalView

Screenshot 2022-06-03 at 12 50 08

https://user-images.githubusercontent.com/97019400/171840509-0dbfa023-e834-419f-abba-54e272a5e380.mov

### Checklist

- [x] update sandbox app


PascalPrecht:
This looks very solid to me!
One question: The private indicator in the top right corner... is it supposed to do anything? It looks like something that wants to be interacted with, but it doesn't seem to give any response when hovering it with the mouse... am I just misunderstanding that thing?

John-44:
@noeliaSD Looks great to me!

One thing I just noticed, the designs I sent you didn't illustrate how the following cases should be displayed:

- a depiction of how a card will appear if the whole community requires three different tokens to unlock
- a depiction of how a card will appear if a whole community requires two different tokens to unlock OR requires a different set of 3 tokens to unlock
- the overflow pattern that should be displayed if a community requires many different combinations of tokens to unlock

@pedro-et will add these variations to the figma

src/StatusQ/Components/StatusCommunityCard.qml
```diff
@@ -102,180 +130,330 @@ Rectangle {
 
     QtObject {
         id: d
-        property int dMargins: 12
+        property int cardWidth: 335
+        property int bannerHeigth: 64
+        property int cardHeigth: 190
+        property int totalHeigth: 230
+        property int margins: 12
+        property int bannerRadius: 20
```
stefandunca, line 138: Wondering if this shouldn't be relative to the banner window size and synced with the banner editor to keep consistency with the banner editing. If that is the case wondering where the relative definitions should reside, `Style`, `Theme`?

src/StatusQ/Components/StatusCommunityCard.qml
```diff
@@ -102,180 +130,330 @@ Rectangle {
...
+        OpacityMask {
+            anchors.fill: image
+            source: image
+            maskSource: mask
+        }
+    }
+
+    // Community logo:
+    Rectangle {
+        z: parent.z + 1
```
stefandunca, line 208: I see that in the designs the content looks like it should be on top of the banner and the logo on top of the content. I'm wondering what do you think about having this dependency explicitly also in the code? Sort of
- `Content Z` = `Banner Z` + 1
- `Logo Z` = `Content Z` + 1

stefandunca (approved):
Looking forward to seeing your work integrated with the app.

richard-ramos (approved):

noeliaSD:
> This looks very solid to me! One question: The private indicator in the top right corner... is it supposed to do anything? It looks like something that wants to be interacted with, but it doesn't seem to give any response when hovering it with the mouse... am I just misunderstanding that thing?

These are the `token permission tags`. AFAIK it is readonly and can have different permutations like @John-44 has described in this PR. I'll add also them following his indications!

noeliaSD:
> @noeliaSD Looks great to me!
> 
> One thing I just noticed, the designs I sent you didn't illustrate how the following cases should be displayed:
> 
> * a depiction of how a card will appear if the whole community requires three different tokens to unlock
> * a depiction of how a card will appear if a whole community requires two different tokens to unlock OR requires a different set of 3 tokens to unlock
> * the overflow pattern that should be displayed if a community requires many different combinations of tokens to unlock
> 
> @pedro-et will add these variations to the figma, then you can implement them. Sorry for forgetting to include these variations in the initial design.

Ok! I'll add the described scenarios here: https://github.com/status-im/status-desktop/issues/7554

noeliaSD:
> > These are the token permission tags. AFAIK it is readonly and can have different permutations like @John-44 has described in this PR. I'll add also them following his indications
> 
> @noeliaSD @John-44 should we add tooltips to these so users can get an idea what that thing is about? Cause if you don't know, you don't know

It will be addressed together with the requirements row update here: https://github.com/status-im/status-desktop/issues/7554

src/StatusQ/Components/StatusCommunityCard.qml
noeliaSD: Still pending answer from the designers. It will be addressed later on here: https://github.com/status-im/StatusQ/issues/722

John-44: @stefandunca just chatted to @noeliaSD, hopefully I understand the question correctly.  Your image cropping and scaling tool should export the image as rectangle, the image it exports shouldn't have rounded corners. This image is then used in multiple contexts in Status Mobile, and currently in one context in Status Desktop (the cards).  These different contexts where the image is used have different aspect ratios, and could hypothetically have different types of corners (square, rounded with a radius), so applying a corner radius to these images should happen in the UI code where the image is used.  

As far as I can see the aspect ratio that the image cropping and scaling is correct e.g. sufficient to meet the needs of all the different contexts.  

We could update the border radius the user sees in the image cropping and scaling tool to match the border radius used in the community cards.  But this is a low priority improvement, and anyway the actual image that the tool exports should be a rectangle that doesn't have rounded corners (because the corner rounding will happen in the UI code where the image is used if corner rounding is needed).

Does this answer your question?

stefandunca: @John-44 yes, you nailed it and agree that this is a light concern and should be addressed later on with low priority and only if we have significant visual differences.

PR #719, opened 2022-06-07
chore: adapt to new user/member backend models structure

osmaczko:
### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusMemberListItem.qml
+++ b/src/StatusQ/Components/StatusMemberListItem.qml
@@ -48,30 +48,35 @@ StatusListItem {
     */
     property string userName: ""
     /*!
-       \qmlproperty string StatusMemberListItem::chatKey
-       This property holds the chat key of the member represented.
+       \qmlproperty string StatusMemberListItem::pubKey
+       This property holds the chat public key of the member represented.
     */
-    property string chatKey: ""
+    property string pubKey: ""
     /*!
-       \qmlproperty string StatusMemberListItem::isMutualContact
-       This property holds if the member represented is a mutual contact.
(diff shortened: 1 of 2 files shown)


src/StatusQ/Components/StatusMemberListItem.qml
```diff
@@ -48,30 +48,35 @@ StatusListItem {
...
+       This property holds if the member represented is verified contact.
     */
-    property var trustIndicator: StatusContactVerificationIcons.TrustedType.None
+    property bool isVerified: false
     /*!
-       \qmlproperty string StatusMemberListItem::isOnline
+       \qmlproperty string StatusMemberListItem::isUntrustworthy
+       This property holds if the member represented is untrustworthy.
+    */
+    property bool isUntrustworthy: false
```
PascalPrecht, line 69: Can you elaborate on the decision to remove this `trustIndicator` enum?
I'm not necessarily opposing this change, just curious what made you do this change.

In my head, there can only be one of the trust states (verified, untrustworthy, none) at a time, so it makes sense for it to be a enumerable.

PascalPrecht (approved):

src/StatusQ/Components/StatusMemberListItem.qml
saledjenic: @osmaczko for me either options are ok, it's more a case of personal taste, both options cover all possible cases. I am more for 2 flags, one of the reasons is cause we cannot register enum as type easily via Nim and use it on the qml side  in c++ that will be easier, but having 2 flags and expose them as model's roles is perfectly fine in this case (again this is my opinion, as already said, I am completely fine whatever you decide).

PR #723, opened 2022-06-10
chore(StatusModal): make StatusModal follow padding

osmaczko:
https://github.com/status-im/status-desktop/pull/6010#discussion_r892148627

Still, the solution is not perfect, because there is no ability to set the top and bottom padding in a convenient way, see point 1 from https://github.com/status-im/status-desktop/issues/7556

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Popups/StatusModal.qml
+++ b/src/StatusQ/Popups/StatusModal.qml
@@ -194,10 +194,11 @@ QC.Popup {
 
     width: 480
 
-    topPadding: headerImpl.implicitHeight
-    bottomPadding: footerImpl.implicitHeight
-    leftPadding: 0
-    rightPadding: 0
+    padding: 0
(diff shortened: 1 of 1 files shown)


0x-r4bbit (approved):

igor-sirotin (approved):
Nice!
We should fix the whole Popup/DIalog story, but I think it's a good temporary solution anyway 👍 

Although, should we make the default `padding` = 16?

osmaczko:
> Nice! 

Thanks!

> We should fix the whole Popup/DIalog story, but I think it's a good temporary solution anyway +1

:100: 

> Although, should we make the default `padding` = 16?

It should be, but I don't want to risk any regression. There are a lot of assumptions on the default 0 padding in client code already. I will make 16 as default in reworked modals.

PR #730, opened 2022-06-13
fix(Theme): Updated `userCustomizationColors` list per designs

igor-sirotin:
Required for https://github.com/status-im/status-desktop/issues/5837

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Core/Theme/ThemePalette.qml
+++ b/src/StatusQ/Core/Theme/ThemePalette.qml
@@ -90,7 +90,20 @@ QtObject {
     property color miscColor11
     property color miscColor12
 
-    property var userCustomizationColors: []
+    property var userCustomizationColors: [
+        getColor('black', 0.8),
+        getColor('grey'),
+        getColor('blue2'),
+        getColor('purple'),
+        getColor('cyan'),
+        getColor('violet'),
+        getColor('red2'),
+        getColor('yellow'),
+        getColor('green2'),
+        getColor('moss'),
+        getColor('brown'),
+        getColor('brown2'),
+    ]
 
     property var identiconRingColors: []
 
--- a/src/StatusQ/Controls/StatusColorSelectorGrid.qml
+++ b/src/StatusQ/Controls/StatusColorSelectorGrid.qml
@@ -14,18 +14,7 @@ Column {
 
     property int selectedColorIndex: 0
     property string selectedColor: ""
-    property var model:[ StatusColors.colors['black'],
-                         StatusColors.colors['grey'],
-                         StatusColors.colors['blue2'],
-                         StatusColors.colors['purple'],
-                         StatusColors.colors['cyan'],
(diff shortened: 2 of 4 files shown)


src/StatusQ/Core/Theme/ThemePalette.qml
```diff
@@ -90,7 +90,20 @@ QtObject {
...
+        getColor('purple'),
+        getColor('cyan'),
+        getColor('violet'),
+        getColor('red2'),
+        getColor('yellow'),
+        getColor('green2'),
+        getColor('moss'),
+        getColor('brown'),
+        getColor('brown2'),
+    ]
```
PascalPrecht, line 106: With this change, both themes get the same set of colors which was not the case before. Are we sure we want this?

osmaczko (commented):
User customization colors are defined [here, see "Customization"](https://www.figma.com/file/WMmmNowe0sTVCaxadUB8mE/Desktop-Foundations?node-id=1%3A609). They are defined per Theme. To which design do these changes relate?

osmaczko:
> I'm currently implementing the "Sync device" settings page.

I believe that already implemented customization colors should be used there as well.
image
image

> A) the user customisation colours are only defined in one place  B) total number of colours in the user customisation colour set is 12, but in some contexts only a subset of these colours will be used e.g. 10 colours, or 8 colours from that 12 colour pallet   C) they can be used and referenced throughout the product wherever user customisation colours are required e.g. this set of colours is used for colouring Accounts in the user's wallet, and is also used in the Community create and edit screens. 

by John_

If there are still any doubts, please ask John for clarification.

src/StatusQ/Core/Theme/ThemePalette.qml
osmaczko: Please note that User Customisation colors have the same name and yet are slightly different. Ideally, we should use color indexes, as we use for user colors now.

osmaczko: Currently, for user icon colors, we use a color index, see: https://github.com/status-im/status-desktop/blob/19235e2b7940a1c686693106a4533ca47c9f8d23/ui/imports/utils/Utils.qml#L665-L674

> But as I checked, we store the color, not it's index.

The color index for user icons is deterministically obtained from pubKey and so there is no need to store it in the database.

The unencrypted accounts database (the one responsible for keeping information about accounts you can log in to) contains a color index though. That's because pubKey is not accessible before we log in.

In other scenarios that's something we should either fix (migrate color to color index) in order to solve the problem you mentioned:
> As we never save the "index of unversal color" but the color value itself, then it's possible that the color was selected using Dark theme and will be viewed later in Light theme.

Or make user customization colors stay the same in every theme.

We should not solve it by simply preferring fixed colors from outdated Designs, IMO we should not merge this change.

PR #735, opened 2022-06-18
fix(icons): Updated arrow icons

igor-sirotin:
~~I noted that the icon in `StatusRoundButton` is not precisely centered.~~

~~There's several options to fix this:~~
~~1. Change icon size from `23` to `22`~~
~~2. Change icon size from `23` to `22` and change button size from `44` to `45`~~
~~3. Change button size from `44` to `45`~~
 
Снимок экрана 2022-06-19 в 12 01 17

~~I think that the second option looks the best, it's in the PR.~~

UPDATE:

As discussed [below in comments](https://github.com/status-im/StatusQ/pull/735#discussion_r901951798), `23px` icon size is not a bug, but a feature. It actually makes all our icons looks better.

It just didn't work on arrow icons, because they were of different source size.
So I simply updated the arrow icon files from [this figma](https://www.figma.com/file/Gk13uZiVwGLHMPvQjaICAMWM/999%3A-DEPRECATED-Status-icons?node-id=38%3A2).

### Checklist

- [x] update sandbox app


PascalPrecht:
This PR doesn't actually anything about the centering being incorrect no?

src/StatusQ/Controls/StatusRoundButton.qml
```diff
@@ -86,9 +86,9 @@ Rectangle {
         readonly property color iconColor: statusRoundButton.enabled ? statusRoundButton.icon.color : statusRoundButton.icon.disabledColor
     }
 
-    implicitWidth: 44
-    implicitHeight: 44
-    radius: width / 2;
+    implicitWidth: 45
```
alexandraB99, line 89: logically 22/44 should do the job however I see in the screenshot that in this case with the arrow icon 22/45 looks slightly better. did you try with other icons?

alexandraB99 (approved):

Khushboo-dev-cpp (approved):

PR #741, opened 2022-06-23
feat(StatusRoundButton): Added `Tertiary` type

noeliaSD:
- Transparent background but when hovered, like Primary type in idle state.

- Updated sandbox with the new type.

### Checklist

- [x] update sandbox app

https://user-images.githubusercontent.com/97019400/175244789-58e0c37a-cb2b-4088-887f-a24fabf292dc.mov

### DIFF
--- a/sandbox/controls/Buttons.qml
+++ b/sandbox/controls/Buttons.qml
@@ -174,6 +174,25 @@ Column {
             loading: true
         }
 
+        // transparent
+
+        StatusRoundButton {
+            type: StatusRoundButton.Type.Tertiary
+            icon.name: "info"
+        }
+
+        StatusRoundButton {
+            type: StatusRoundButton.Type.Tertiary
+            icon.name: "info"
+            enabled: false
+        }
+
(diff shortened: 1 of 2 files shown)


alexandraB99 (approved):
The PR looks good as PR, what doesn't look that good is those 160 lines of code for a button, we should try thinking at some whether we wanna continue expanding components or separate in various smaller files/components for simplicity. So many different variations in a single component makes it hard to use and understand.

noeliaSD:
> The PR looks good as PR, what doesn't look that good is those 160 lines of code for a button, we should try thinking at some point whether we wanna continue expanding the existing components or separate variants in various smaller files/components for simplicity. So many different variations in a single component makes it hard to use and understand.

160 lines do not seem as bad to me for a single component, but maybe more than 300, for a small one, yes. In those particular cases, yes, we should decide how to proceed. Maybe we can add some rules into the `contributing.md` file. WDYT?

jrainville (approved):

PR #750, opened 2022-07-04
[Community Permissions] Created new components and updated existing ones to add `Who holds` tokens dropdown functionality

noeliaSD:
Part of: https://github.com/status-im/status-desktop/issues/6038

### What does the PR do
All needed changes for task https://github.com/status-im/status-desktop/issues/6038
- Creates new components:
  - `StatusDropdown`
  - `StatusIconTextButton` 
  - `StatusItemSelector`
  
- Minor changes in:
  - `StatusFloatValidator`
  - `StatusPickerButton`
  - `StatusSwitchTabButton`
  - `StatusBaseInput` 
  - `StatusInput`

- `StatusItemSelector` and `StatusDropdown`:

https://user-images.githubusercontent.com/97019400/177334320-63349752-74c5-41f1-86fd-023a86d4fea9.mov

- `StatusIconTextButton`:

https://user-images.githubusercontent.com/97019400/177334425-98cfde3a-a0fb-4d5f-b372-b7d1a640a87f.mov
 
### Checklist

- [x] update sandbox app


src/StatusQ/Controls/StatusDropdown.qml
```diff
@@ -0,0 +1,49 @@
...
+            y: root.y + margins
+            contentItem: ColumnLayout {
+                ...
+            }
+        }
+   \endqml
+
+   For a list of components available see StatusQ.
+*/
+QC.Popup {
```
caybro, line 30: You already have a dedicated QC.MenuItem component, I suggest creating this new custom one on something pre-existing (MenuItem already inherits from Popup ;)

src/StatusQ/Controls/StatusIconTextButton.qml
```diff
@@ -0,0 +1,42 @@
+import QtQuick 2.0
+import QtQuick.Layouts 1.14
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+
+Item {
```
caybro, line 7: This feels again like it needs to be based on QQC2.AbstractButton (just customize the contentItem: https://doc.qt.io/qt-5/qtquickcontrols2-customize.html#customizing-button). Then you won't need any of the custom properties, signals or the MouseArea below

caybro (changes requested):

src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -0,0 +1,140 @@
...
+                        function operatorTextFormat(operator) {
+                            switch(operator)
+                            {
+                            case StatusItemSelector.Operators.And:
+                                return root.andOperatorText
+                            case StatusItemSelector.Operators.Or:
+                                return root.orOperatorText
+                            case StatusItemSelector.Operators.None:
+                                return ""
+                            }
```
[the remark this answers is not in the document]
noeliaSD: I will move the `Operators` enum to a common file like `Utils` and this function to the toplevel but private since it is specific for this component. Thanks for the suggestion!

src/StatusQ/Controls/StatusIconTextButton.qml
noeliaSD: I've started with your suggestion but faced to an issue when I want to change the default `cursorShape` to `Qt.PointingHandCursor`. The solution found is to add a custom `MouseArea` to change it so.. not sure which are the benefits in that case. Also if we could directly use the `text` or the `icon` from `Button`, it certainly would be an improvement, but we have our custom controls that define the style so I cannot see a benefit by doing this.
Do you know an alternative way of doing it simpler? Thanks!!

src/StatusQ/Controls/StatusDropdown.qml
noeliaSD: I'm not sure about using `QC.MenuItem` as a base class since this component is just a `popup` container with a specific background, a dropdown container. If `MenuItem` inherits from `Popup` but has more functionality, we will be inheriting more functionality than needed for this specific case. WDYT?

caybro: Sorry, of course I meant a `Menu`, that one is the container, see here: https://doc.qt.io/qt-5/qml-qtquick-controls2-menu.html

src/StatusQ/Controls/StatusIconTextButton.qml
caybro: AbstractButton is definitely the way to go as the base class, even regarding custom styling. You just have to override `background` and/or `contentItem`. As for the cursor shape, there is `HoverHandler`, too bad the `cursorShape` property has only been introduced in Qt 5.15 (https://doc.qt.io/qt-5/qml-qtquick-hoverhandler.html#cursorShape-prop) :/

caybro: For now, you can still setup a temporary MouseArea to change the cursor shape (with a TODO to remove when we switch to Qt 5.15), sth like this:

```qml
MouseArea { 
  cursorShape: Qt.IBeamCursor
  acceptedButtons: Qt.NoButton
  hoverEnabled: true 
}
```

src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -0,0 +1,226 @@
...
+*/
+Rectangle {
+    id: root
+    /*!
+       \qmlproperty string StatusItemSelector::icon
+       This property holds the icon name for the icon represented on top of the component as a title icon.
+    */
+    property string icon
+    /*!
+       \qmlproperty string StatusItemSelector::iconSize
```
caybro, line 49: `int`

src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -0,0 +1,226 @@
...
+       This property holds the titel shown on top of the component.
+    */
+    property string title
+    /*!
+       \qmlproperty string StatusItemSelector::defaultItemText
+       This property holds the default item text shown when the list of items is empty.
+    */
+    property string defaultItemText
+    /*!
+       \qmlproperty string StatusItemSelector::defaultItemImageSource
```
caybro, line 64: `url`

src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -0,0 +1,226 @@
...
+            }
+        }
+       \endqml
+    */
+    property ListModel itemsModel: ListModel { }
+    /*!
+       \qmlproperty string StatusItemSelector::andOperatorText
+       This property holds the string text representation for an `AND` logical operator.
+    */
+    property string andOperatorText: "and"
```
caybro, line 98: if those are displayed to the user, it should be marked so, eg. `qsTr("and")`

src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -0,0 +1,226 @@
...
+    /*!
+       \qmlproperty string StatusItemSelector::orOperatorText
+       This property holds the string text representation for an `OR` logical operator.
+    */
+    property string orOperatorText: "or"
```
caybro, line 103: Same here

src/StatusQ/Controls/StatusPickerButton.qml
```diff
@@ -1,25 +1,36 @@
...
-    implicitWidth: 446
-    implicitHeight: 44
 
     property color bgColor: Theme.palette.baseColor2
     property color contentColor: Theme.palette.baseColor1
     property var type: StatusPickerButton.Type.Next
     property int lateralMargins: 16
     property int textPixelSize: 15
+    /*!
+       \qmlproperty string StatusPickerButton::image
```
caybro, line 18: not a `string`

src/StatusQ/Controls/StatusSwitchTabButton.qml
```diff
@@ -8,6 +8,8 @@ import StatusQ.Core.Theme 0.1
 TabButton {
     id: statusSwitchTabButton
 
+    property int pixelSize: 15
```
caybro, line 11: `fontPixelSize`? otherwise one might get the impression it's the size (height?) of the button itself

src/StatusQ/Controls/Validators/StatusFloatValidator.qml
```diff
@@ -24,13 +24,13 @@ import StatusQ.Controls 0.1
    For a list of components available see StatusQ.
 */
 StatusValidator {
-    id: root
+    id: root    
 
     /*!
        \qmlproperty string StatusFloatValidator::bottom
```
caybro, line 30: `real`

src/StatusQ/Controls/Validators/StatusFloatValidator.qml
```diff
@@ -40,9 +40,15 @@ StatusValidator {
 
     /*!
        \qmlproperty string StatusFloatValidator::top
```
caybro, line 42: `real`

src/StatusQ/Controls/Validators/StatusFloatValidator.qml
```diff
@@ -40,9 +40,15 @@ StatusValidator {
...
-       This property holds the validator's highest acceptable value. By default, this property's value is derived from the highest float available (typically 2147483647).
+       This property holds the validator's highest acceptable value. By default, this property's value is derived from the highest float available.
     */
-    property real top
+    property real top: qmlDoubleValidator.top
+
+    /*!
+       \qmlproperty string StatusFloatValidator::qmlDoubleValidator
```
caybro, line 48: not a `string`

caybro (changes requested):

src/StatusQ/Components/StatusItemSelector.qml
noeliaSD: Oh!!! Copy / Paste issues!! :)

src/StatusQ/Components/StatusItemSelector.qml
noeliaSD: We are not using `qsTr` in `StatusQ`. It is still a discussion in how to approach this but for now in `StatusQ` we are providing just only a default value and exposing the property to be correctly assigned by the consumer.

caybro: I see... but providing the correct default here would be sufficient in this case

sandbox/pages/StatusItemSelectorPage.qml
```diff
@@ -0,0 +1,44 @@
...
+ColumnLayout {
+    spacing: 20
+    StatusItemSelector {
+        id: selector
+        icon: "qrc:/images/SNT.png"
+        iconSize: 24
+        title: qsTr("Item Selector Title")
+        defaultItemText: qsTr("Example: Empty items")
+        andOperatorText: qsTr("and")
+        orOperatorText: qsTr("or")
```
caybro, line 17: And here the `qsTr()` is useless :) It's just the sandbox app

src/StatusQ/Controls/StatusDropdown.qml
noeliaSD: I'm still with doubts about this suggestion since the main purpose for the `StatusDropdown` is not just to be a `Menu` with different `MenuItem` childrens, but it is more generic. It could be used to simply display a list of items, but only for informational purposes, or simply display an image or some behavior that is not specifically of a 'Menu'. It's a simple `Popup` container where you can customize its `contentItem`.

caybro (approved):
LGTM! Nice work :)

sandbox/controls/Buttons.qml
```diff
@@ -343,4 +343,26 @@ Column {
...
+            Layout.alignment: Qt.AlignVCenter
+            spacing: 0
+            statusIcon: "next"
+            icon.width: 24
+            icon.height: 24
+            iconRotation: 180
+            text: "Previous page"
+            font.pixelSize: 15
+            onClicked: testText.visible = !testText.visible
+        }
```
PascalPrecht, line 360: How exactly is this different from `StatusFlatButton` that has an `icon`

src/StatusQ/Components/StatusItemSelector.qml
PascalPrecht: To give some context here, we originally didn't use qsTr() in StatusQ because our setup would change the source files from qsTr() to qsTrId().

It didn't feel like to use something like StatusQ and then have its source being changed by some mechanism that's coming from the app.

I believe this is no longer an issue though (would have to verify with @osmaczko and @jrainville on what's the latest state).

We should definitely move to `qsTr()` if this has been solved.

src/StatusQ/Components/StatusItemSelector.qml
```diff
@@ -0,0 +1,227 @@
...
+                        }
+                    }
+                    StatusListItemTag {
+                        title: model.text
+                        image.source: model.imageSource
+                        color: Theme.palette.primaryColor3
+                        closeButtonVisible: false
+                        titleText.color: Theme.palette.primaryColor1
+                        titleText.font.pixelSize: 15
+                        //onClicked: // TODO: Open remove or edit dialog
```
PascalPrecht, line 208: This a leftover?

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -359,6 +364,9 @@ Item {
             Layout.fillWidth: true
             Layout.fillHeight: true
 
+            leftPadding: root.leftPadding
+            rightPadding: root.rightPadding
```
PascalPrecht, line 368: This will cause the padding to double no? Because AFAIK we use it already inside `StatusInput` to control padding of itself. Adding this here, would add the padding as well to StatusBaseInput which means it's doubled (unless I'm missing something) That desired?

PascalPrecht (approved):
LGTM I've added some comments though that you may or may not want to leave feedback on

src/StatusQ/Components/StatusItemSelector.qml
noeliaSD: No, it is here for the next steps, that are just comming!

sandbox/controls/Buttons.qml
noeliaSD: It doesn't have an hover color, just only like a text link. Yeah.. I'm thinking maybe it is redundant and the way to go for it is by adding a variant in `StatusFlatButton`... WDYT?

src/StatusQ/Controls/StatusInput.qml
noeliaSD: `StatusBaseInput` effectively has the `leftPadding` and `rightPadding` properties exposed, so if you use `StatusBaseInput` you can customize the padding properly. On the other hand, `StatusInput` already had the same properties exposed but not linked to anything, so the padding in this case was not customizable. What I've done here is simply bind these properties to the `StatusBaseInput` properties, so now `StatusInput` allows you to customize them as well.

src/StatusQ/Components/StatusItemSelector.qml
noeliaSD: Oh, nice to know it! I will update then it to `qsTr()`

sandbox/controls/Buttons.qml
PascalPrecht: Not sure from the top of my head, but maybe we expose the `sensor` in `StatusFlatButton`, then you could simply make it not have a hover (although, it sounds backwards to me to introduce/have a control that looks like a link/or button but doesn't have a hover indication. How will users know they can click it?)

noeliaSD: I have a list of design questions for the task I'm working on and one of them is this, so I'll merge this PR as it is and move forward but once I've the response from design, I'll update `StatusQ` accordingly (It doesn't make sense to do more changes here if at the end the design team decides that it also needs hover color.. so then I'll remove the `statusIconTextButton` added here).

BTW, this is an example of the `link` button from a design:
Screenshot 2022-07-13 at 13 10 46

PR #756, opened 2022-07-06
fix(StatusIcon): flag/currency icons are rendered blurred

caybro:
- set the Image's `sourceSize` only if it's an SVG, otherwise it is
rendered blurry
- remove antialiasing/mipmap; it's not the correct fix, doesn't really
improve the visual quality for smallish items like icons and just causes
memory/GPU overhead

### Screenshots (before/after)
Snímek obrazovky z 2022-07-05 22-35-38
Snímek obrazovky z 2022-07-05 23-08-53

### status-desktop screenshot
Snímek obrazovky z 2022-07-06 11-52-19

### Checklist

- [ ] update sandbox app

### DIFF
--- a/sandbox/demoapp/data/Models.qml
+++ b/sandbox/demoapp/data/Models.qml
@@ -1000,7 +1000,7 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
             name: "United States Dollar"
             shortName: "USD"
             symbol: "$"
-            imageSource: "../../assets/twemoji/26x26/1f4b4.png"
+            imageSource: "../../assets/twemoji/svg/1f4b4.svg"
             category: ""
             selected: false
         }
@@ -1009,7 +1009,7 @@ CExPynn1gWf9bx498P7/nzPcxEzGExhBdJGYihtAYQlO+tUZvqrPbqeudo5iJGEJjCE15a3VtodH3q2I
             name: "British Pound"
             shortName: "GBP"
             symbol: "£"
-            imageSource: "../../assets/twemoji/26x26/1f4b5.png"
+            imageSource: "../../assets/twemoji/svg/1f4b5.svg"
             category: ""
             selected: false
         }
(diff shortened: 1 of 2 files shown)


caybro:
> 1. Take a look on the theme `Switch` in the top-right corner of your screenshots. They got broken slightly_frowning_face
>    Not sure if it's because of your changes. But that would make sense.

@igor-sirotin here's your theme switch "icon" :D

```qml
Text {
  text: "🌤"
  font.pixelSize: 15
}
```

caybro:
> > Hmm, I'll take a look but they seem the same to me? (compare before/after). How should they look like?
> 
> First of all, yes, I should have noted the before/after. It's probably not your changes' consequences. But that's interesting. I'm on latest master ([c4421d3](https://github.com/status-im/StatusQ/commit/c4421d35523db1d1aafac630917eddcd788f185f)), the switch looks like this: imageimage

You're probably usign a different font... what OS?

> > Those flag/currency icons are in PNG, with no `@2x` variant so can't do much more here (yes, this screenshot is taken on a 4k screen)
> 
> Okaay disappointed Should we ask anyone where did we get these icons and try to get the high-res ones?

OK, I'll make a note in the original BR when this gets merged

MishkaRogachev (approved):

caybro:
For the record, the flags come from this site: https://twemoji.twitter.com/

caybro:
> > For the record, the flags come from this site: https://twemoji.twitter.com/
> 
> Well, if the flags are coming from Twemoji, can't we just use the `svg` for those? We even have them in the assets already.

@igor-sirotin I was curious so I just gave it a quick try in statusq/sandbox and it looks a wee better than the fixed version of PNGs (code stays the same, I just exchanged the model to use SVGs), see here:

Snímek obrazovky z 2022-07-06 13-42-55

caybro:
@igor-sirotin can you actually please give it a test locally on macOS (with this version first), and then eventually if it's still not good enough, I can push the SVG version to see the difference. It's really hard to capture on a screenshot, has to be seen live.

One note though: if we decide to go with SVGs, they are much bigger (in file size), so they are going to use a lot more GPU memory

PR #758, opened 2022-07-06
fix(StatusSwitch): Removed `running` property on transition animation

igor-sirotin:
This was probably made by mistake and doesn't actually makes sense.
This was causing an annoying warning:
```
QML NumberAnimation: setRunning() cannot be used on non-root animation nodes.
```

### Checklist

- [ ] update sandbox app


caybro:
A dupe of #755?

caybro (changes requested):
Looks good but there's a better (more complete) PR out there :)

PR #772, opened 2022-07-13
fix(StatusAssetSelector): Fixes the issue of longer text overlapping with the arrow.

Khushboo-dev-cpp:
fix(StatusAssetSelector): Fixes the issue of longer text overlapping with the arrow. 
Have also added a max width for the text after which it should elide

### Checklist

- [x] update sandbox app

### DIFF
--- a/src/StatusQ/Controls/StatusAssetSelector.qml
+++ b/src/StatusQ/Controls/StatusAssetSelector.qml
@@ -18,7 +18,7 @@ Item {
     }
     // Define this in the usage to get balance in currency selected by user
     property var getCurrencyBalanceString: function (currencyBalance) { return "" }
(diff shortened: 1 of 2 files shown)


src/StatusQ/Controls/StatusAssetSelector.qml
```diff
@@ -18,7 +18,7 @@ Item {
     }
     // Define this in the usage to get balance in currency selected by user
     property var getCurrencyBalanceString: function (currencyBalance) { return "" }
-    implicitWidth: 86
+    implicitWidth: 106
```
caybro, line 21: This should rather imho get its natural width based on the contents (childrenRect.width)

PascalPrecht (approved):

PR #783, opened 2022-07-15
fix(demo/CreateChatView): fixed layout issues

alexandraB99:
As part of https://github.com/status-im/status-desktop/issues/6452

### Checklist

- [ ] update sandbox app


sandbox/demoapp/CreateChatView.qml
```diff
@@ -73,26 +74,43 @@ Page {
         }
 
         StatusButton {
+            implicitWidth: 106
             implicitHeight: 44
             Layout.alignment: Qt.AlignTop
             enabled: (tagSelector.namesModel.count > 0)
             text: "Confirm"
```
caybro, line 81: `qsTr()`, unless we override it in status-desktop

caybro (changes requested):

caybro (approved):

sandbox/demoapp/CreateChatView.qml
```diff
@@ -73,26 +74,43 @@ Page {
...
     }
 
     contentItem: Item {
         anchors.fill: parent
 
         StatusBaseText {
-            visible: (contactsModel.count === 0)
-            anchors.centerIn: parent
+            anchors.left: parent.left
+            anchors.leftMargin: 252*0.66
```
PascalPrecht, line 102: Any  reason we're calculating this (despite those numbers looking kinda odd), instead of hardcoding the result?
The values don't seem variable so just curious why we're doing it this way

PascalPrecht (approved):
Cool changes!

sandbox/demoapp/CreateChatView.qml
alexandraB99: right, it was there from before in the width property, so I moved it to the margis but it's not needed as it seems! removed :)

PR #788, opened 2022-07-20
feat(StatusQ)!: Merging Icon, Background and Image Settings

### LINKED ISSUE #781: Merge StatusIconSettings, StatusImageSettings and StatusIconBackgroundSettings
Update all components/places of usage

alexandraB99:
Closes #781

### Checklist

- [ ] update sandbox app
- [x] is this a breaking change?


PascalPrecht:
This needs thorough testing at those changes are very invasive.
Also @alexandraB99 it'd be great if you could put the breaking change information into the commit (i.e. adding the dedicated section, mentioning what is breaking how [before vs after] etc).

Also, since there was no more information given on https://github.com/status-im/StatusQ/issues/781#issuecomment-1185658340, would be nice if you could leave some information retroactively on why this change was needed in the first place

src/StatusQ/Core/StatusAssetSettings.qml
```diff
@@ -4,22 +4,35 @@ import StatusQ.Core 0.1
...
+    //icon bg
+    property real bgWidth
+    property real bgHeight
+    property int bgRadius
+    property color bgColor: "transparent"
+
+    //image
+    property real imgWidth
+    property real imgHeight
+    property url imgSource
```
caybro, line 35: Just one question here: how is `imgWidth` different from (icon) `width`? Same for `imhHeight` and `imgSource`; are there valid cases when both an image and icon are used together (sounds weird)

alexandraB99: yeah, in very rare cases (I think I saw 1 or 2) there image and icon together

caybro: Ah hmmm, yeah inside a Loader which decides whether to use an icon or an image?

alexandraB99: Need to check if those 2 are used simultaneously, can't remember now.

src/StatusQ/Core/StatusAssetSettings.qml
```diff
@@ -4,22 +4,35 @@ import StatusQ.Core 0.1
...
+    property int imgStatus
```
caybro, line 36: Just curious what's the purpose of this prop?

PascalPrecht: I assume it's the state in which it is.
We have usages of images that are being loaded asynchronously and that can fail. 

Maybe it's something else though

src/StatusQ/Core/StatusAssetSettings.qml
alexandraB99: @caybro Indeed, it's either this or that, however some times there are minor differences when we load an image or an icon, mostly in width and height, which are manageable but I would like a better way of detecting this than what I used here as a quick solution. To give more insights: depending on if we have icon name set we load StatusIcon or if we have image StatusRoundedImage. It's either this or that. As a quick solution I'm using `root.asset.name.indexOf(".") === -1 ? roundedIcon : roundedImage` to load the correct component, however I don't like this, would you have any better idea?

src/StatusQ/Core/StatusAssetSettings.qml
alexandraB99: I have checked only in StatusQ and is used in StatusSmartIdenticon to load the letteridenticon instead if the image is not loaded because of an error. probably we could remove it, but we need to check and test in desktop as well.

alexandraB99:
> This needs thorough testing as those changes are very invasive. Also @alexandraB99 it'd be great if you could put the breaking change information into the commit (i.e. adding the dedicated section, mentioning what is breaking how [before vs after] etc).
> 
> Also, since there was no more information given on [#781 (comment)](https://github.com/status-im/StatusQ/issues/781#issuecomment-1185658340), would be nice if you could leave some information retroactively on why this change was needed in the first place

@PascalPrecht indeed! this is nevertheless still a draft, we need to agree on everything here, finilize the implementation in StatusQ and then adapt/test desktop. I'll mark it as draft for now as I might be adding more commits.

src/StatusQ/Core/StatusAssetSettings.qml
PascalPrecht: If we remove this, then we'll lose the fallback, which I don't think we want.
Especially with use cases like the discord import tool, chances are getting higher that there'll be assets that can't be loaded.

caybro: Ok, makes sense but then we could simply add the default fallback as:
```
property real imgWidth: width
```
and so on

alexandraB99:
@PascalPrecht, @caybro the `AssetSettings` API could remain as this: https://github.com/status-im/StatusQ/pull/788/files#diff-2440e5c2e506ac9729f4930d5c29071c6239027e68f67f513298216f4f213ef3 

Same as we do with `isIdenticon`, `isLetterIdenticon` we'll simply have to indicate as well if `isImage`. For sandbox app this is handled in QML, for desktop we need to check if this needs to be provided by the model as the rest of the properties. With this API we can work with `icon.name` and/or `icon.source` making the transition to QQC2 easier as in when we're there, we can just remove `name`, `source`, `width` and `height` properties from `AssetSettings` and keep only the necessary additions. Please let me know your thoughts! If ok, I would go ahead and update desktop so that we can move on with testing this.

caybro (approved):
Let's bite the bullet :)

PascalPrecht:
Are these commits supposed to be squashed? 
If not we gotta ensure the second commit follows the converntions

PR #793, opened 2022-07-21
test(StatusChatToolBar): add object name for tests

jrainville:
### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusChatToolBar.qml
+++ b/src/StatusQ/Components/StatusChatToolBar.qml
@@ -83,6 +83,7 @@ Item {
 
             StatusFlatRoundButton {
                 id: menuButton
+                objectName: "chatToolbarMoreOptionsButton"
                 width: 32
                 height: 32
                 icon.name: "more"


src/StatusQ/Components/StatusChatToolBar.qml
```diff
@@ -83,6 +83,7 @@ Item {
 
             StatusFlatRoundButton {
                 id: menuButton
+                objectName: "channelMoreOptionsButton"
```
PascalPrecht, line 86: Wonder if it should be rather called `chatToolbarMoreOptionsButton` because I could imagine this to conflict with the more options button of channels otherwise (in case those will be tested as well and need a name)

Otherwise I trust your choice

PascalPrecht (approved):

src/StatusQ/Components/StatusChatToolBar.qml
jrainville: Good idea. Updated

PR #797, opened 2022-07-22
Make StatusListPicker work with backend models

osmaczko:
closes: https://github.com/status-im/status-desktop/issues/6500

### Checklist

- [x] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusListPicker.qml
+++ b/src/StatusQ/Components/StatusListPicker.qml
@@ -17,7 +17,7 @@ import SortFilterProxyModel 0.2
 
    The \c StatusListPicker is populated with a data model. The data model is commonly a JavaScript array or a ListModel object.
 
-   StatusListPicker can be made as a single or multiple options picker by setting its StatusListPicker::multiSelection property properly or will be auto-set if the model contains / provides more than one selected items.
+   StatusListPicker can be made as a single or multiple options picker by setting its StatusListPicker::multiSelection property properly.
    The StatusPickerButton text holds the text of the current item selected in the list picker. If there is more than one item selected, a string composed will be displayed.
    The drop-down list incorporates a searcher by the following model roles: `name` or / and `shortName`.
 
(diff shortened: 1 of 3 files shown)


osmaczko:
@noeliaSD could you please check in the sandbox app if `StatusListPicker` works as expected?

stefandunca (approved):
Beautifully simple.

src/StatusQ/Components/StatusListPicker.qml
```diff
@@ -197,40 +175,43 @@ Item {
...
-                     // Clear all the list
-                    root.inputList.get(i).selected = false
-                }
+        // If it is a single selection we must ensure that when a new key is selected others are cleared.
+        id: itemsSelectorHelper
+
+        function selectItem(modelData, checked) {
+            modelData.selected = checked
+            if (!root.multiSelection && checked) {
+                itemsSelectorHelper.uncheckOthersThan(modelData.key)
```
caybro, line 199: I wonder why is this needed for? If you put all the checkboxes/radio buttons in a `ButtonGroup` (https://doc.qt.io/qt-5/qml-qtquick-controls2-buttongroup.html) then it should be done automatically

caybro (commented):
Quite an improvement! But I still think it could be done even simpler :)

noeliaSD (changes requested):
Nice to see the sort proxy model integrated here!! :)

Tested and some issues found in sandbox:

1. Initialization in `Multiple selection` and `Multiple selectin and diplayed names...` is not correct. Shown USD and GBP but selected EUR and ETH in both cases.
2. Multiple selection option doesn't update the displayed name.
3. Single selection: New selection not updated when the searcher is used.

Attached a video for better understanding: 

https://user-images.githubusercontent.com/97019400/180936558-a98d009b-17da-466b-a83d-2be80152a140.mov

osmaczko:
@noeliaSD great findings (seems I didn't perform tests with care :sweat: ). All of them should be fixed now, could you please re-check?

src/StatusQ/Components/StatusListPicker.qml
osmaczko: Yeah, that's a good question. Unfortunately, the delegate of `StatusListPicker` is not `AbstractButton` based..

noeliaSD (approved):
LGTM!

PR #809, opened 2022-07-27
feat(StatusDialog): Add background color property to make the dialog background customizable

Khushboo-dev-cpp:
feat(StatusDialog): Add background color property to make the dialog background customizable

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Popups/Dialog/StatusDialog.qml
+++ b/src/StatusQ/Popups/Dialog/StatusDialog.qml
@@ -22,10 +22,7 @@ Dialog {
         color: Theme.palette.backdropColor
     }
 
-    background: Rectangle {
-        color: Theme.palette.statusModal.backgroundColor
-        radius: 8
-    }
+    background: StatusDialogBackground {}
 
     header: StatusDialogHeader {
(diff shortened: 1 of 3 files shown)


osmaczko (changes requested):
Please don't add properties to `StatusDialog`, we should keep it as simple as possible to avoid (in the future) big number of API properties and aliases as we do have now for many other components. If a custom color is needed please simply do the following on the client side:
```
background: Rectangle {
    color: yourColor
    radius: 8
}
```
You could also extract `DialogBackground` component to avoid `radius: 8` duplication:
```
background: DialogBackground {
    color: yourColor
}
```
Take a look at similar discussion: https://discord.com/channels/624205794384281629/697843216048717905/999253801347334144

Khushboo-dev-cpp:
makes sense @osmaczko ! will do :)

Khushboo-dev-cpp:
> makes sense @osmaczko ! will do :)

All done, check when you can :) @osmaczko

osmaczko (approved):
Looks good!

PR #812, opened 2022-07-28
feat(StatusBaseButton)!: Converting to QQC2 Button

alexandraB99:
### Checklist

- [x] update sandbox app

### DIFF
--- a/sandbox/controls/Buttons.qml
+++ b/sandbox/controls/Buttons.qml
@@ -21,18 +21,31 @@ Column {
         StatusButton {
             text: "Button"
             type: StatusBaseButton.Type.Primary
+
+            StatusToolTip {
+                visible: parent.hovered
+                text: "Look I'm a tooltip on a button!"
+            }
+            onClicked: console.warn("Primary button clicked")
         }
 
         StatusButton {
             text: "Button"
             enabled: false
             type: StatusBaseButton.Type.Primary
+            onClicked: console.warn("Primary disabled button clicked, this should not happen !!!")
         }
 
         StatusButton {
             text: "Button"
             loading: true
             type: StatusBaseButton.Type.Primary
+
+            StatusToolTip {
+                visible: parent.hovered
+                text: "Look I'm a tooltip on a loading button!"
+            }
+            onClicked: console.warn("Primary loading button clicked, this should not happen !!!")
         }
 
         // Large
@@ -50,6 +63,25 @@ Column {
             loading: true
         }
 
+        // Large + Icon
+        StatusButton {
+            text: "Button"
+            icon.name: "info"
+        }
+
+        StatusButton {
+            text: "Button"
+            enabled: false
+            icon.name: "info"
+        }
+
+        StatusButton {
+            text: "Button"
+            loading: true
+            icon.name: "info"
+        }
+
+        // Danger
         StatusButton {
             text: "Button"
             type: StatusBaseButton.Type.Danger
@@ -67,6 +99,7 @@ Column {
             type: StatusBaseButton.Type.Danger
         }
 
+        // Flat
         StatusFlatButton {
             text: "Button"
         }
@@ -74,7 +107,6 @@ Column {
         StatusFlatButton {
             text: "Button"
             enabled: false
-
         }
 
         StatusFlatButton {
@@ -82,6 +114,7 @@ Column {
             loading: true
         }
 
+        // Flat + Danger
         StatusFlatButton {
             text: "Button"
             type: StatusBaseButton.Type.Danger
@@ -117,6 +150,7 @@ Column {
             loading: true
         }
 
+        // Small + danger
         StatusButton {
             text: "Button"
             type: StatusBaseButton.Type.Danger
@@ -137,6 +171,7 @@ Column {
             loading: true
         }
 
+        // Flat + small
         StatusFlatButton {
             text: "Button"
             size: StatusBaseButton.Size.Small
@@ -359,7 +394,7 @@ Column {
     // Button with emoji
     StatusButton {
         text: "Button with Emoji"
-        icon.emoji: "🖼️️"
+        asset.emoji: "🖼️️"
     }
 
     RowLayout {
--- a/src/StatusQ/Controls/StatusBaseButton.qml
+++ b/src/StatusQ/Controls/StatusBaseButton.qml
@@ -1,11 +1,14 @@
 import QtQuick 2.14
+import QtQuick.Layouts 1.14
+import QtQuick.Controls 2.14
+
 import StatusQ.Core 0.1
 import StatusQ.Core.Theme 0.1
 import StatusQ.Components 0.1
 import StatusQ.Core.Utils 0.1
 
-Rectangle {
-    id: statusBaseButton
+Button {
+    id: root
 
     enum Size {
         Tiny,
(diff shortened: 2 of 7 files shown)


src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
-            case StatusBaseButton.Size.Large:
-            default:
-              return 11
+        case StatusBaseButton.Size.Tiny:
+            return 5
+        case StatusBaseButton.Size.Small:
+            return 10
+        case StatusBaseButton.Size.Large:
+        default:
+            return 11
```
caybro, line 57: This could really just `return defaultTopPadding` as they're the same :)

caybro (changes requested):
(almost) perfect! :) So many thanks for this

0x-r4bbit:
Unfortunately, I can't run sandbox due to status-im/status-desktop#7551 so I have to trust @caybro 's judgment here.
The one thing I note right out of the bat is that this is a breaking change, so the commit should make this very explicit.
Screenshot from 2022-07-28 17-35-26

Other than that it'll need a lot of testing since we use this component all over the place.

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
-        height: 24
-    }
+    property StatusAssetSettings asset: StatusAssetSettings { }
 
     property bool loading: false
 
-    property alias hovered: sensor.containsMouse
+    property color normalColor
+    property color hoverColor
+    property color disaledColor
```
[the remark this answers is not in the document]
alexandraB99: that was there from before, so updated all other places used :D

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
         }
     }
 
+    implicitWidth: contentItem.width + root.leftPadding + root.rightPadding
+    implicitHeight: contentItem.height + root.topPadding + root.bottomPadding
+    enabled: !loading
+    leftPadding: defaultLeftPadding
+    rightPadding: defaultRightPadding
+    topPadding: defaultTopPadding
+    bottomPadding: defaultBottomPadding
```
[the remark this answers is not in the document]
alexandraB99: I changed it a bit, based on the documention, please have a second look! Above the very left button is set to have 
```
            horizontalPadding: 40
            verticalPadding: 80
```
and the reset the default values

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
-
-    radius: size !== StatusBaseButton.Size.Tiny ? 8 : 6
-
-    color: {
-        if (statusBaseButton.enabled)
-            return sensor.containsMouse || highlighted ? hoverColor
-                                                       : normalColor;
-        return disaledColor
+    background: Rectangle {
+        anchors.fill: parent
```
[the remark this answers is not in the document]
alexandraB99: it does, it's personal taste but I can remove it

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
+        color: {
+            if (root.enabled)
+                return root.hovered || highlighted ? hoverColor : normalColor;
+            return disaledColor
+        }
+        MouseArea {
+            id: mouseArea
+            anchors.fill: parent
+            cursorShape: loading ? Qt.ArrowCursor : Qt.PointingHandCursor;
+            onPressed: mouse.accepted = false;
```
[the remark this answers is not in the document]
alexandraB99: done

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
-    MouseArea {
-        id: sensor
-        anchors.fill: parent
-        cursorShape: loading ? Qt.ArrowCursor
-                             : Qt.PointingHandCursor
-
-        hoverEnabled: true
-        enabled: !loading && statusBaseButton.enabled
+    contentItem: Item {
+        id: contentItem
```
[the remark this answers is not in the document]
alexandraB99: yeah, same reason as for the `anchors.fill: parent` in `background`, removed :)

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
             sourceComponent: StatusLoadingIndicator {
                 color: d.textColor
-            } // Indicator
-        } // Loader
+            }
+        }
 
         Row {
             id: layout
             anchors.centerIn: parent
```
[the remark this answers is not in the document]
alexandraB99: I guess it's the same since implicitWidth/implicitHeight are calculated based on the content item's dimensions plus the height, as an example here below the very left button has `leftPadding: 0` and the rest the default values:
jj

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
-                emojiId: Emoji.iconId(statusBaseButton.icon.emoji, statusBaseButton.icon.emojiSize) || ""
-            } // Emoji
+                visible: root.asset.emoji
+                emojiId: Emoji.iconId(root.asset.emoji, root.asset.emojiSize) || ""
+            }
             StatusBaseText {
                 id: label
                 opacity: !loading
                 anchors.verticalCenter: parent.verticalCenter
                 font.pixelSize: size === StatusBaseButton.Size.Large ? 15 : 13 // by design
```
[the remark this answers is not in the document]
alexandraB99: true! done

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -18,159 +18,119 @@ Rectangle {
...
+    property color borderColor: "transparent"
 
     property int size: StatusBaseButton.Size.Large
     property int type: StatusBaseButton.Type.Normal
 
-    property alias text: label.text
-    property alias font: label.font
-
     property real defaultLeftPadding: size === StatusBaseButton.Size.Large ? 24 : 12
     property real defaultRightPadding: size === StatusBaseButton.Size.Large ? 24 : 12
```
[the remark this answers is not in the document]
alexandraB99: removed, used built in properties instead

src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: done.

caybro (approved):

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,131 +20,110 @@ Rectangle {
...
-
-    radius: size !== StatusBaseButton.Size.Tiny ? 8 : 6
-
-    color: {
-        if (statusBaseButton.enabled)
-            return sensor.containsMouse || highlighted ? hoverColor
-                                                       : normalColor;
-        return disaledColor
+    implicitWidth: (contentItem.width + root.leftPadding + root.rightPadding)
+    implicitHeight: (contentItem.height + root.topPadding + root.bottomPadding)
```
caybro, line 56: You don't need to specify these

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,131 +20,110 @@ Rectangle {
...
-              return 5
+                return 5;
             case StatusBaseButton.Size.Small:
-              return 10
+                return 10;
             case StatusBaseButton.Size.Large:
             default:
-              return 11
+                return 11;
+            }
```
caybro, line 51: Just thinking... those "default" values, shouldn't they be readonly and hidden behind some d-pointer?

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,131 +20,110 @@ Rectangle {
...
-                                                       : normalColor;
-        return disaledColor
+    implicitWidth: (contentItem.width + root.leftPadding + root.rightPadding)
+    implicitHeight: (contentItem.height + root.topPadding + root.bottomPadding)
+    verticalPadding: 0
+    horizontalPadding: 0
+    leftPadding: defaultHorizontalPadding
+    rightPadding: defaultHorizontalPadding
+    topPadding: defaultVerticalPadding
+    bottomPadding: defaultVerticalPadding
```
caybro, line 62: You don't need these 4 paddings

src/StatusQ/Controls/StatusBaseButton.qml
caybro: Ah, could be then much simpler: just rename s/defaultHorizontalPadding/horizontalPadding and be done :) Same for vertical

caybro (changes requested):
Just some things to even further simplify the implementation

src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: how come?

src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: regarding paddings, for example we could leave it as:
```
    leftPadding: size === StatusBaseButton.Size.Large ? 24 : 12
    rightPadding: size === StatusBaseButton.Size.Large ? 24 : 12
    bottomPadding: {
        switch (size) {
        case StatusBaseButton.Size.Tiny:
            return 5;
        case StatusBaseButton.Size.Small:
            return 10;
        case StatusBaseButton.Size.Large:
        default:
            return 11;
        }
    }
    topPadding: {
        switch (size) {
        case StatusBaseButton.Size.Tiny:
            return 5;
        case StatusBaseButton.Size.Small:
            return 10;
        case StatusBaseButton.Size.Large:
        default:
            return 11;
        }
    }

    implicitWidth: (contentItem.width + root.leftPadding + root.rightPadding)
    implicitHeight: (contentItem.height + root.topPadding + root.bottomPadding)
```
but then if the user sets horizontal or vertical paddings, nothing changes.

src/StatusQ/Controls/StatusBaseButton.qml
caybro: They are calculated automatically from the contentItem/padding, try it :)

src/StatusQ/Controls/StatusBaseButton.qml
caybro: Yeah but since left/right and top/bottom are the same, you could specify horizontalPadding and verticalPadding only

src/StatusQ/Controls/StatusFloatingButtonsSelector.qml
```diff
@@ -113,8 +113,8 @@ Row {
             implicitHeight: 32
             topPadding: 8
             bottomPadding: 0
-            defaultLeftPadding: 4
-            defaultRightPadding: 4
+            leftPadding: 4
+            rightPadding: 4
```
caybro, line 117: Just horizontalPadding :)

PascalPrecht:
Raising awareness of https://github.com/status-im/StatusQ/pull/885#issuecomment-1239281664 to ensure this PR keeps feature parity

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,158 +20,106 @@ Rectangle {
...
-              return 11
+                vertPadding = 22;
+                break;
+            }
+            return (vertPadding + root.verticalPadding);
         }
     }
 
+    implicitWidth: (layout.width + d.defaultHorizontalPadding)
+    implicitHeight: (layout.height + d.defaultVerticalPadding)
```
caybro, line 61: Drop those... they are calculated automatically from the `contentItem` + `*padding`

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,158 +20,106 @@ Rectangle {
...
-              return 11
-        }
-    }
-    property real defaultBottomPadding: {
-        switch (size) {
+    QtObject {
+        id: d
+        readonly property color textColor: root.enabled ? root.textColor : root.disabledTextColor
+        property int defaultHorizontalPadding: (((size === StatusBaseButton.Size.Large) ? 48 : 24) + root.horizontalPadding)
+        property int defaultVerticalPadding: {
```
caybro, line 42: No :) You don't need these `defaultHorizontalPadding/defaultVerticalPadding` here in the `d` object, just put them at the root level:

```qml
Button {
   horizontalPadding: foo
   verticalPadding: bar
```

caybro (changes requested):

src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: the only thing I see if we do this, is that if user sets for example horizontal and vertical padding to 10, this is what they get:
f
So I guess it's not bad to keep some default (and fix) sizes, right?

caybro: It's because your `contentItem` doesn't have an implicitSize (you're just using a plain Item with a Row inside which has `anchors.centerIn`).

alexandraB99 (commented):
@caybro thanks for looking into this! as a first check and testing looks good! just one small comment.

src/StatusQ/Controls/StatusFlatButton.qml
```diff
@@ -3,19 +3,15 @@ import StatusQ.Core 0.1
...
+    disabledColor: "transparent"
 
     textColor: type === StatusBaseButton.Type.Normal ? Theme.palette.primaryColor1
                                                      : Theme.palette.dangerColor1
     disabledTextColor: Theme.palette.baseColor1
 
-    border.color: type === StatusBaseButton.Type.Normal || hovered ? "transparent"
-                                                          : Theme.palette.baseColor2
-    rightPadding: icon.name !== "" ? 18 : defaultRightPadding
-    leftPadding: icon.name !== "" ? 14 : defaultLeftPadding
```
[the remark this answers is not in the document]
caybro: Yes, it's actually when there _is_ an icon, and it's here:
```qml
    horizontalPadding: {
        if (root.icon.name) {
            return size === StatusBaseButton.Size.Large ? 18 : 16
        }
    }
```

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,158 +22,109 @@ Rectangle {
...
+    // stop the mouse clicks in the "loading" state w/o disabling the whole button
+    // as this would make it impossible to have hover events or a tooltip
+    MouseArea {
+        anchors.fill: parent
+        acceptedButtons: Qt.AllButtons
+        enabled: root.loading
+        hoverEnabled: enabled
+        onPressed: mouse.accepted = true
+        onWheel: wheel.accepted = true
+    }
```
PascalPrecht, line 129: Does this ensure that, when `loading`, there are no mouse related events emitted (clicked, pressed, etc) ?

alexandraB99: the `MouseArea` here is set to `enabled: root.loading`, @caybro shouldn't this be `enabled: !root.loading` and also shouldn't the `Button` be set to `enabled: !root.loading` as well?

src/StatusQ/Controls/StatusFlatButton.qml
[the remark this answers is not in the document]
alexandraB99: I overlooked at the code, thanks!

src/StatusQ/Controls/StatusBaseButton.qml
caybro: Pascal is right, this MouseArea is here to prevent the mouse items when the button is in the "loading" state; you can check using the sandbox app; no console msg should be printed when you click those

PascalPrecht: Would love to check in sandbox, but can't run it anymore :D 

I trust you peoples that it works as intended!

0x-r4bbit (approved):

micieslak (approved):
lgtm

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,158 +22,109 @@ Rectangle {
...
+
+    // stop the mouse clicks in the "loading" state w/o disabling the whole button
+    // as this would make it impossible to have hover events or a tooltip
+    MouseArea {
+        anchors.fill: parent
+        acceptedButtons: Qt.AllButtons
+        enabled: root.loading
+        hoverEnabled: enabled
+        onPressed: mouse.accepted = true
+        onWheel: wheel.accepted = true
```
alexandraB99, line 128: is seems `cursorShape: loading ? Qt.ArrowCursor  : Qt.PointingHandCursor` is deleted somehow, could you bring it back?

src/StatusQ/Controls/StatusBaseButton.qml
alexandraB99: @caybro actually I put this mouse area there to get the pointing hand cursor, as there is no way currently to change the cursor shape of the QQC2 component, that's why it was inside the background property, however it can be also used to block events, but we should put the cursorShape back

caybro: This can't work imho, esp. together with the fact we want the tooltip to appear eventhough we block the mouse events in the "loading" state

igor-sirotin (commented):
Nice 👍 
Just a small question about `icon` and `asset` property.

alexandraB99 (commented):
Apologies, I overlooked the code :) please rebase and merge

sandbox/controls/Buttons.qml
```diff
@@ -21,18 +21,31 @@ Column {
...
+                text: "Look I'm a tooltip on a button!"
+            }
+            onClicked: console.warn("Primary button clicked")
         }
 
         StatusButton {
             text: "Button"
             enabled: false
             type: StatusBaseButton.Type.Primary
+            onClicked: console.warn("Primary disabled button clicked, this should not happen !!!")
```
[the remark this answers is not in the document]
caybro: It's just the sandbox and I don't think it makes any actual difference (at least in QML)

src/StatusQ/Controls/StatusBaseButton.qml
```diff
@@ -19,158 +22,109 @@ Rectangle {
...
-        onPressed: {
-            if (!loading) {
-                statusBaseButton.pressed(mouse)
-            }
+    contentItem: RowLayout {
+        spacing: root.size === StatusBaseButton.Size.Large ? 6 : 4
+        StatusIcon {
+            Layout.preferredWidth: visible ? root.icon.width : 0
+            Layout.preferredHeight: visible ? root.icon.height : 0
+            icon: root.icon.name
```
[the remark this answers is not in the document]
caybro: `icon` in StatusIcon is really just the `icon.name` (being an `Image`)

PR #816, opened 2022-08-01
fix(Utils): moved emojilists.js instance inside Emoji singleton

alexandraB99:
Needed for: https://github.com/status-im/status-desktop/issues/6729

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Core/Utils/Emoji.qml
+++ b/src/StatusQ/Core/Utils/Emoji.qml
@@ -19,7 +19,7 @@ QtObject {
         "svg": "svg"
     }
     property string base: Qt.resolvedUrl("../../../assets/twemoji/")
-
+    property var emojiJSON: EmojiJSON
     function parse(text, renderSize = size.small, renderFormat = format.svg) {
         const renderSizes = renderSize.split("x");
         if (!renderSize.includes("x") || renderSizes.length !== 2) {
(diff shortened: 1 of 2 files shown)


caybro (approved):

stefandunca (approved):

src/StatusQ/Core/Utils/Emoji.qml
```diff
@@ -19,7 +19,7 @@ QtObject {
         "svg": "svg"
     }
     property string base: Qt.resolvedUrl("../../../assets/twemoji/")
-
+    property var emojiJSON: EmojiJSON
```
PascalPrecht, line 22: Probably a stupid questoin: Where does `EmojiJSON` come from now? I guess the answer lies in it being a singleton instance? Is this a breaking change?

alexandraB99: @PascalPrecht EmojiJSON is imported above `import "./emojiList.js" as EmojiJSON` here we just create an instance of it so that it can be referenced in the EmojiPopup, see https://github.com/status-im/status-desktop/pull/6745/files#diff-298b4f701bf6ed9246dd67854beec455eb9c421be43554248043dea78c15da39

PascalPrecht: Ah thanks for clarifying!
Just to be clear though: without the change in Status Desktop, that will break?

alexandraB99: @PascalPrecht as far as I can see, this is only used in `StatusEmojiPopup.qml`
em

PascalPrecht: Cool, thanks for veriyfing!

PascalPrecht (approved):

PR #825, opened 2022-08-03
chore(StatusChatList): Keep sync with underlying model after reordering

micieslak:
### Checklist

After reordering chats or chat categories the order of elements in the list should reflect order in the underlying model. Reorders done via DelegateModel when dragging reordered item should be revoked to avoid improper sorting.

required by: status-im/status-desktop#6597, status-im/status-desktop#6722

affected:
StatusChatList, StatusChatListAndCategories

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusChatList.qml
+++ b/src/StatusQ/Components/StatusChatList.qml
@@ -9,7 +9,7 @@ import StatusQ.Components 0.1
 import StatusQ.Controls 0.1
 
 Column {
-    id: statusChatList
+    id: root
 
     spacing: 4
     width: 288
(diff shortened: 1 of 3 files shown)


src/StatusQ/Components/StatusChatList.qml
```diff
@@ -33,6 +33,27 @@ Column {
         }
     }
 
+    QtObject {
+        id: internal
```
MishkaRogachev, line 37: we usually name this internal QObject as`d`

src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -91,13 +91,30 @@ Item {
...
+
                 model: SortFilterProxyModel {
                     sourceModel: statusChatListAndCategories.model
 
                     filters: ValueFilter { roleName: "isCategory"; value: true }
                     sorters: RoleSorter { roleName: "position" }
                 }
 
+                // undo any reorders done in a DelegateModel
+                function resetDelegateModel() {
```
MishkaRogachev, line 104: There are two quite-similar functions, can we move it to js helper?

micieslak:
Thanks @MishkaRogachev for the review. I fixed what you mentioned but also changed the approach because I noticed some issues with the previous one when new items are added to the list. Now it seems to work fine and is slightly more "standard".

MishkaRogachev (approved):
Looks solid!

caybro (approved):
Changes look good

I just don't think this DelegateModel is a viable choice for the future :) So much imperative code

osmaczko (approved):

PR #829, opened 2022-08-04
fix(StatusImageCropPanel) vertically align slider in ImageCropPanel

faust4exe:
- change StatusSlider y to be vertically centered

Fixes [#](https://github.com/status-im/status-desktop/issues/6725

Required by https://github.com/status-im/status-desktop/pull/6811

### Checklist

- [ ] update sandbox app

StatusSlider after changes
Screenshot from 2022-08-05 12-12-26
Screenshot from 2022-08-05 12-12-31

StatusImageCropPanel after changes
Screenshot from 2022-08-05 12-11-43
Screenshot from 2022-08-05 12-11-49

Status-desktop fixed
Screenshot from 2022-08-05 12-22-38

### DIFF
--- a/src/StatusQ/Controls/StatusSlider.qml
+++ b/src/StatusQ/Controls/StatusSlider.qml
@@ -31,7 +31,8 @@ Slider {
         id: bgRect
 
         x: root.leftPadding
-        y: root.topPadding
+        anchors.verticalCenter: root.verticalCenter
+
         implicitWidth: 100
         implicitHeight: bgHeight
         width: root.availableWidth
--- a/src/StatusQ/Components/StatusImageCropPanel.qml
+++ b/src/StatusQ/Components/StatusImageCropPanel.qml
@@ -242,13 +242,14 @@ Item {
             StatusIcon {
                 icon: "remove-circle"
 
+                Layout.alignment: Qt.AlignVCenter
                 Layout.preferredWidth: 20
                 Layout.preferredHeight: 20
             }
             StatusSlider {
                 Layout.fillWidth: true
                 Layout.topMargin: 20
-                Layout.bottomMargin: 25
+                Layout.bottomMargin: 20
                 Layout.alignment: Qt.AlignVCenter
 
                 enabled: root.interactive
@@ -262,6 +263,7 @@ Item {
             StatusIcon {
                 icon: "add-circle"
 
+                Layout.alignment: Qt.AlignVCenter
                 Layout.preferredWidth: 20
                 Layout.preferredHeight: 20
             }


caybro (approved):

src/StatusQ/Controls/StatusSlider.qml
```diff
@@ -31,7 +31,7 @@ Slider {
         id: bgRect
 
         x: root.leftPadding
-        y: root.topPadding
+        y: root.topPadding + root.availableHeight / 2 - height / 2
```
alexandraB99, line 34: could you add parenthesis just to be sure the this is interpreted as we want? I've seen cases where the calculations were weird because of "free will" interpretation, for instance it was calculating (root.availableHeight / (2 - height / 2) and so on...

faust4exe: Wow :O - never heart of "free will" interpretation - sounds crazy and I dont remember to meet it even I had pretty complex constructions. But ok I will write it as `topPaddin + (available - height) / 2` , hope in case of "free will" there are no other interpretations.

caybro: Well it definitely helps readability and that's a good thing. It doesn't add any overhead

src/StatusQ/Controls/StatusSlider.qml
```diff
@@ -31,7 +31,7 @@ Slider {
         id: bgRect
 
         x: root.leftPadding
-        y: root.topPadding
+        y: root.topPadding + (root.availableHeight - height) / 2
```
stefandunca, line 34: How about having pixel alignment?
```suggestion
        y: root.topPadding + ((root.availableHeight - height) / 2).toFixed()
```

stefandunca (approved):
Looks good. How about fixing the colour of the zoom buttons also with this PR?

alexandraB99 (approved):

src/StatusQ/Controls/StatusSlider.qml
faust4exe: This is a good point. But should we consider that topPadding is real value?

faust4exe: @alexandraB99 when you have free time - check @stefandunca  suggested code - results are very interesting :D

faust4exe:
> Looks good. How about fixing the colour of the zoom buttons also with this PR?

Done

src/StatusQ/Controls/StatusSlider.qml
caybro: > This is a good point. But should we consider that topPadding is real value?

It is a valid concern and we should strive for getting this pixel aligned

caybro (approved):

caybro:
@faust4exe have a look at https://github.com/status-im/status-desktop/issues/6835 as well

faust4exe:
> @faust4exe have a look at [status-im/status-desktop#6835](https://github.com/status-im/status-desktop/issues/6835) as well

Thanks - my bad - for some reason I took a look only on issues from StatusQ

faust4exe:
> @faust4exe have a look at [status-im/status-desktop#6835](https://github.com/status-im/status-desktop/issues/6835) as well

Ok - I was told that this Issue should go to a different PR

faust4exe:
> Looks good. How about fixing the colour of the zoom buttons also with this PR?

This one is part of Issue mentioned by @caybro and will be done in separate PR.

alexandraB99 (approved):

src/StatusQ/Controls/StatusSlider.qml
stefandunca: > This is a good point. But should we consider that topPadding is real value?

I consider that the user of the control has the responsibility to sanitize the values and do it efficiently (only when you know might be needed).

So anything coming from the interface can be assumed that is valid. Same in the provided example, if we know that all values are multiple of two (due to dual spacing or another way) then should not be sanitized. We already had some issues hard to track/fix due to pixel misalignment (blurry emojis, some rare corner cases in redrawing) so taking care to stay aligned should be considered.

> @alexandraB99 when you have free time - check @stefandunca suggested code - results are very interesting :D "Free will" should never be mentioned as it appears when you mention it :)

I had a quick look and the way `toFixed()` generates this side effect of malforming the value is concerning, maybe a Qt JS compiler bug/"optimization". I have never experienced it but plan to follow up later on.

caybro (approved):

PR #831, opened 2022-08-04
fix(StatusInput): Add labelPadding customisation and improve height calculation

MishkaRogachev:
Screenshot 2022-08-04 at 17 34 24

P.S i've fixed error message display in the component but there is still problem with Utils.getErrorMessage

### Checklist

- [ ] update sandbox app
- [x ] test changes in [status-desktop](https://github.com/status-im/status-desktop)

### DIFF
--- a/src/StatusQ/Controls/StatusInput.qml
+++ b/src/StatusQ/Controls/StatusInput.qml
@@ -94,6 +94,11 @@ Item {
         \qmlproperty string StatusInput::label
         This property sets the label text.
     */
+    property int labelPadding: 8
+    /*!
+        \qmlproperty string StatusInput::label
+        This property sets the label text.
+    */
     property string label: ""
     /*!
         \qmlproperty string StatusInput::secondaryLabel
@@ -347,9 +352,7 @@ Item {
     }
 
     implicitWidth: 448
-    implicitHeight: ((internal.inputHeight + topRow.height + errorMessage.height) +
-                    ((topRow.height > 0) && (errorMessage.height > 0) ? 16 :
-                    (topRow.height > 0) || (errorMessage.height > 0) ? 8 : 0))
+    implicitHeight: inputLayout.implicitHeight
 
     Component.onCompleted: {
         validate()
@@ -363,7 +366,8 @@ Item {
         RowLayout {
             id: topRow
             Layout.fillWidth: true
-            Layout.preferredHeight: (!!root.label || !!root.secondaryLabel || root.charLimit > 0) ? 22 :0
+            Layout.preferredHeight: (!!root.label || !!root.secondaryLabel || root.charLimit > 0) ? 22 : 0
+
             StatusBaseText {
                 id: label
                 visible: !!text
@@ -404,7 +408,7 @@ Item {
             implicitWidth: parent.width
             implicitHeight: internal.inputHeight
             Layout.alignment: Qt.AlignTop
-            Layout.topMargin: (topRow.height > 0) ? 8 : 0
+            Layout.topMargin: (topRow.height > 0) ? labelPadding : 0
             maximumLength: root.charLimit
             onTextChanged: root.validate()
             Keys.forwardTo: [root]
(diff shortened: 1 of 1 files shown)


src/StatusQ/Controls/StatusInput.qml
```diff
@@ -419,13 +423,13 @@ Item {
 
         StatusBaseText {
             id: errorMessage
-            visible: !!text && !statusBaseInput.valid
-            height: visible ? contentHeight : 0
+            visible: !statusBaseInput.valid && statusBaseInput.dirty && !!text
```
jrainville, line 426: This depends on the type of validation. The always mode doesn't depend on the dirty state

jrainville (commented):
Looks good, I just think the error message visibility needs to be checked

src/StatusQ/Controls/StatusInput.qml
MishkaRogachev: Ok, i'll add this check

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -347,9 +352,7 @@ Item {
     }
 
     implicitWidth: 448
-    implicitHeight: ((internal.inputHeight + topRow.height + errorMessage.height) +
-                    ((topRow.height > 0) && (errorMessage.height > 0) ? 16 :
-                    (topRow.height > 0) || (errorMessage.height > 0) ? 8 : 0))
+    implicitHeight: inputLayout.implicitHeight
```
alexandraB99, line 356: could you please let me know how this works, `StatusInput` takes it's height from `ColumnLayout` but  `ColumnLayout` is set to `anchors.fill: parent`?

alexandraB99 (changes requested):

src/StatusQ/Controls/StatusInput.qml
MishkaRogachev: StatusInput takes `implicit` size from ColumnLayout what is calculated from implicit sizes of layout's children, spacing and margins. So there is no need to calculate it manually. `anchors.fill: parent` affects explicit size (width and height properties), not implicit one. [Here](https://stackoverflow.com/questions/45946808/difference-between-width-height-and-implicitwidth-height-and-corresponding-use) is good answer.

src/StatusQ/Controls/StatusInput.qml
MishkaRogachev: Screenshot 2022-08-05 at 14 01 14
 @jrainville for me it's a bit strange that input borders are blue when `ValidationMode.Always`

src/StatusQ/Controls/StatusInput.qml
caybro: Are you sure a Column/Layout has an implicitHeight?

alexandraB99: it looks and feels weird. How about something like:
```
        width: parent.width
        height: childrenRect.height
```
instead of `anchors.fill: parent`?

src/StatusQ/Controls/StatusInput.qml
jrainville: I think it's blue in this case because we are focusing on it? If you click on another input, does it show red?

src/StatusQ/Controls/StatusInput.qml
```diff
@@ -419,13 +423,22 @@ Item {
...
             id: errorMessage
-            visible: !!text && !statusBaseInput.valid
-            height: visible ? contentHeight : 0
+            visible: {
+                if (!text)
+                    return false;
+
+                if ((root.validationMode == StatusInput.ValidationMode.OnlyWhenDirty && statusBaseInput.dirty) ||
+                    root.validationMode == StatusInput.ValidationMode.Always)
+                return !statusBaseInput.valid;
```
jrainville, line 432: This needs a tab I think, or brackets, to show it's with the if above.

src/StatusQ/Controls/StatusInput.qml
MishkaRogachev: @caybro yes, otherwise it was not working :) @alexandraB99 `height: childrenRect.height` can cause a binding loop. 
I'm strongly believe that using implicit sizes of reusable components is a good practice because we always have two flows and mixing them together won't work properly in every places. And this is actually the reason that we getting problems with layouts. Most modern QML libraries like QQC2 and QQL using implicit sizing flow. And it's mentioned in the [documentation](https://doc.qt.io/qt-6/qtquicklayouts-overview.html).

src/StatusQ/Controls/StatusInput.qml
MishkaRogachev: fixed

src/StatusQ/Controls/StatusInput.qml
caybro: Ok, I'll trust you on that :) Otherwise another :beer: :D

src/StatusQ/Controls/StatusInput.qml
MishkaRogachev: No, border is defined by StatusBaseInput and it's not aware of ValidationMode. Imho we should managed valid flag in StatusInput and move there logic from StatusBaseInput

jrainville (approved):

src/StatusQ/Controls/StatusInput.qml
MishkaRogachev: @caybro can you please approve then?)

PR #832, opened 2022-08-04
fix(StatusChatListCategory): Clicking on a category should toggle it

stefandunca:
The culprit here is at the very bottom, in StatusListItem: the toplevel
item was a MouseArea handling the clicks and this causes problems as any
extra buttons placed on top of it get their mouseClick events delivered
in wrong order (after StatusListItem). Fix this by having a MouseArea
handling click events behind the actual toplevel item as a last resort,
catch all handler  plus the actual "sensor" being just a MouseArea that
handles merely the hover events

- drop unneeded onPressed/onReleased handlers in StatusChatListCategoryItem
- fix some warnings (typos, unreachable code, shadowed variables)

Fixes: status-im/status-desktop#6733

### Checklist

- [x] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusChatListCategory.qml
+++ b/src/StatusQ/Components/StatusChatListCategory.qml
@@ -37,20 +37,20 @@ Column {
         title: statusChatListCategory.name
         opened: statusChatListCategory.opened
         sensor.pressAndHoldInterval: 150
-
+        propagateTitleClicks: true // title click is handled as a normal click (fallthru)
         showMenuButton: showActionButtons && !!statusChatListCategory.popupMenu
         highlighted: statusChatListCategory.dragged
-        sensor.onClicked: {
+        onClicked: {
             if (sensor.enabled) {
                 if (mouse.button === Qt.RightButton && showActionButtons && !!statusChatListCategory.popupMenu) {
                     highlighted = true;
                     popupMenuSlot.item.popup(mouse.x + 4, mouse.y + 6);
-                    return
+                } else if (mouse.button === Qt.LeftButton) {
+                    statusChatListCategory.opened = !statusChatListCategory.opened
                 }
             }
         }
-        onTitleClicked: statusChatListCategory.opened = !opened
-        onToggleButtonClicked: statusChatListCategory.opened = !opened
+        onToggleButtonClicked: statusChatListCategory.opened = !statusChatListCategory.opened
         onMenuButtonClicked: {
             highlighted = true
             menuButton.highlighted = true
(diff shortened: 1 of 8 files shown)


src/StatusQ/Components/StatusChatListCategory.qml
```diff
@@ -41,7 +41,11 @@ Column {
         showMenuButton: showActionButtons && !!statusChatListCategory.popupMenu
         highlighted: statusChatListCategory.dragged
         sensor.onClicked: {
-            if (sensor.enabled) {
+            if (mouse.button === Qt.LeftButton) {
+                statusChatListCategory.opened = !opened
+                mouse.accepted = true
```
caybro, line 46: Don't want to spoil the party but the Qt docs say:

> When handling this signal, changing the [accepted](https://doc.qt.io/qt-5/qml-qtquick-mouseevent.html#accepted-prop) property of the mouse parameter has no effect, unless the [propagateComposedEvents](https://doc.qt.io/qt-5/qml-qtquick-mousearea.html#propagateComposedEvents-prop) property is true.

stefandunca: Thanks for the heads up. Just tested again and it doesn't change anything. It is still ignored and all events are propagated from all the children for those controls regardless of `propagateComposedEvents`/`accepted` and `preventStealing` combination. It looks like something breaks the `MouseArea`s expected behaviour and in that mix of controls in controls and parenting/enablind disabling of mouse areas workaround is hard to tell which one.

caybro: @stefandunca yeah, avoid nested MouseAreas :) Try with TapHandler instead, it should work

stefandunca: @caybro I was referring to that. I tried it and it makes no difference, see the comment in PR: https://github.com/status-im/status-desktop/issues/6733#issuecomment-1206614512

src/StatusQ/Components/StatusListItem.qml
```diff
@@ -23,7 +23,6 @@ Rectangle {
     property string titleTextIcon: ""
     property real leftPadding: 16
     property real rightPadding: 16
-    property bool enabled: true
```
caybro, line 26: Self note: this was a pain to discover :)

PascalPrecht: This is a breaking change and should be denoted as such.
Also gotta make sure this will be taken into account in desktop

caybro: It's not a breaking change; it was just wrongly shadowing an already existing property (https://doc.qt.io/qt-5/qml-qtquick-item.html#enabled-prop), this is just fixing it

PascalPrecht: Ah cool!

PascalPrecht (approved):

stefandunca:
Thanks, @caybro for the great fix

PR #849, opened 2022-08-11
test(StatusChatListAndCategories) add object names for testing

stefandunca:
Updates #6751

### Checklist

- [x] ~~add documentation if necessary (new component, new feature)~~
- [x] ~~update sandbox app~~
- [x] ~~test changes in both light and dark theme?~~
- [x] ~~is this a breaking change?~~
- [x] ~~test changes in [status-desktop](https://github.com/status-im/status-desktop)~~

### DIFF
--- a/src/StatusQ/Components/StatusChatListAndCategories.qml
+++ b/src/StatusQ/Components/StatusChatListAndCategories.qml
@@ -71,6 +71,7 @@ Item {
             spacing: 4
 
             StatusChatList {
+                objectName: "statusChatListAndCategoriesChatList"
                 id: statusChatList
                 visible: statusChatList.model.count > 0
                 onChatItemSelected: root.chatItemSelected(categoryId, id)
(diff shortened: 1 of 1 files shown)


saledjenic (approved):

src/StatusQ/Components/StatusChatListAndCategories.qml
```diff
@@ -71,6 +71,7 @@ Item {
             spacing: 4
 
             StatusChatList {
+                objectName: "statusChatList"
```
jrainville, line 74: This name is not precise enough, it would clash with other `statusChatList`s if people name their's like that as well.
Try something like `chatListStatusChatList` or `statusChatListAndCategoriesChatList`. The name is very weird, but that component is very weird I guess, because we have a StatusChatList inside StatusChatListAndCategories

stefandunca: At first, I understood that we can use containers to deconflict object names. Now I see this is not very practical and we don't do it. Fixed it.

jrainville (approved):

PR #857, opened 2022-08-16
fix(StatusChatListAndCategories): duplicate objectName

stefandunca:
Due to repository automerge.

### Checklist


caybro (approved):

caybro:
@stefandunca just wondering whether there are actually 2 tests, each using a different `objectName`

stefandunca:
Now there is none. I updated one of my tests and forgot to remove the objectName. However, in the meantime, another landed, but could not find the usage but I think this is part of @MishkaRogachev's ongoing work.

PR #859, opened 2022-08-17
fix(StatusTabButton): display hand cursor on hover when enabled

micieslak:
Needed for https://github.com/status-im/status-desktop/issues/6787

StatusTabButton gains proper mouse cursor (hand) when active

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Controls/StatusTabButton.qml
+++ b/src/StatusQ/Controls/StatusTabButton.qml
@@ -25,16 +25,20 @@ import StatusQ.Components 0.1
 TabButton {
     id: root
 
-    property alias badge: statusBadge
+    readonly property alias badge: statusBadge
 
     leftPadding: 12
     rightPadding: 12
 
     background: Item { }
 
-    contentItem: Item {
+    contentItem: MouseArea {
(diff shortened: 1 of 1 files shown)


src/StatusQ/Controls/StatusTabButton.qml
```diff
@@ -25,17 +25,21 @@ import StatusQ.Components 0.1
...
 
-    contentItem: Item {
+    contentItem: MouseArea {
         implicitWidth: contentItemGrid.implicitWidth
         implicitHeight: contentItemGrid.implicitHeight + 15
 
+        cursorShape: Qt.PointingHandCursor
+        acceptedButtons: Qt.NoButton
+        enabled: root.enabled
+
```
caybro, line 42: Don't you need `hoverEnabled` here as well?

micieslak: No, `hoverEnabled` is not necessary for handling cursor shape. The is example in official docs for that: https://doc.qt.io/qt-6/qml-qtquick-mousearea.html#cursorShape-prop.

PR #860, opened 2022-08-17
feat(StatusTagSelector): fixed admin / non admin permissions

alexandraB99:
Needed for: https://github.com/status-im/status-desktop/issues/5941

### Checklist

- [x] update sandbox app

### DIFF
--- a/sandbox/demoapp/CreateChatView.qml
+++ b/sandbox/demoapp/CreateChatView.qml
@@ -70,6 +70,12 @@ Page {
             Component.onCompleted: {
                 textEdit.forceActiveFocus();
                 sortModel(root.contactsModel);
+                //add admin tag
+                namesModel.append({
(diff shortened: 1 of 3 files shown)


alexandraB99:
Uploading rer.mov…

noeliaSD (commented):
As discussed before, I'm not sure if it is necessary to update `StatusQ` since `StatusTagItem` allows to setup a `readonly tag + icon` that is enough to categorize a `status desktop admin`. 
For mi the concept of `admin` is not an abstract one and not necessary to define the api of this general component.

PascalPrecht:
I think I'll have to agree with @noeliaSD here.
A UI component doesn't need to know about a notion of "admin".
Applications should configure the component as needed, whether that is in the context of admin permissions, or something else, should not be business of the UI component.

alexandraB99:
@noeliaSD Updated!

alexandraB99:
all fixed [here](https://github.com/status-im/status-desktop/pull/7188)

PR #868, opened 2022-08-25
feat(StatusScrollView): add ensureVisible function

faust4exe:
Required for https://github.com/status-im/status-desktop/issues/7093

### Checklist

- [X] follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] update sandbox app
- [X] test changes in [status-desktop](https://github.com/status-im/status-desktop)

### DIFF
--- a/src/StatusQ/Core/StatusScrollView.qml
+++ b/src/StatusQ/Core/StatusScrollView.qml
@@ -2,6 +2,7 @@ import QtQuick 2.14
 import QtQuick.Controls 2.14
 
 import StatusQ.Controls 0.1
+import StatusQ.Core.Utils 0.1
 
 /*!
    \qmltype StatusScrollView
@@ -62,5 +63,9 @@ Flickable {
     ScrollBar.vertical: StatusScrollBar {
         policy: ScrollBar.AsNeeded
         visible: resolveVisibility(policy, root.height, root.contentHeight)
+    }
+
+    function ensureVisible(rect) {
+        Utils.ensureVisible(this, rect)
     }
 }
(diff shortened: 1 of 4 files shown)


src/StatusQ/Core/StatusScrollView.qml
```diff
@@ -63,4 +63,15 @@ Flickable {
         policy: ScrollBar.AsNeeded
         visible: resolveVisibility(policy, root.height, root.contentHeight)
     }
+
+    function ensureVisible(rect) {
```
micieslak, line 67: We have an identical method in `StatusBaseInput.qml` and sth similar in `StatusSelectableText.qml`. I would suggest to move this method to utils.

micieslak (changes requested):

src/StatusQ/Core/StatusScrollView.qml
faust4exe: @micieslak Yes - I observed that and I believe it is identical. The thing is those functions are also inside of Flickable same as here. So I guess we dont need to move it to utils, but maybe StatusScrollView must be reused in place of those Flickable and have function inside of shared component.

micieslak: I would opt for standalone function in utils taking Flickable as a first argument. It gives slightly more flexibility. On top of that we could add ensureVisible to StatusScrollView as well and use standalone version for implementation:
```
function ensureVisible(rect) {
   StatusUtils.ensureVisible(this, rect)
}
```
Than we are safe in case StatusScrollView is refactored and decided to rely on different top level component than Flickable.

faust4exe: Ok - agree

micieslak (approved):
Looks great!

caybro (approved):
LGTM

PR #871, opened 2022-08-26
feat(StatusMessage): add `messageOriginInfo` property

PascalPrecht:
This is a property that allows for adding information about the origin
of a message (e.g. "Imported from discord").

This will most likely be only useful for messages that have been
imported from any other platform.

Because importing from other platforms requires those message to be
signed by a Status Account, it's undesirable to render the public key of
the account that signed imported messages.

Hence, when there's a `messageOriginInfo`, we don't render the elided
public key.

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Components/private/statusMessage/StatusMessageHeader.qml
+++ b/src/StatusQ/Components/private/statusMessage/StatusMessageHeader.qml
@@ -23,6 +23,7 @@ Item {
     property bool isContact: sender.isContact
     property int trustIndicator: sender.trustIndicator
     property bool amISender: false
+    property string messageOriginInfo: ""
 
     signal clicked(var sender, var mouse)
     signal resendClicked()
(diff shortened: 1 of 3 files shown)


PascalPrecht:
This is what it looks like in action (notice the "Imported from Discord" label):
Screenshot from 2022-08-26 12-23-07

src/StatusQ/Components/private/statusMessage/StatusMessageHeader.qml
```diff
@@ -78,7 +87,7 @@ Item {
         }
         StatusBaseText {
             id: tertiaryDetailText
-            visible: !root.amISender
+            visible: !root.amISender && messageOriginInfo == ""
```
PascalPrecht, line 90: > Because importing from other platforms requires those message to be
signed by a Status Account, it's undesirable to render the public key of
the account that signed imported messages.

> Hence, when there's a messageOriginInfo, we don't render the elided
public key.

PR #879, opened 2022-09-01
feat(StatusListItem): Added new properties to handle inline tags needed in the wallet section

Khushboo-dev-cpp:
feat(StatusListItem): Added new properties to handle inline tags needed in the wallet section

### Checklist

- [x] update sandbox app

### DIFF
--- a/sandbox/controls/ListItems.qml
+++ b/sandbox/controls/ListItems.qml
@@ -576,4 +576,42 @@ ColumnLayout {
         asset.isLetterIdenticon: true
         loadingFailed: true
     }
+
+    StatusListItem {
+        implicitWidth: 600
+        title: "List Item with Tags"
+        subTitle: "03:32"
+        asset.isLetterIdenticon: true
+        inlineTagModel: 6
+        inlineTagDelegate: StatusListItemTag {
+            height: 24
+            title: "tag"
+            asset.isLetterIdenticon: true
+        }
+        components: [
+            ColumnLayout {
+                Row {
+                    Layout.alignment: Qt.AlignRight
+                    spacing: 4
+                    StatusIcon {
+                        color:  Theme.palette.successColor1
+                        icon: "arrow-up"
+                        rotation: 135
+                        height: 18
+                    }
+                    StatusBaseText {
+                        text: "0.0000015 ETH"
+                        font.pixelSize: 15
+                        color: Theme.palette.directColor1
+                    }
+                }
+                StatusBaseText {
+                    Layout.alignment: Qt.AlignRight
+                    text: "1201.10 USD"
+                    font.pixelSize: 15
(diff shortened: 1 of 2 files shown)


PascalPrecht:
Code-wise this looks straight forward... what I've noticed though, is that we already have a `tagsModel`. I assume it's a different thing otherwise you wouldn't introduce a new thing, but I'm wondering if this is getting a little bit out of hand.

Would it be possible to maybe have only one tagsModel and a flag which specificies whether we have "inline" tags or not?
Do we have cases were we want to render both types of tags at the same time?

alexandraB99 (changes requested):
@Khushboo-dev-cpp could you please create a new list item, `StatusTagListItem` or similar please? this StatusListItem is getting very big with 400 lines of code, thus unmaintainable.

Khushboo-dev-cpp:
> Code-wise this looks straight forward... what I've noticed though, is that we already have a `tagsModel`. I assume it's a different thing otherwise you wouldn't introduce a new thing, but I'm wondering if this is getting a little bit out of hand.
> 
> Would it be possible to maybe have only one tagsModel and a flag which specificies whether we have "inline" tags or not? Do we have cases were we want to render both types of tags at the same time?

Hey @PascalPrecht I had thought of the same solution, I dont have a use case right now where we need both, I could add a flag for isInlineTag and update this :)

Khushboo-dev-cpp:
> @Khushboo-dev-cpp could you please create a new list item, `StatusTagListItem` or similar please? this StatusListItem is getting very big with 400 lines of code, thus unmaintainable.

Hey @alexandraB99 ! I agree this StatusListItem is rather growing rapidly, however should I make this separate already? I think with new design system we will have a clear overview of what a list item will be and we can then decide how to separate them? let me know what you think?

alexandraB99 (approved):
let's revise when we got the new design system work

Khushboo-dev-cpp:
> > Code-wise this looks straight forward... what I've noticed though, is that we already have a `tagsModel`. I assume it's a different thing otherwise you wouldn't introduce a new thing, but I'm wondering if this is getting a little bit out of hand.
> > Would it be possible to maybe have only one tagsModel and a flag which specificies whether we have "inline" tags or not? Do we have cases were we want to render both types of tags at the same time?
> 
> Hey @PascalPrecht I had thought of the same solution, I don't have a use case right now where we need both, I could add a flag for isInlineTag and update this :)

Hey @PascalPrecht I was looking at adding the flag I remembered why I didn't do it in the first place, its because adding this is making add many if else cases to handle height etc and it doesn't work perfectly. If you do not think its a very inconvenient to the end user I'd like to keep it as it is

PascalPrecht:
@Khushboo-dev-cpp perfectly fine, I was just throwing out ideas :)

PR #885, opened 2022-09-07
fix(StatusBaseButton): keep `MouseArea` enabled when in `loading` state

PascalPrecht:
`StatusBaseButton` comes with a `MouseArea` that is disabled when the button is set to being disabled. Prior to this commit it's *also* disabled when the button is in `loading` state.

This makes sense because a button that is in `loading` state shouldn't not emit any click signals or trigger hover indications.

There's a scenario though in which we want render a tooltip on top of the button which is in `loading` state. For the tooltip to show, the `MouseArea` of the button needs to be enabled.

Hence, this commit adjust `StatusBaseButton`'s behaviour to *not* disable the `MouseArea` when it's `loading`. Mouse events are already prevented via the `loading` flag. So the only thing left to do is to ensure the button doesn't trigger any hover indication when in `loading` state.

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Controls/StatusBaseButton.qml
+++ b/src/StatusQ/Controls/StatusBaseButton.qml
@@ -89,7 +89,7 @@ Rectangle {
 
     color: {
         if (statusBaseButton.enabled)
-            return sensor.containsMouse || highlighted ? hoverColor
+            return !statusButton.loading && (sensor.containsMouse || highlighted) ? hoverColor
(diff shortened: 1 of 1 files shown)


caybro (approved):
Makes sense, nice!

caybro:
@PascalPrecht however, we need to watch out when the StatusBaseButton gets converted to be a real `Button`

PR #890, opened 2022-09-08
feat(Controls): introduce `StatusPasswordInput`

saledjenic:
`StatusPasswordInput` is a wrapper around  `TextField` with an option to display signing phrase.

### DIFF
--- a/src/StatusQ/Controls/StatusPasswordInput.qml
+++ b/src/StatusQ/Controls/StatusPasswordInput.qml
@@ -1,0 +1,108 @@
+import QtQuick 2.14
+import QtQuick.Layouts 1.12
+import QtQuick.Controls 2.14
+
+import StatusQ.Core 0.1
+import StatusQ.Core.Theme 0.1
+import StatusQ.Controls 0.1
+
+/*!
+   \qmltype StatusPasswordInput
+   \inherits Item
+   \inqmlmodule StatusQ.Controls
+   \since StatusQ.Controls 0.1
+   \brief The StatusPasswordInput control provides a generic user password input with an option to display signing phrase
+
+   Example of how to use it:
+
+   \qml
+        StatusPasswordInput {
+            signingPhrase: "orange hello cygnet"
+            placeholderText: qsTr("Password")
+        }
+   \endqml
+
+   For a list of available components see StatusQ.
+*/
+
+TextField {
+    id: root
+
+    /*!
+        \qmlproperty string StatusPasswordInput::signingPhrase
+        This property sets the signingPhrase of TextField field and signing seed phrase.
+    */
+    property string signingPhrase: ""
+
+    QtObject {
+        id: d
+
+        readonly property int inputTextPadding: 16
+        readonly property int pixelSize: 15
+        readonly property int radius: 8
+        readonly property int signingPhrasePadding: 8
+        readonly property int signingPhraseWordPadding: 8
+        readonly property int signingPhraseWordsSpacing: 8
+        readonly property int signingPhraseWordsHeight: 30
+
+    }
+
+    leftPadding: d.inputTextPadding
+    rightPadding: root.signingPhrase !== ""?
+                      phrase.width + phrase.anchors.leftMargin + phrase.anchors.rightMargin :
+                      d.inputTextPadding
+    verticalAlignment: Text.AlignVCenter
+    implicitWidth: 480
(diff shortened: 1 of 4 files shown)


src/StatusQ/Controls/StatusPasswordInput.qml
```diff
@@ -0,0 +1,145 @@
...
+        readonly property int signingPhraseWordPadding: 8
+        readonly property int signingPhraseWordsSpacing: 8
+        readonly property int signingPhraseWordsHeight: 30
+
+    }
+
+    implicitWidth: 480
+    implicitHeight: 44
+
+    TextField {
```
caybro, line 88: Please make the TextField the toplevel component, and just put the signing phrase repeater inside it

caybro: You won't need to introduce all those toplevel properties

caybro (changes requested):

saledjenic:
@caybro good point, I missed that cause firstly I did it bit differently and then change to this way. Anyhow... thanks, updated.

osmaczko (changes requested):

src/StatusQ/Controls/StatusPasswordInput.qml
```diff
@@ -0,0 +1,108 @@
...
+                height: d.signingPhraseWordsHeight
+                color: Theme.palette.indirectColor1
+                radius: d.radius
+
+                StatusBaseText {
+                    id: signingPhraseWord
+                    anchors.centerIn: parent
+                    color: Theme.palette.primaryColor1
+                    font.pixelSize: d.pixelSize
+                    font.family: Theme.palette.baseFont.name
```
caybro, line 102: No need to set the font.family

src/StatusQ/Controls/StatusPasswordInput.qml
```diff
@@ -0,0 +1,108 @@
...
+                width: signingPhraseWord.implicitWidth + 2 * d.signingPhraseWordPadding
+                height: d.signingPhraseWordsHeight
+                color: Theme.palette.indirectColor1
+                radius: d.radius
+
+                StatusBaseText {
+                    id: signingPhraseWord
+                    anchors.centerIn: parent
+                    color: Theme.palette.primaryColor1
+                    font.pixelSize: d.pixelSize
```
caybro, line 101: I'd have the `font.pixelSize` just follow `root.font.pixelSize`

caybro (approved):
Minor stuff but otherwise good to go imo

src/StatusQ/Controls/StatusPasswordInput.qml
```diff
@@ -0,0 +1,108 @@
...
+        readonly property int signingPhraseWordPadding: 8
+        readonly property int signingPhraseWordsSpacing: 8
+        readonly property int signingPhraseWordsHeight: 30
+
+    }
+
+    leftPadding: d.inputTextPadding
+    rightPadding: root.signingPhrase !== ""?
+                      phrase.width + phrase.anchors.leftMargin + phrase.anchors.rightMargin :
+                      d.inputTextPadding
```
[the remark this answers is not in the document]
osmaczko: Ok, I see the implementation has changed in the meantime when I was writing this comment :sweat_smile: 

Making `TextField` a root component and doing this:
```
    RowLayout {
        id: phrase
        anchors.right: parent.right
    }
```
is not a good idea. `StatusPasswordInput` will have the wrong implicit sizes and as a consequence, it will be impossible to use it correctly in layouts. If the layout is clipped the `phrase` will become invisible.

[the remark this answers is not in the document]
saledjenic: @osmaczko I am not sure that I follow what you want to be changed here?

> Making TextField a root component and doing this:

Have you seen how this component should look like?

> As an experiment you could try to do this:

Have you tried this?

I don't see any issues in the way it's done.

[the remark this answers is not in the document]
osmaczko: > Have you tried this?

No, sorry  :sweat: , I just tried it now, and it works as expected within layouts

I analyzed the code and built the wrong picture in my head. My above concerns would be only true if: `anchors.left: parent.right` which is not the case. Forget about the comment :D

osmaczko (approved):

PR #892, opened 2022-09-08
feat(StatusMessage): add support for `messageAttachments`

PascalPrecht:
Prior to this commit, a `StatusMessage` can hold only a single `messageImage`.

There will be scenarios where messages can have multiple attachments. This is the case when importing messages from discord. Hence, this commit introduces a new `messageAttachments` property which is a whitespace separated list of attachment URLs (very analogous to the already existing `linkUrls` property).

For now, we can safely assume these URLs will resolve to image content. In future versions however, we might want to support additional content types to handle any byte stream.
We will then change `messageAttachments` from `string` to a list model that holds `attachment`s where we also have access to its `contentType`. This will allow us to decide at runtime which component should be used to view/play the attachment.

### Checklist

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Components/StatusMessage.qml
+++ b/src/StatusQ/Components/StatusMessage.qml
@@ -37,6 +37,7 @@ Rectangle {
     property string errorLoadingImageText: ""
     property string audioMessageInfoText: ""
     property string pinnedMsgInfoText: ""
+    property string messageAttachments: ""
     property var reactionIcons: [
         Emoji.iconSource("❤"),
         Emoji.iconSource("👍"),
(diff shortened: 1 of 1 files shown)


src/StatusQ/Components/StatusMessage.qml
```diff
@@ -37,6 +37,7 @@ Rectangle {
     property string errorLoadingImageText: ""
     property string audioMessageInfoText: ""
     property string pinnedMsgInfoText: ""
+    property string messageAttachments: ""
```
caybro, line 40: Any reason why this is not an array? (I assume it's because NIM doesn't support array-like Qt properties? :/

caybro (approved):
I guess this will work for now; however I'd prefer a regular QAIM model where you could generate and specify additional "metadata", like MIME type or preview image

src/StatusQ/Components/StatusMessage.qml
PascalPrecht: Correcto

PR #904, opened 2022-09-14
feat(StatusPinInput): additional spacing for every N-th digit

saledjenic:
Two new props exposed so we can sed an extra spacing for every N-th item.
- `additionalSpacingOnEveryNItems` - determines for which item to add an extra space
- `additionalSpacing` - determines the value of extra space which will be added

### DIFF
--- a/sandbox/pages/StatusPinInputPage.qml
+++ b/sandbox/pages/StatusPinInputPage.qml
@@ -72,4 +72,73 @@ Column {
         text: "Introduced PIN: " + regexPinInput.pinInput
         font.pixelSize: 12
     }
+
+    // PUK input that accepts only numbers
+    StatusBaseText {
+        topPadding: 100
+        anchors.horizontalCenter: parent.horizontalCenter
+        color: Theme.palette.directColor1
+        text: "Enter Keycard PUK"
+        font.pixelSize: 30
+        font.bold: true
+    }
+
+    StatusPinInput {
+        id: numbersPukInput
+        anchors.horizontalCenter: parent.horizontalCenter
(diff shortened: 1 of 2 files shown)


src/StatusQ/Controls/StatusPinInput.qml
```diff
@@ -151,8 +165,8 @@ Item {
         }
     }
 
-    width: (root.circleDiameter + root.circleSpacing) * root.pinLen - root.circleSpacing
-    height: root.circleDiameter
+    width: childrenRect.width
+    height: childrenRect.height
```
osmaczko, line 169:
```suggestion
    implicitWidth: childrenRect.width
    implicitHeight: childrenRect.height
```

osmaczko (approved):

sandbox/pages/StatusPinInputPage.qml
```diff
@@ -72,4 +72,73 @@ Column {
...
+        id: numbersPukInput
+        anchors.horizontalCenter: parent.horizontalCenter
+        validator: StatusRegularExpressionValidator { regularExpression: /[0-9]+/ }
+        pinLen: 12
+    }
+
+    StatusBaseText {
+        anchors.horizontalCenter: parent.horizontalCenter
+        color: Theme.palette.dangerColor1
+        text: "Only numbers allowed"
```
stefandunca, line 96: Won't this and the below user facing text need translations?

stefandunca (approved):

sandbox/pages/StatusPinInputPage.qml
stefandunca: Ok, it makes sense. Still I personally use qstr to label user-facing strings vs. internal debugging.

PR #908, opened 2022-09-15
fix(StatusBaseButton): font family and weight fixed

micieslak:
### Checklist

Font family set for StatusBaseButton, font variants moved to internal object to avoid accessing them directly.

In the future we should consider switching to variable fonts in order to avoid loading multiple font variants.

Screenshot 2022-09-15 16:25:12

- [ ] update sandbox app

### DIFF
--- a/src/StatusQ/Controls/StatusBaseButton.qml
+++ b/src/StatusQ/Controls/StatusBaseButton.qml
@@ -42,6 +42,7 @@ Button {
         readonly property color textColor: root.enabled || root.loading ? root.textColor : root.disabledTextColor
     }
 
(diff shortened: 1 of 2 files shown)


faust4exe (approved):
Looks good, just add few screenshots which can confirm how it looks for you.

caybro (approved):

osmaczko (approved):

alexandraB99 (approved):
Just to make sure - Did you check all other components that work as expected?

micieslak:
> Just to make sure - Did you check all other components that work as expected?

Yes, I checked, but I would feel more comfortable if somebody could test it as well.

